PR #59, opened 2025-09-03
add ```num_worst_token```

inkcherry:
## Motivation

Adjustable ```max_token_recv_per_rank```, allowing for reduced memory overhead in some load balancing scenarios.
FYI @zhenhuan12

### DIFF
--- a/include/mori/ops/dispatch_combine/dispatch_combine.hpp
+++ b/include/mori/ops/dispatch_combine/dispatch_combine.hpp
@@ -79,6 +79,7 @@ struct EpDispatchCombineConfig {
   int maxNumInpTokenPerRank{128};
   int numExpertPerRank{1};
   int numExpertPerToken{2};
+  int numWorstToken{0};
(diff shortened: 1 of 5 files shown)


include/mori/ops/dispatch_combine/dispatch_combine.hpp
```diff
@@ -98,6 +99,9 @@ struct EpDispatchCombineConfig {
   }
 
   inline __host__ __device__ int MaxNumTokensToRecv() const {
+    if (numWorstToken != 0) {
+      return numWorstToken;
```
isytwu: I wonder if return worldSize * numWorstToken will be better?

isytwu: Perhaps using min(numWorstToken, worldSize * MaxNumTokensToRecvPerRank()) could prevent users from passing  large values. And should  MaxNumTokensToSend() also be changed to add numWorstToken?

PR #64, opened 2025-09-05
Feature: support bnxt ibgda

jhchouuu:
**Broadcom BNXT IBGDA Support**
BNXT IBGDA functions are now supported. 
Use `USE_BNXT=ON` to enable BNXT features.

**Shmem API implementation modification**
Shmem API now returns one CQE per batch of WQE.

**IBGDA API Unification**
BNXT and MLX5 share the same basic operation API. 
Added `cqeSignal` option for CQE signaling control.


src/application/transport/rdma/providers/bnxt/bnxt.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+  memcpy(&attr.ah_attr.grh.dgid, remote_handle.eth.gid, 16);
+  attr.ah_attr.grh.sgid_index = config.gidIdx;
+  attr.ah_attr.grh.hop_limit = 1;
+  attr.ah_attr.sl = 1;
+  attr.ah_attr.is_global = 1;
+  attr.ah_attr.port_num = config.portId;
+
+  // TODO: max_dest_rd_atomic whether affect nums of amo/rd
+  attr.max_dest_rd_atomic = 1;
```
TianDi101: @jhchouuu I suggest setting this filed as the following line. This parameter determines the max # of read/atomic wr inflight, setting to 1 makes it very hard to saturate read/atomic bandwidth. I met this problem earlier when developing MORI-IO.
`attr.max_dest_rd_atomic = devAttr->orig_attr.max_qp_rd_atom;`

include/mori/core/transport/rdma/providers/bnxt/bnxt_device_primitives.hpp
```diff
@@ -0,0 +1,787 @@
...
+  void* msntblAddr = wq.msntblAddr;
+  uint32_t wqeNum = wq.sqWqeNum;
+  uint32_t mtuSize = wq.mtuSize;
+
+  struct bnxt_re_bsqe hdr;
+  struct bnxt_re_send send;
+  struct bnxt_re_sge sge;
+  constexpr int sendWqeSize =
+      sizeof(struct bnxt_re_bsqe) + sizeof(struct bnxt_re_send) + sizeof(struct bnxt_re_sge);
+  constexpr int slotsNum = CeilDiv(sendWqeSize, BNXT_RE_SLOT_SIZE);
```
TianDi101, line 114: @jhchouuu I remember that BNXT requires all wqe have the same size and no empty slot should exist, for send/recv/read/write/inline/atomic, can we satisfy this condition?

include/mori/core/transport/rdma/providers/bnxt/bnxt_device_primitives.hpp
```diff
@@ -0,0 +1,787 @@
...
+  // In bnxt, wqeNum mean total sq slot num
+  uint8_t signalFlag = cqeSignal ? BNXT_RE_WR_FLAGS_SIGNALED : 0x00;
+  void* queueBuffAddr = wq.sqAddr;
+  void* msntblAddr = wq.msntblAddr;
+  uint32_t wqeNum = wq.sqWqeNum;
+  uint32_t mtuSize = wq.mtuSize;
+
+  struct bnxt_re_bsqe hdr;
+  struct bnxt_re_send send;
+  struct bnxt_re_sge sge;
```
TianDi101, line 111: @jhchouuu  Maybe initialize it explicitly as the following code? On CPU-side code, since bnxt dv header is C code, initialize it without brackets may have undefined fields which often cause weird memory corruption bugs(I've met this kind of bugs multiple times and it's hard to debug). I'm not sure if GPU-side code has the same issue, but since HIPCC is based on Clang/LLVM, I think it's safer to use bracket initialization.
`struct bnxt_re_bsqre hdr{}`

TianDi101:
Amazing PR! I've left several comments but overall, the code LGTM. Thanks for the hard work @jhchouuu

src/application/transport/rdma/providers/bnxt/bnxt.cpp
jhchouuu: Very good suggestion. I will make this modification.

include/mori/core/transport/rdma/providers/bnxt/bnxt_device_primitives.hpp
jhchouuu: In fact, bnxt has a very flexible size for each wqe, currently a fixed-sized wqe is being used, with 3 slots.

include/mori/core/transport/rdma/providers/bnxt/bnxt_device_primitives.hpp
jhchouuu: Good suggestion. This indeed can enhance robustness.

jhchouuu: After checking, although the fields were not explicitly initialized, they will all be filled with new values after. Therefore, brackets may not be necessary, thus saving some cycles.

PR #70, opened 2025-09-16
Optim: bnxt ibgda basic ops and shmem basic ops

jhchouuu:
- Modify the management method of bnxt ibgda slot index.
- Optimize the queue size to a power of 2.
- Calculate the queue depth instead of using the bnxt_re API.
- Modified some default parameters of bnxt NIC.


TianDi101:
LGTM

PR #74, opened 2025-09-22
Fix: issue that use same hostname within cluster

jhchouuu:
## Motivation

In some clusters, all the nodes will use the same hostname. The method of determining whether they are the same node by using the hostname is not robust. Therefore, it is recommended to modify the judgment method to <hostname:ip>.

### DIFF
--- a/src/application/context/context.cpp
+++ b/src/application/context/context.cpp
@@ -21,9 +21,15 @@
 // SOFTWARE.
 #include "mori/application/context/context.hpp"
 
+#include <arpa/inet.h>
 #include <hip/hip_runtime.h>
+#include <ifaddrs.h>
+#include <netdb.h>
+#include <string.h>
 #include <unistd.h>
 
+#include <iostream>
+#include <string>
 #include <vector>
 
 #include "mori/application/utils/check.hpp"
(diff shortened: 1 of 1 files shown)


src/application/context/context.cpp
```diff
@@ -38,18 +44,60 @@ Context::Context(BootstrapNetwork& bootNet) : bootNet(bootNet) {
 
 Context::~Context() {}
 
+std::string GetLocalIP() {
+  struct ifaddrs *ifaddr, *ifa;
+  char host[NI_MAXHOST];
+  std::string localIP = "127.0.0.1";
+
+  if (getifaddrs(&ifaddr) == -1) {
```
TianDi101, line 52: @jhchouuu Are the interfaces returned by this API sorted? Is the order affected by docker image configurations? I'm worrying about if the order is not deterministic, ranks may be considered as on different nodes.

jhchouuu: Sorry, I think we really should consider alternatives ifaddrs like docker, rather than just consider localhost ifaddr. We will make further improvements.

TODO:
- [ ] Add a judgment for docker ifaddr

PR #80, opened 2025-10-02
Fix: fix inconsistency in internode dispatch/combine

dongmin-ra:
## Motivation
Fixed an intermittent issue where the results of dispatch would be corrupted when running dispatch/combine repeatedly
- The condition for this issue to occur is that there must be no global barrier (i.e. `torch.distributed.barrier()`) between dispatch and combine.
## Technical Details
**Issue**
- After integrating mori-EP into vLLM ([refer](https://github.com/vllm-project/vllm/pull/26013)), an intermittent GPU memory access fault occurred during multi-node EP.
  - This happened because the expert indices after dispatch were corrupted on some ranks.
- After investigating it, I figured out that the dispatch result could be corrupted in internode EP.

**Cause**
- Combine and dispatch share the same input buffer, `shmemInpTokMemObj`.
  - In internode dispatch, the last warp, after collecting data from remaining warps, sends data to the `shmemInpTokMemObj` buffer on remote GPU. 
    - During the recv phase, data is copied from local `shmemInpTokMemObj` to local `shmemOutTokMemObj`.
  - In combine’s send phase, similar to dispatch, The last warp sends data to `shmemInpTokMemObj` on the remote GPU.
- If combine starts immediately after dispatch on a fast GPU, its send phase may perform RDMA writes to `shmemInpTokMemObj` before **dispatch’s recv phase has completed the memcpy**. 
  - This can overwrite the buffer and cause data corruption.

**Fix**
- Separated the input buffers used by dispatch and combine.
  
## Test Plan
```
pytest ./tests/python/ops/test_dispatch_combine_internode_inconsistency.py -s
```
> This should be tested on a single node. Internally, the `MORI_DISABLE_P2P` environment variable is enabled to force communication via RDMA within the single node.
## Test Result
- Before modification : incorrect expert index values are produced as the dispatch result.
```
=================================================================================================================================================== test session starts ====================================================================================================================================================
platform linux -- Python 3.12.11, pytest-8.4.1, pluggy-1.6.0
rootdir: /app/mori
plugins: assume-2.4.3, anyio-4.9.0, asyncio-1.0.0
asyncio: mode=Mode.STRICT, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collecting ...
collected 1 item

tests/python/ops/test_dispatch_combine_internode_inconsistency.py Multiprocessing start method set to spawn

rank 0 RDMA devices: mlx5_0, mlx5_2, mlx5_3, mlx5_4
rank 0 rankInNode 0 select device [0] mlx5_0
rank 3 rankInNode 3 select device [3] mlx5_4
rank 6 rankInNode 6 select device [0] mlx5_0
rank 5 rankInNode 5 select device [3] mlx5_4
rank 7 rankInNode 7 select device [1] mlx5_2
rank 2 rankInNode 2 select device [2] mlx5_3
rank 1 rankInNode 1 select device [1] mlx5_2
rank 4 rankInNode 4 select device [2] mlx5_3
Passed 0/2048
Passed 1/2048
Passed 2/2048
...
Passed 33/2048
Passed 34/2048
Passed 35/2048
Invalid expert id: 1261946812
```
- After modification : no error occurs
## Submission Checklist

### DIFF
--- a/tests/python/ops/test_dispatch_combine_internode_inconsistency.py
+++ b/tests/python/ops/test_dispatch_combine_internode_inconsistency.py
@@ -157,7 +157,7 @@ class EpDispatchCombineTestCase:
 
 @pytest.fixture(scope="session")
 def torch_dist_process_manager():
-    os.environ['MORI_DISABLE_P2P'] = '1'
+    os.environ["MORI_DISABLE_P2P"] = "1"
     try:
         torch.multiprocessing.set_start_method("spawn", force=True)
         print("Multiprocessing start method set to spawn")
@@ -192,7 +192,7 @@ def _test_dispatch_combine(
         num_experts_per_token=num_experts_per_token,
         max_token_type_size=2,
         block_num=16,
-        warp_num_per_block=1,
+        warp_num_per_block=16,
         kernel_type=mori.ops.EpDispatchCombineKernelType.InterNode,
     )
     op = mori.ops.EpDispatchCombineOp(config)
@@ -200,9 +200,9 @@ def _test_dispatch_combine(
     test_data = test_case.gen_test_data(True)
     num_reps = 2048
     for idx in range(num_reps):
-      test_case.run_test_once(op, test_data)
-      if rank == 0:
-          print(f"Passed {idx}/{num_reps}")
+        test_case.run_test_once(op, test_data)
+        if rank == 0:
+            print(f"Passed {idx}/{num_reps}")
 
 
 # TODO: create a sub process group so that we can test worlds size < 8
@@ -211,7 +211,7 @@ def _test_dispatch_combine(
 @pytest.mark.parametrize("hidden_dim", (7168,))
 @pytest.mark.parametrize("scale_dim", (56,))
 @pytest.mark.parametrize("scale_type_size", (4,))
-@pytest.mark.parametrize("max_num_inp_token_per_rank", (256,))
+@pytest.mark.parametrize("max_num_inp_token_per_rank", (4096,))
 @pytest.mark.parametrize("num_experts_per_rank", (32,))
 @pytest.mark.parametrize("num_experts_per_token", (8,))
 def test_dispatch_combine(


include/mori/ops/dispatch_combine/dispatch_combine.hpp
```diff
@@ -169,7 +169,8 @@ class EpDispatchCombineHandle {
   uint8_t* scalesBuf{nullptr};
 
   // Registered buffers for tokens, shmemOutTokMemObj will be returned to user as output
-  mori::application::SymmMemObjPtr shmemInpTokMemObj;
+  mori::application::SymmMemObjPtr dispatchShmemInpTokMemObj;
+  mori::application::SymmMemObjPtr combineShmemInpTokMemObj;
```
jhchouuu: Could you please change the variable names?
`dispatchShmemInpTokMemObj -> shmemDispatchInpTokMemObj`
`combineShmemInpTokMemObj -> shmemCombineInpTokMemObj`
We don't have the permission for **moreh-dev**, so we can't submit the commit.

dongmin-ra: Renamed the variables as suggested 👍

tests/python/ops/test_dispatch_combine_internode_inconsistency.py
```diff
@@ -0,0 +1,255 @@
...
+        world_size=world_size,
+        hidden_dim=hidden_dim,
+        scale_dim=scale_dim,
+        scale_type_size=scale_type_size,
+        max_num_inp_token_per_rank=max_num_inp_token_per_rank,
+        num_experts_per_rank=num_experts_per_rank,
+        num_experts_per_token=num_experts_per_token,
+        max_token_type_size=2,
+        block_num=16,
+        warp_num_per_block=1,
```
jhchouuu: @dongmin-ra  Could you please help to change the "warp_num_per_block" to 16, `warp_num_per_block=16` ? Otherwise, the test will not pass. However, it might be caused by the new changes incorporated into the kernel.

jhchouuu: After our testing on vllm, this fix also resolves the issue. We will merge it soon.
@dongmin-ra thanks to moreh team again

dongmin-ra: Changed `warp_num_per_block` to 16. Since it’s a timing issue that tends to occur when each warp has a lot of work to do, I also increased the number of input tokens.

PR #92, opened 2025-10-23
Enhancement of mori ep unit test

dongmin-ra:
This is a work by a team from Moreh Inc.
## Motivation
Currently, Mori EP unit test (i.e. `test_dispatch_combine.py`) does not sufficiently cover real cases.
#### Problem 1. Each test case performed dispatch/combine only once.
 - In practice, depending on the model, dispatch/combine is executed multiple times corresponding to the number of layers.
   - e.g. In DeepSeek-R1, dispatch/combine is performed 58 times per step.
- When dispatch/combine is executed multiple times, there may be cases where the results are incorrect, but currently this is not being checked.
#### Problem 2. There is  synchronization across devices before checking the results.
- In practice, for performance reasons, explicit synchronization is not performed after dispatch and combine. 
- Most issues with the output values occur due to device-to-device timing differences.
  - Since the unit test is synchronizing across devices after dispatch/combine, it cannot detect the timing issue.
## Technical Details
Updated the unit tests to more closely resemble real use cases:
 - Introduced the `num_reps` parameter to allow each test case to perform dispatch/combine multiple times.
 - Removed explicitly synchronize across devices after dispatch/combine.
    - Instead, results are stored in a list during execution.
    - verification is performed after all dispatch/combine operations are complete.
> [!Note] 
> Currently, there is an issue in Mori EP where results are incorrect if `num_reps` is greater than 1. I will create a PR to fix this.

Fixed an issue where the test would hang if an error occurred on some ranks:
- Previously, the parent process waited for results from all processes, but if an error occurred on some ranks, other ranks continued executing dispatch/combine and never sent a test result to the result queue.
  - As a result, the parent process would wait indefinitely for responses that never arrive, causing the test to hang.
- Changed the behavior so that if an error occurs in any process, the error is raised immediately without waiting for the result queue.
## Test Plan
```
pytest -s -v tests/python/ops/test_dispatch_combine.py
```
## Test Result
```
=================================================================================================================================================== test session starts ====================================================================================================================================================
platform linux -- Python 3.12.11, pytest-8.4.2, pluggy-1.6.0 -- /usr/bin/python
cachedir: .pytest_cache
rootdir: /app/mori
plugins: assume-2.4.3, anyio-4.10.0, asyncio-1.1.0
asyncio: mode=Mode.STRICT, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 48 items

[...]

PR #93, opened 2025-10-23
Fix: fix output buffer overwrite issue in intranode EP.

dongmin-ra:
This is a work by a team from Moreh Inc.
## Motivation
Fixed the problem where the output occasionally becomes incorrect.

### Issue
When running the intranode EP, the output occasionally becomes incorrect.
### Cause
This issue occurs because both the dispatch and combine phases share the same output buffers, 
`shmemOutTokMemObj` and `shmemOutWeightsMemObj`.
Since these buffers are reused across phases, they can be overwritten due to differences in execution timing between ranks.

In intranode EP, each dispatch/combine kernel behaviors are as follows:
- Dispatch (send phase): Sends data to the remote device’s `shmemOutTokMemObj` and `shmemOutWeightsMemObj`.
- Combine (send phase): Writes the combine output and output weights into its own`shmemOutTokMemObj` and `shmemOutWeightsMemObj`.
The Core Issue

Because the same buffers are reused for every dispatch/combine call, the combine result should be copied to the user buffer.
- Within a single rank, the sequence dispatch → memcpy → combine → memcpy is strictly ordered, so it’s safe.
- However, across ranks, there’s no guaranteed order between dispatch and combine phases except for the cross barrier used inside combine.

While one rank is performing the **combine recv phase**, another rank might already be executing the **dispatch send phase**.
- If there’s a delay between the combine call and the memcpy that follows it, another rank’s dispatch may overwrite the shared buffer in that gap.
- This situation makes output incorrect.
image

### Fix
Split `shmemOutTokMemObj` and `shmemOutWeightsMemObj` into seperate objects for dispatch and combine

## Test Plan
1) Merge [moreh-dev:enhance_mori_ep_test](https://github.com/moreh-dev/mori/tree/enhance_mori_ep_test) and [moreh-dev:fix_ep_result_error](https://github.com/moreh-dev/mori/tree/fix_ep_result_error) branches
  - `moreh-dev:fix_ep_result_error` : fixed possible result value error in internode and intranode EP (https://github.com/ROCm/mori/pull/87)
  - `moreh-dev:enhance_mori_ep_test` : enhanced mori ep unit test. (https://github.com/ROCm/mori/pull/92)
2) Modify  `tests/python/ops/test_dispatch_combine.py` as below:
```
diff --git tests/python/ops/test_dispatch_combine.py tests/python/ops/test_dispatch_combine.py
index c515434..d9a4dd7 100644
--- tests/python/ops/test_dispatch_combine.py
+++ tests/python/ops/test_dispatch_combine.py
@@ -454,7 +454,7 @@ def _test_dispatch_combine(
 @pytest.mark.parametrize("max_num_inp_token_per_rank", (1, 128, 2048))
 @pytest.mark.parametrize("num_experts_per_rank", (32,))
 @pytest.mark.parametrize("num_experts_per_token", (8,))
-@pytest.mark.parametrize("num_reps", (1,))
+@pytest.mark.parametrize("num_reps", (8,))
 def test_dispatch_combine(
     torch_dist_process_manager,
     world_size,
```
3. Execute the test script
```
pytest -s -v tests/python/ops/test_dispatch_combine.py
```
## Test Result
Before modification: the test fails
```
E       ValueError: Caught ValueError in rank 0.
E       Original Traceback (most recent call last):
E         File "/app/mori/tests/python/utils.py", line 171, in _worker
E           result = func(rank, *args)
E                    ^^^^^^^^^^^^^^^^^
E         File "/app/mori/tests/python/ops/test_dispatch_combine.py", line 445, in _test_dispatch_combine
E           test_case.run_test(op, test_data)
E         File "/app/mori/tests/python/ops/test_dispatch_combine.py", line 375, in run_test
E           self.check_result(test_data, mori_result, i)
E         File "/app/mori/tests/python/ops/test_dispatch_combine.py", line 400, in check_result
E           self.check_combine_result(test_data, combine_output, combine_weights, round)
E         File "/app/mori/tests/python/ops/test_dispatch_combine.py", line 269, in check_combine_result
E           raise ValueError(error_msg)
E       ValueError: 0-th Combine result mismatch for token 0:
E         indices[0]: [104, 49, 208, 235, 114, 18, 28, 55]
E         got: tensor([ 9.3259e-15, -5.9292e-20, -1.2821e-22,  ...,  3.1641e-01,
E                3.7969e+00, -8.4375e-01], device='cuda:0', dtype=torch.bfloat16)
E         expected : tensor([ 1.9922,  0.9961, -1.2500,  ...,  0.3164,  3.7969, -0.8438],
E              device='cuda:0', dtype=torch.bfloat16)
```
After modification: all test passed
## Submission Checklist

### DIFF
--- a/examples/ops/dispatch_combine/test_dispatch_combine.cpp
+++ b/examples/ops/dispatch_combine/test_dispatch_combine.cpp
@@ -249,7 +249,8 @@ class EpDispatchCombineTestCase {
         index_t srcPe = srcTokId / config.maxNumInpTokenPerRank;
         index_t localSrcTokId = srcTokId - srcPe * config.maxNumInpTokenPerRank;
 
-        T* localTokBuf = handle.shmemOutTokMemObj->template GetAs<T*>() + i * config.hiddenDim;
+        T* localTokBuf =
+            handle.shmemDispatchOutTokMemObj->template GetAs<T*>() + i * config.hiddenDim;
         T* srcTokBuf = reinterpret_cast<T*>(globalInpTokBufCpu) + srcPe * inpTokEleNum +
                        localSrcTokId * config.hiddenDim;
 
@@ -291,7 +292,8 @@ class EpDispatchCombineTestCase {
         index_t srcTokDispatchId = peSortToTokenIdxMapsVec[srcPe][peSortedId];
         index_t srcTokId = srcTokDispatchId / config.numExpertPerToken;
 
-        T* localTokBuf = handle.shmemOutTokMemObj->template GetAs<T*>() + i * config.hiddenDim;
+        T* localTokBuf =
+            handle.shmemDispatchOutTokMemObj->template GetAs<T*>() + i * config.hiddenDim;
 
         T* srcTokBuf = reinterpret_cast<T*>(globalInpTokBufCpu) + srcPe * inpTokEleNum +
                        srcTokId * config.hiddenDim;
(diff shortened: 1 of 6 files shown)


jhchouuu:
Thanks to moreh team! We will check these PRs as soon as possible.

jhchouuu:
LGTM, the same problem as the same input buffer between dispatch and combine, thanks

PR #96, opened 2025-10-27
Feature: support symmetric heap and unified memory space

jhchouuu:
## Motivation

support symmetric heap and unified memory space

## Technical Details

1. support static symmetric heap and unified memory space

      shmem support two style api:
      SymmMemObjPtr + offset: 
      ```
      ShmemPutMemNbiThread(
          const application::SymmMemObjPtr dest, size_t destOffset,
          const application::RdmaMemoryRegion& source, size_t sourceOffset, size_t bytes, int pe,
          int qpId = 0) 
      ```
      
      unified memory :
      ```
      ShmemPutMemNbiThread(void* dest, const void* source, size_t bytes,
                                                  int pe, int qpId = 0)
      ```

2. The symmetric heap currently only supports static heaps.
use env variable MORI_SHMEM_HEAP_SIZE to set default heap memory size

3. enhance compatibility by implementing isolation mode
use env variable MORI_SHMEM_MODE=isolation to use the pre-existing non-heap implementation mode

## Test Result

test pass:

- `mpirun --allow-run-as-root -np 2 ./examples/concurrent_put_thread`
- `mpirun --allow-run-as-root -np 2 ./examples/concurrent_put_imm_thread `
- `mpirun --allow-run-as-root -np 2 ./examples/atomic_fetch_thread `
- `mpirun --allow-run-as-root -np 2 ./examples/atomic_nonfetch_thread `
- `mpirun --allow-run-as-root -np 2 ./examples/concurrent_put_signal_thread `
- `torchrun --nnodes=2 --node_rank=0 --nproc_per_node=1 --master_addr=xxx --master_port=1236 examples/ops/dispatch_combine/test_dispatch_combine_internode.py`

### DIFF
--- a/src/shmem/init.cpp
+++ b/src/shmem/init.cpp
@@ -57,53 +57,58 @@ void MemoryStatesInit() {
   states->memoryStates->mrMgr =
       new application::RdmaMemoryRegionManager(*context->GetRdmaDeviceContext());
 
-  // Allocate static symmetric heap
-  // Size can be configured via environment variable
-  const char* heapSizeEnv = std::getenv("MORI_SHMEM_HEAP_SIZE");
-  size_t heapSize = DEFAULT_SYMMETRIC_HEAP_SIZE;
-  
-  if (heapSizeEnv) {
-    std::string heapSizeStr(heapSizeEnv);
-    size_t multiplier = 1;
+  // Only allocate static heap in StaticHeap mode
+  if (states->mode == ShmemMode::StaticHeap) {
+    // Allocate static symmetric heap
+    // Size can be configured via environment variable
+    const char* heapSizeEnv = std::getenv("MORI_SHMEM_HEAP_SIZE");
+    size_t heapSize = DEFAULT_SYMMETRIC_HEAP_SIZE;
     
-    // Check for suffix
-    if (!heapSizeStr.empty()) {
-      char lastChar = heapSizeStr.back();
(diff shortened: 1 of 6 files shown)


jhchouuu:
rebase and enhance compatibility by implementing isolation mode

src/shmem/init.cpp
```diff
@@ -53,6 +56,59 @@ void MemoryStatesInit() {
...
+    // the first ShmemMalloc would return staticHeapBasePtr, which is the same address
+    // as the heap itself in memObjPool, causing the heap's SymmMemObj to be overwritten.
+    constexpr size_t HEAP_INITIAL_OFFSET = 256;
+    states->memoryStates->staticHeapUsed = HEAP_INITIAL_OFFSET;
+    states->memoryStates->staticHeapObj = heapObj;
+
+    MORI_SHMEM_INFO("Static symmetric heap allocated at {} (local), size {} bytes, initial offset {} bytes",
+                    states->memoryStates->staticHeapBasePtr, heapSize, HEAP_INITIAL_OFFSET);
+  } else {
+    MORI_SHMEM_INFO("Running in isolation mode (no static heap)");
```
TianDi101, line 110: @jhchouuu With isolation mode, we cannot use pointer-based shmem API, right?

jhchouuu: Yes, it is difficult to maintain all the SymmMemObj on the device side.

PR #128, opened 2026-01-08
Optimization: EPV1 dispatch & combine kernel

TianDi101:


jhchouuu:
It seems that the main improvement was made to the recv phase of the dispatch LL kernel. 
if there is data with 64 tokens, and also with CX7 and MI300?

TianDi101:
> It seems that the main improvement was made to the recv phase of the dispatch LL kernel. if there is data with 64 tokens, and also with CX7 and MI300?

Yes, the perf gains come from recv phase, mainly XGMI part. I guess it should also improve CX7+MI300X.
I just attached token=64 perf improvement for your reference.

jhchouuu:
cc @isytwu

PR #173, opened 2026-02-24
Feature: make MORI framework agnostic and compatible with JAX-XLA

TianDi101:


i-chaochen:
Thanks @TianDi101 for bring it as the PR!

> Hi @i-chaochen , thanks for the work, some questions:
> 
> * How does JAX calls custom library? For Torch, we use pybind + torch module, I'm wondering whether JAX shares the same path?

in JAX, it uses FFI https://docs.jax.dev/en/latest/ffi.html#foreign-function-interface-ffi to have front code and gpu code, you can see [our simple use code here](https://github.com/ROCm/jaxpp/blob/pemeliya/mori_integration/src/custom_kernels/mori_demo.py)  // jaxpp is just another wrapper atop of jax.

Back to your question, for jax, yes, we're using pybind11 + xla module, and you can see our initial POC branch https://github.com/ROCm/mori/commit/b2a2bdbc17e1eb53fb17882b367fdd2d6d20317e#diff-fe7afb5c9c916e521401d3fcfb4277d5071798c3baf83baf11d6071742823584 

> * Ever considered using TVM-FFI? It is a cross-framework / cross-language FFI standard which seems suits this scenario.

Yes, it's FFI as I mentioned in the above, but I'm not familar with TVM-FFI, not sure whether is same stuff and compatible with other frameworks. Essentially, it's using "pjrt plugin" which is a set of C APIs (within XLA) to define a binary interface.

As a side note, we have rebased and simplified this separation here https://github.com/ROCm/mori/commits/p_cleanup_torch_mpi_deps/ Hope this is easier to accept.

- the separation between host and device code is still not moved here, we still need it.


cc @pemeliya

jhchouuu:
Hi @i-chaochen and @pemeliya , I've been working on a similar direction on the `jiahzhou/refactor_framework` branch — making the pybind layer fully torch-free. Here's a summary of what I've done that might overlap or help with this PR:
https://github.com/ROCm/mori/tree/jiahzhou/refactor_framework

## 1. pybind_ops.cpp: torch dependency fully removed

All `torch::Tensor` parameters replaced with raw `int64_t` pointers, `torch::from_blob` replaced with returning raw pointers, `TORCH_CHECK` moved to Python, `getCurrentHIPStream` passed from Python as `int64_t`. The C++ pybind layer now has zero torch includes — no `torch/python.h`, no `ATen/hip/HIPContext.h`, no `torch_python` link dependency.

## 2. Python-side tensor conversion via DLPack + `__cuda_array_interface__`

A new `tensor_utils.py` provides `GpuTensorView` — a lightweight GPU array descriptor implementing both `__dlpack__`/`__dlpack_device__` and `__cuda_array_interface__`. Any framework can consume it zero-copy:

```python
view = gpu_tensor_view(ptr, shape, dtype)
t = torch.as_tensor(view)           # PyTorch
a = jax.dlpack.from_dlpack(view)    # ? JAX
```

https://github.com/ROCm/mori/blob/jiahzhou/refactor_framework/python/mori/tensor_utils.py
Supports float32, bfloat16, float8, float4_e2m1fn_x2, int32.

I haven't been able to test the JAX path end-to-end — I don't have a jax[rocm] environment set up. The DLPack capsule is created correctly (`__dlpack_device__` returns the proper ROCm device type), but JAX consumption is untested. Would appreciate guidance on how to set up a jax[rocm] test environment if you have one available.

These changes are on the `jiahzhou/refactor_framework` branch and still a work in progress. I've also looked at the `p_cleanup_torch_mpi_deps` branch — there's significant overlap in goals but different approaches (e.g., `#ifdef ENABLE_TORCH` vs fully removing torch from the pybind API). Happy to discuss whether these conflict or can complement each other.

cc @TianDi101

PR #180, opened 2026-03-03
[UMP] Feat: add control plane implementation for unified memory pool

TianDi101:
## Motivation

This PR implements the prototype UMBP master control-plane.

## Technical Details

See design-master-control-plane.md

## Test Plan

TBD

## Test Result

TBD


TianDi101:
@maning00 Please help to review master control plane design

src/umbp/doc/design-master-control-plane.md
```diff
@@ -0,0 +1,2049 @@
...
+                node_address: target.node_address,
+                tier: tier}
+
+    return nullopt
+```
+
+**Why most available space?** Spreading writes to the least-loaded node
+balances memory pressure across the cluster, avoids creating hotspots on
+nearly-full nodes, and reduces the chance of a write failing due to a race
+between RoutePut and the actual MORI-IO write.
```
maning00, line 561: "most available space" strategy might be too simplistic.

src/umbp/doc/design-master-control-plane.md
```diff
@@ -0,0 +1,2049 @@
...
+
+    std::shared_ptr<grpc::Channel> channel_;
+    std::unique_ptr<UMBPMaster::Stub> stub_;
+
+    std::thread heartbeat_thread_;
+    std::atomic<bool> heartbeat_running_{false};
+    uint64_t heartbeat_interval_ms_ = 5000;
+
+    void HeartbeatLoop();
+};
```
maning00: how about renaming it to distinguish it from my `UMBPClient`? eg, `UMBPgRPCClient`, then as a member of `UMBPClient`?

src/umbp/doc/design-master-control-plane.md
```diff
@@ -0,0 +1,2049 @@
...
+│                                                                 │
+│    "keyA" → { locs: [Loc{nodeX, loc-a1}, Loc{nodeY, loc-a2}],  │
+│               metrics: {created: T0, accessed: T5, count: 42} } │
+│    "keyB" → { locs: [Loc{nodeX, loc-b1}],                       │
+│               metrics: {created: T1, accessed: T3, count: 7} }  │
+│                                                                 │
+│  registry_: ClientRegistry*  (for ownership tracking)           │
+└────────────────────────────────────────────────────────────────┘
+```
+
```
maning00, line 139: Does `BlockIndex` have a capacity limit?

src/umbp/doc/design-master-control-plane.md
TianDi101: Yes, at PoC stage we use a simple strategy for baseline. The ideal implementation is we expose a set of abstract interfaces so that new strategies (for different workloads) can be plug in easily.

src/umbp/doc/design-master-control-plane.md
TianDi101: Yes, we can do this

src/umbp/doc/design-master-control-plane.md
TianDi101: Good catch! Currently we don't enforce capacity limit to this in-memory BlockIndex. Since we only store metadata in BlockIndex, I would believe that the real data storage get exhausted earlier than metadata storage.

But we will see how it performs when we scale to very large number of nodes, I guess we probably need a independent persistent kv storage to manage metadata, such as Redis.

PR #250, opened 2026-04-03
fix(packaging): rename _jit_sources to _jit-sources to suppress setuptools warnings

isytwu:
## Summary
- Rename `_jit_sources/` to `_jit-sources/` so setuptools no longer treats bundled C++/HIP source directories as implicit namespace packages
- Add `umbp_master` and `spdk_proxy` build artifacts to `.gitignore`

## Details
Since PR #182 introduced `_copy_jit_sources()`, every `pip install -e` produces dozens of "Package would be ignored" warnings from setuptools. The root cause is that `_jit_sources` is a valid Python identifier, so Python 3.3+ namespace package rules make every subdirectory appear importable. Using a hyphenated name (`_jit-sources`) breaks `str.isidentifier()`, which is exactly what setuptools checks.

## Test plan
- [x] `BUILD_UMBP=ON BUILD_TESTS=ON pip install -e . --no-build-isolation -v` produces no "Package would be ignored" warnings
- [x] `python -c "from mori.jit.config import get_mori_source_root; print(get_mori_source_root())"` still resolves correctly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,8 +11,10 @@ dist/*
 .pytest_cache
 __pycache__
 python/mori/_version.py
-python/mori/_jit_sources/
+python/mori/_jit-sources/
 python/mori/*.so
+python/mori/umbp_master
+python/mori/spdk_proxy
 
 .vscode
 *.log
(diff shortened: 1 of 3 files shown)


jhchouuu (commented):
LGTM

PR #257, opened 2026-04-10
Refactor(umbp): introduce IUMBPClient interface for dual-scheme architecture

isytwu:
## Summary

- Introduce `IUMBPClient` ABC and `CreateUMBPClient()` factory for dual-scheme UMBP architecture (Standalone / Distributed)
- Implement `StandaloneClient` extracted from old `UMBPClient`, all PoolClient code removed
- Extract shared I/O drivers (POSIX / io_uring / SPDK) to `io/` and `spdk/`
- Decouple `umbp_core` from `umbp_common` — standalone builds no longer depend on gRPC/Protobuf
- pybind uses factory pattern, Python class name and method names unchanged; sglang `umbp_store.py` requires zero modifications

## Design

See `src/umbp/doc/design-umbp-dual-scheme-refactor.md` for full design document.

**Why ABC + Factory**: Standalone (client-local DRAM+SSD) and Distributed (master-led routing, similar to Mooncake) have fundamentally different control flows — not "base + hooks", but two independent execution paths behind a shared interface.

**Factory behavior**: currently always returns `StandaloneClient`. `config.distributed` is reserved for future `DistributedClient` (see `src/umbp/doc/handoff-next-steps.md`).

## Test plan

- [x] Full build passes (`BUILD_UMBP=ON`, zero errors)
- [x] All UMBP test/benchmark targets compile and link
- [x] pybind module compiles
- [ ] Run `test_umbp_local_client`, `test_umbp_e2e`, `test_follower_mode`, `test_prefix_aware_eviction`
- [ ] Verify sglang integration with `--hicache-storage-backend umbp`


TianDi101:
@isytwu Thanks for PR, I have two questions:

1. There're two storage folders, one inside umbp/local and one at the same level with umbp/local, how to understand the differences between them? 
2. Do you plan to put DistributedClient in local folder (same level with StandaloneClient) or distributed folder? I'd suggest the former.

isytwu:
> @isytwu Thanks for PR, I have two questions:
> 
> 1. There're two storage folders, one inside umbp/local and one at the same level with umbp/local, how to understand the differences between them?
> 2. Do you plan to put DistributedClient in local folder (same level with StandaloneClient) or distributed folder? I'd suggest the former.

1. storage/ - shared IO layer of posix/io_uring/spdk. Used by both standalone and distributed.
local/storage/ - standalone-specific dram/ssd tier manager. I'll rename local/storage/ to local/tiers/ to make the distinction clearer. 

2. I plan to put it in distributed/.  DistributedClient depends on gRPC/RDMA modules that live in distributed/. Placing it in local/ would create a reverse dependency.

PR #269, opened 2026-04-17
perf(ep): split AUTO launch config for IntraNode dispatch and combine

inkcherry:
## Summary

IntraNode dispatch and combine kernels have different optimal launch configurations on MI355X, but Mori's current `AUTO` mode applies a single `(block_num, rdma_block_num, warp_per_block)` tuple to both. This forces a compromise that costs measurable performance on whichever side gets the suboptimal value.

This PR splits the `AUTO` config into independent dispatch and combine tuples, plumbs an `is_dispatch` flag through `_resolve_launch_params` / `get_launch_config`, and routes the combine call sites accordingly. InterNode kernels keep the historical single-config behavior (no separate measurement available yet).

## Motivation

Profiling on MI355X shows the two kernels are bound by different resources:

- **Dispatch** is XGMI write-bandwidth bound and benefits from more blocks (≥256) to saturate the link.
- **Combine** is shared-memory / per-block work bound and prefers fewer, fatter blocks (~128).

A single shared tuple cannot satisfy both. The same observation holds across EP2 / EP4 / EP8 in our measurements (see below).

## Change

For IntraNode (`IntraNodeV1`), `AUTO` mode now uses:

| kernel   | new          | old          |
|----------|--------------|--------------|
| dispatch | `(256, 0, 16)` | `(128, 0, 16)` |
| combine  | `(128, 0, 16)` | `(128, 0, 16)` (unchanged) |

`InterNodeV1` and `InterNodeV1LL` keep their original single tuple for both sides — we have not measured them yet and explicitly do not want to change behavior there.

`MANUAL` mode is completely unaffected: SGLang and other consumers that pass explicit `block_num` / `warp_per_block` see no change.

## Implementation notes

- `EpDispatchCombineOp.__init__` now resolves two tuples in `AUTO` mode and stores them as `self.auto_*` (dispatch) and `self.auto_combine_*` (combine).
- `_resolve_launch_params(..., is_dispatch=True)` and `get_launch_config(is_dispatch=True, ...)` pick the right tuple.
- All combine entry points (`combine`, `combine_recv`, `combine_standard_moe`, `convert_combine_input`) pass `is_dispatch=False`.
- `MANUAL` mode keeps every `auto_*` field as `None`, so the resolver falls through to the user-supplied values exactly as before.

## Performance

All numbers from `tests/python/ops/bench_dispatch_combine.py`, FP4, hidden=7168, top-8, scale_dim=224, on a single MI355X node, `MORI_EP_LAUNCH_CONFIG_MODE=AUTO`, 3 runs each.

**EP2, 12k tokens/rank** (primary tuning target):

| kernel   | before (us avg / min) | after (us avg / min) | delta |
|----------|-----------------------|----------------------|-------|
| dispatch | 1087 / 1069 | 1037 / 1018 | **−4.6%** |
| combine  |  806 /  791 |  806 /  791 | unchanged |

**EP4 / EP8 regression check** (combine config unchanged → only dispatch can move):

| config         | dispatch before → after | combine before → after |
|----------------|-------------------------|------------------------|
| EP4, 12k/rank  | 353 → 334 us (**−5.4%**) | 295 → 293 us (−0.7%) |
| EP4,  4k/rank  | 164 → 162 us (**−1.0%**) | 159 → 159 us (0%)    |
| EP8,  6k/rank  | 284 → 280 us (**−1.5%**) | 271 → 271 us (0%)    |

No regression on any EP size; combine is exactly equal because its `AUTO` value did not change.

## Test plan

- [x] `pytest tests/python/ops/test_dispatch_combine_intranode.py` under `MORI_EP_LAUNCH_CONFIG_MODE=AUTO` → 160 passed, 160 skipped
- [x] `pytest tests/python/ops/test_dispatch_combine_intranode.py` under `MORI_EP_LAUNCH_CONFIG_MODE=MANUAL` → 160 passed, 160 skipped
- [x] EP2 / EP4 / EP8 perf measured (table above), no regression on any topology
- [ ] InterNode kernels: behavior unchanged by construction; not re-benchmarked in this PR

## Risk / scope

- IntraNode `AUTO`: dispatch defaults change `128 → 256` blocks. Validated on EP2/EP4/EP8 above.
- IntraNode `MANUAL`: unchanged.
- InterNode (V1, V1LL): unchanged.
- Public API: `_resolve_launch_params` and `get_launch_config` gain an `is_dispatch` keyword-only arg with a default of `True`, so existing callers are unaffected.

---

### DIFF
--- a/python/mori/ops/dispatch_combine.py
+++ b/python/mori/ops/dispatch_combine.py
@@ -265,36 +265,47 @@ class EpDispatchCombineOp:
 
         self.launch_config_mode = os.environ.get("MORI_EP_LAUNCH_CONFIG_MODE", "MANUAL")
         if self.launch_config_mode == "AUTO":
+            # Each entry is (block_num, rdma_block_num, warp_per_block).
+            # IntraNode dispatch and combine have different optimal launch
+            # configurations: dispatch is XGMI write-bandwidth bound and benefits
+            # from more blocks (≥256), while combine is shared-memory / per-block
+            # work bound and prefers fewer, fatter blocks (~128). Keeping a single
+            # config for both leaves performance on the table for whichever side
+            # gets the suboptimal value. We therefore split the AUTO config into
+            # a dispatch tuple and a combine tuple. Inter-node kernels keep the
+            # historical single config for both sides (no measured benefit yet).
             if (
                 self.config.kernel_type.value
                 == EpDispatchCombineKernelType.InterNodeV1.value
             ):
-                (
-                    self.auto_block_num,
-                    self.auto_rdma_block_num,
-                    self.auto_warp_per_block,
-                ) = (96, 64, 8)
+                dispatch_cfg = combine_cfg = (96, 64, 8)
             elif (
                 self.config.kernel_type.value
                 == EpDispatchCombineKernelType.InterNodeV1LL.value
             ):
-                (
-                    self.auto_block_num,
-                    self.auto_rdma_block_num,
-                    self.auto_warp_per_block,
-                ) = (256, 128, 8)
+                dispatch_cfg = combine_cfg = (256, 128, 8)
             else:
-                (
-                    self.auto_block_num,
-                    self.auto_rdma_block_num,
-                    self.auto_warp_per_block,
-                ) = (128, 0, 16)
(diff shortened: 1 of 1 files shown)


inkcherry:
move to https://github.com/ROCm/mori/pull/271 close.

PR #280, opened 2026-04-20
Flydsl shmem

TennyWang1223:
## Motivation

  Add FlyDSL integration for the mori shmem device API so that FlyDSL users can call mori shmem device functions (e.g. my_pe, putmem_nbi_warp, quiet_thread_pe) directly inside @flyc.kernel, enabling shmem-based GPU programming on top of FlyDSL.

  FlyDSL targets ABI 600 (cov=6) while Triton uses cov=5. The existing bitcode build/cache pipeline cannot serve both ABIs at the same time, so cache lookup and compilation must be partitioned by code-object-version.

## Technical Details

  - New python/mori/ir/flydsl/ subpackage:
    - ops.py: auto-generates FlyDSL ExternFunction wrappers from MORI_DEVICE_FUNCTIONS metadata. Each entry maps 1:1 to a C symbol in libmori_shmem_device.bc and becomes callable inside @flyc.kernel.
    - runtime.py: exposes get_bitcode_path(), install_hook(), and install_jit_hook(). install_hook registers a FlyDSL post-compile callback that invokes shmem_module_init on the freshly compiled HIP module so the globalGpuStates device symbol gets initialized;
  install_jit_hook does the equivalent for flyc.jit module loads.
    - __init__.py: re-exports the ops and runtime entry points.
  - Per-cov bitcode build & cache:
    - python/mori/jit/cache.py: get_cache_dir gains a keyword-only cov argument and appends _cov<N> to the cache directory name so cov5 / cov6 artifacts no longer collide.
    - python/mori/jit/core.py: ensure_bitcode, _build_bitcode, and _hipcc_device_bc accept cov (default 5 for backward compatibility) and forward it to hipcc as -mcode-object-version=<cov>.
    - python/mori/ir/bitcode.py: find_bitcode / get_bitcode_path accept cov; the module-level cache becomes a {cov: path} dict. When cov is explicitly requested, pre-built bitcode is skipped (only the JIT path is valid); when cov is None, legacy behavior is preserved.

## Test Plan

  - Build a @flyc.kernel that invokes shmem ops from mori.ir.flydsl (my_pe, n_pes, putmem_nbi_warp, quiet_thread_pe, etc.) and exercise cross-PE communication.
  - Verify that after install_hook() / install_jit_hook(), globalGpuStates is initialized on the loaded module and kernels run correctly.
  - In the same environment, trigger JIT compilation from both Triton (cov=5) and FlyDSL (cov=6); confirm the two bitcode caches live in separate directories and do not overwrite each other.
  - Regress the existing Triton / default-cov path to confirm find_bitcode() and the pre-built bitcode lookup still work.

## Submission Checklist

### DIFF
--- a/python/mori/ir/bitcode.py
+++ b/python/mori/ir/bitcode.py
@@ -37,10 +37,11 @@ import os
 from pathlib import Path
 
 _BC_FILENAME = "libmori_shmem_device.bc"
-_cached_path: str | None = None
+# Per-cov caches: {cov: path}
+_cached_paths: dict[int | None, str] = {}
 
 
-def _find_jit_cached_bitcode() -> str | None:
+def _find_jit_cached_bitcode(*, cov: int | None = None) -> str | None:
     """Look for a previously JIT-compiled bitcode in the cache directory."""
     try:
         from mori.jit.config import (
@@ -62,7 +63,7 @@ def _find_jit_cached_bitcode() -> str | None:
             mori_root / "include" / "mori" / "shmem",
             mori_root / "include" / "mori" / "core",
         ]
-        cache_dir = get_cache_dir(cfg.arch, source_paths, nic)
+        cache_dir = get_cache_dir(cfg.arch, source_paths, nic, cov=cov)
         bc_path = cache_dir / _BC_FILENAME
         if bc_path.is_file():
             return str(bc_path)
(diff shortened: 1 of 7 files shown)


jhchouuu:
@TennyWang1223  help fix pre-commit?

jhchouuu (approved):
LGTM

PR #298, opened 2026-04-29
fix(rdma): bnxt PollCq/QP bugs and ibverbs path_mtu override

jaeyoun98:
## Motivation

These issues came up while bringing up MORI on node with 8x MI300X and 8x Broadcom Thor2 NICs (400G, `bnxt_re`).
Three of them are latent bnxt provider bugs that surface on specific code paths or fabric topologies, and the last is a small ibverbs knob for switched fabrics. Sharing upstream in case they're useful to others on similar setups.

## Technical Details

bnxt provider:
- PollCq epoch flag: the bitmask was applied before the shift, producing an incorrect flag value. Isolate the LSB of the wrap count first (`& 0x1`), then shift to `BNXT_RE_FLAG_EPOCH_HEAD_SHIFT`.
- `BnxtQpContainer` ctor: the RQ backing buffer was allocated with `qpMemInfo.sq_len`. Use `rq_len` instead.
- `ModifyInit2Rtr`: `hop_limit = 1` prevents RoCE packets from crossing any router. Use 255 (max) to allow routed fabrics; no effect on flat L2.

ibverbs provider:
- Add `MORI_IB_PATH_MTU` env var override for `attr.path_mtu` in `ConnectEndpoint`, for switched fabrics that negotiate a lower `active_mtu` than the port supports. Accepted values are 256, 512, 1024, 2048, 4096. When unset, preserves current behavior (use `portAttr->active_mtu`).

## Test Plan

These fixes were developed and verified on:
- A 2-node cluster, each node equipped with 8x MI300X and 8x Broadcom Thor2 400G NICs (`bnxt_re`)
- A 10-node cluster, also each node equipped with 8x MI300X and 8x Broadcom Thor2 400G NICs (`bnxt_re`)

We use these as a precondition for our vLLM P/D disaggregated serving stack with the MoRI-IO connector. Without these fixes, MoRI-IO failed at startup.

## Test Result

- vLLM 1P1D (TP=8) disaggregated serving comes up and runs end-to-end with MoRI-IO over Thor2 with these fixes applied.
- Various xPyD setups succeeded in bring-up including 1P3D, 1P4D, etc.

We have not done isolated per-fix regression tests; the fixes were validated as a set during cluster bring-up. Would be happy to provide more targeted reproducers if useful for review.

## Submission Checklist

Signed-off-by: Chaemin Lim <chaemin.lim@mangoboost.io>

### DIFF
--- a/include/mori/core/transport/rdma/providers/bnxt/bnxt_device_primitives.hpp
+++ b/include/mori/core/transport/rdma/providers/bnxt/bnxt_device_primitives.hpp
@@ -741,7 +741,7 @@ inline __device__ int PollCq<ProviderType::BNXT>(WorkQueueHandle& wqHandle,
 template <>
 inline __device__ void UpdateCqDbrRecord<ProviderType::BNXT>(CompletionQueueHandle& cq,
                                                              uint32_t consIdx) {
-  uint8_t flags = ((consIdx + 1) / cq.cqeNum) & (1UL << BNXT_RE_FLAG_EPOCH_HEAD_SHIFT);
+  uint8_t flags = (((consIdx + 1) / cq.cqeNum) & 0x1) << BNXT_RE_FLAG_EPOCH_HEAD_SHIFT;
   uint32_t epoch = (flags & BNXT_RE_FLAG_EPOCH_TAIL_MASK) << BNXT_RE_DB_EPOCH_TAIL_SHIFT;
   // uint64_t dbrVal = bnxt_re_init_db_hdr(((consIdx + 1) | epoch), 0, flags, BNXT_RE_QUE_TYPE_CQ);
   uint64_t dbrVal =
--- a/src/application/transport/rdma/providers/bnxt/bnxt.cpp
+++ b/src/application/transport/rdma/providers/bnxt/bnxt.cpp
@@ -227,8 +227,8 @@ BnxtQpContainer::BnxtQpContainer(ibv_context* context, const RdmaEndpointConfig&
         hipExtMallocWithFlags(&rqUmemAddr, qpMemInfo.rq_len, hipDeviceMallocUncached));
     HIP_RUNTIME_CHECK(hipMemset(rqUmemAddr, 0, qpMemInfo.rq_len));
   } else {
-    err = posix_memalign(&rqUmemAddr, config.alignment, qpMemInfo.sq_len);
-    memset(rqUmemAddr, 0, qpMemInfo.sq_len);
+    err = posix_memalign(&rqUmemAddr, config.alignment, qpMemInfo.rq_len);
+    memset(rqUmemAddr, 0, qpMemInfo.rq_len);
     assert(!err);
   }
   qpMemInfo.rq_va = reinterpret_cast<uint64_t>(rqUmemAddr);
@@ -395,7 +395,7 @@ void BnxtQpContainer::ModifyInit2Rtr(const RdmaEndpointHandle& local_handle,
 
   memcpy(&attr.ah_attr.grh.dgid, remote_handle.eth.gid, 16);
   attr.ah_attr.grh.sgid_index = local_handle.eth.gidIdx;
-  attr.ah_attr.grh.hop_limit = 1;
+  attr.ah_attr.grh.hop_limit = 255;
   attr.ah_attr.is_global = 1;
   attr.ah_attr.port_num = config.portId;
   attr.ah_attr.sl = ReadRdmaServiceLevelEnv().value_or(1);
--- a/src/application/transport/rdma/providers/ibverbs/ibverbs.cpp
+++ b/src/application/transport/rdma/providers/ibverbs/ibverbs.cpp
@@ -127,7 +127,28 @@ void IBVerbsDeviceContext::ConnectEndpoint(const RdmaEndpointHandle& local,
   assert(portAttr);
   // RTR
   attr.qp_state = IBV_QPS_RTR;
-  attr.path_mtu = portAttr->active_mtu;
+  {
+    ibv_mtu path_mtu = portAttr->active_mtu;
+    const char* envMtu = std::getenv("MORI_IB_PATH_MTU");
+    if (envMtu != nullptr) {
+      int mtuBytes = std::atoi(envMtu);
+      if (mtuBytes == 256)
+        path_mtu = IBV_MTU_256;
+      else if (mtuBytes == 512)
+        path_mtu = IBV_MTU_512;
+      else if (mtuBytes == 1024)
+        path_mtu = IBV_MTU_1024;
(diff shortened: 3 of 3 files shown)


jhchouuu:
Hi @jaeyoun98 @chaeminlim-mb Could you please tell me which version of brcm driver/userlib  you are using? Because the current brcm one-cqe mode allows cqe overwrite the first position of cq, so the UpdateCqDbrRecord operation is actually a redundant operation now. This should not cause any problems...

chaeminlim-mb:
Hi @jhchouuu, thanks for the heads-up on the redundancy.

  ### Our environment

  - Broadcom Thor2 NICs
  - Userspace: Broadcom rdma-core v58 (`libbnxt_re-rdmav57.so`,
    `236.1.155.0`, built 2026-04-13)
  - Kernel ABI: `bnxt_re` v8

  ### Re-examining each hunk

  I want to be upfront — these three changes were collected during our
  Thor2 bring-up and the evidence behind them isn't equally strong.

  **1. `UpdateCqDbrRecord` operator order**

  This one came from a code-review reading of the bit operation rather than from a symptom we can point to on our stack. The original
  ```cpp
  uint8_t flags = ((consIdx + 1) / cq.cqeNum) & (1UL << SHIFT);
```
  selects bit SHIFT of the wrap count, which seemed inconsistent with  the apparent intent of putting the wrap-count LSB into the SHIFT  position; the (... & 0x1) << SHIFT form makes that explicit. But as  you note, if the doorbell update is effectively a no-op on current  one-cqe mode then the distinction doesn't materialize on v58 / ABI v8.  We're happy to drop this hunk if you'd prefer the PR without it.

  2. BnxtQpContainer ctor — sq_len vs rq_len for the RQ buffer

  The block (bnxt.cpp ~227–232):
```
  err = posix_memalign(&rqUmemAddr, config.alignment, qpMemInfo.sq_len);
  memset(rqUmemAddr, 0, qpMemInfo.sq_len);
  // ...
  qpMemInfo.rq_va = reinterpret_cast<uint64_t>(rqUmemAddr);
```
  The variable name (rqUmemAddr) and the immediately following  qpMemInfo.rq_va = ... read like an RQ allocation, so sizing it with  sq_len looked like a copy-paste typo when we passed through the  file. We don't have a concrete failure traceable to it — in our  deployment the SQ and RQ depths haven't diverged enough for it to  matter — so this is more a tidy-up than a bug fix. Also fine to drop or split out if you'd rather review it separately.

  3. ModifyInit2Rtr — hop_limit = 1 → 255

  This is the one we actually rely on. On a flat L2 the connection  establishes fine, but on our multi-node Thor PD setup where prefill  and decode end up on different subnets, RoCEv2 packets get  TTL-dropped at the first router with hop_limit = 1 and the data  path silently stops. Setting hop_limit = 255 (the max) is the  typical default for routed RoCE; flat-L2 topologies aren't affected.

  If you'd like, I can:
  - drop (1) and (2), keep just the hop_limit change as a small,  symptom-driven fix, or
  - leave the PR as is if you're OK with (1)/(2) as defensive  consistency tweaks, or
  - split into separate PRs

Thank you!

**Comment by jhchouuu** [comment]
Hi @chaeminlim-mb thanks for your information
The information I have is that the 236.1 version may not provide good support for IBGDA. There might be random hangs. It is recommended to use the stable 235.2 version.

File: include/mori/core/transport/rdma/providers/bnxt/bnxt_device_primitives.hpp [hunk 1]
```diff
@@ -741,7 +741,7 @@ inline __device__ int PollCq<ProviderType::BNXT>(WorkQueueHandle& wqHandle,
 template <>
 inline __device__ void UpdateCqDbrRecord<ProviderType::BNXT>(CompletionQueueHandle& cq,
                                                              uint32_t consIdx) {
-  uint8_t flags = ((consIdx + 1) / cq.cqeNum) & (1UL << BNXT_RE_FLAG_EPOCH_HEAD_SHIFT);
+  uint8_t flags = (((consIdx + 1) / cq.cqeNum) & 0x1) << BNXT_RE_FLAG_EPOCH_HEAD_SHIFT;
```
jhchouuu, line 744: I have reviewed the updated spec from BRCM. There are some differences here. I'm currently checking them. Could you please refrain from making any changes for now? I'll make the corrections after I've verified them.

src/application/transport/rdma/providers/bnxt/bnxt.cpp
```diff
@@ -227,8 +227,8 @@ BnxtQpContainer::BnxtQpContainer(ibv_context* context, const RdmaEndpointConfig&
         hipExtMallocWithFlags(&rqUmemAddr, qpMemInfo.rq_len, hipDeviceMallocUncached));
     HIP_RUNTIME_CHECK(hipMemset(rqUmemAddr, 0, qpMemInfo.rq_len));
   } else {
-    err = posix_memalign(&rqUmemAddr, config.alignment, qpMemInfo.sq_len);
-    memset(rqUmemAddr, 0, qpMemInfo.sq_len);
+    err = posix_memalign(&rqUmemAddr, config.alignment, qpMemInfo.rq_len);
```
jhchouuu, line 230: our typo, although the two lengths are currently equal. thanks

src/application/transport/rdma/providers/ibverbs/ibverbs.cpp
```diff
@@ -127,7 +127,20 @@ void IBVerbsDeviceContext::ConnectEndpoint(const RdmaEndpointHandle& local,
...
+  {
+    ibv_mtu path_mtu = portAttr->active_mtu;
+    const char* envMtu = std::getenv("MORI_IB_PATH_MTU");
+    if (envMtu != nullptr) {
+      int mtuBytes = std::atoi(envMtu);
+      if      (mtuBytes == 256)  path_mtu = IBV_MTU_256;
+      else if (mtuBytes == 512)  path_mtu = IBV_MTU_512;
+      else if (mtuBytes == 1024) path_mtu = IBV_MTU_1024;
+      else if (mtuBytes == 2048) path_mtu = IBV_MTU_2048;
+      else if (mtuBytes == 4096) path_mtu = IBV_MTU_4096;
```
jhchouuu: Add illegal verification?

jhchouuu (commented):
and pls use pre-commit to format code

src/application/transport/rdma/providers/ibverbs/ibverbs.cpp
jaeyoun98: I added a minimal `else` branch to display warning message for any values outside the supported set.

jhchouuu: Thanks!

PR #300, opened 2026-04-30
warn when AsyncLL runs without SDMA

inkcherry:
## Motivation

- Add a warning when `KernelType::AsyncLL` is selected while `MORI_ENABLE_SDMA` is not enabled.
- AsyncLL without SDMA still uses compute units for communication, which limits overlap benefits and can severely degrade performance.
- Recommend using a non-AsyncLL kernel path or enabling SDMA with `MORI_ENABLE_SDMA=1`.

## Submission Checklist


inkcherry:
My end-to-end test: the non-SDMA async kernel seems meaningless, as cu-free is the main feature of async. right? Should I upgrade it to raise an error? @isytwu 


Mean TPOT	Total tok/s	E2E latency	Bench duration
AsyncLL w/p MORI_ENABLE_SDMA 	3062.76 ms	685.59	192.95 s	~13 min
AsyncLL w MORI_ENABLE_SDMA=1 	80.11 ms	24199.05	5.05 s	21.7 s

isytwu:
LGTM.

PR #307, opened 2026-05-08
(JIT) Detect CCQE support in runtime instead of compiling

QizhouZhang97:
## Motivation

MORI must support both CCQE and non-CCQE ionic deployments from a single
binary. The previous build-time `IONIC_CCQE` macro forced a recompile per
host and broke when `libionic.so` and `mori_application.so` were out of sync.

This change moves the decision to runtime: dlsym `ionic_dv_create_cq_ex` at
CQ creation time — if the symbol is present and firmware build ≥ 58, use
CCQE; otherwise fall back to `ibv_create_cq_ex`. A new environment variable
`MORI_DISABLE_IONIC_CCQE=1` overrides the runtime detection and forces the
non-CCQE path even on hardware that supports it.

## Issues resolved

- Removed the build-time `IONIC_CCQE` macro and replaced it with a runtime
  capability check (`IsCcqeSupported`).
- Fixed a bug in the ionic GDA implementation where only the non-CCQE poll-CQ
  path was implemented; removed the now-redundant duplicate poll-CQ code.

## Tests

1. `mi355_n06_21` with a77 firmware (CCQE supported) — verified CCQE path is
   taken and `MORI_DISABLE_IONIC_CCQE=1` correctly forces fallback.
2. a94 firmware on AINIC team machines (CCQE)

### DIFF
--- a/include/mori/core/transport/rdma/providers/ionic/ionic_device_primitives.hpp
+++ b/include/mori/core/transport/rdma/providers/ionic/ionic_device_primitives.hpp
@@ -622,8 +622,6 @@ inline __device__ int PollCqOnce2(WorkQueueHandle& wqHandle, CompletionQueueHand
   volatile struct ionic_v1_cqe* cqe = reinterpret_cast<ionic_v1_cqe*>(cqeAddr);
   uint32_t old, msn = HTOBE32(cqe->send.msg_msn);
 
-  printf("poll cccqe2\n");
-
   consIdx = wqHandle.dbTouchIdx;
 
   // MORI_PRINTF("ABH %s:%d here cons %#x msn %#x\n", __func__, __LINE__, consIdx, msn);
@@ -644,8 +642,6 @@ inline __device__ int PollCqOnce2(WorkQueueHandle& wqHandle, CompletionQueueHand
                                   uint32_t consIdx) {
   uint32_t my_logical_lane_id = get_active_lane_num(activemask);
   uint32_t my_cq_pos = cqHandle.cq_consumer + my_logical_lane_id;
-
-  printf("poll nonccqe 2\n");
 
   uint32_t cqeIdx = my_cq_pos & (cqeNum - 1);
   char* Addr = reinterpret_cast<char*>(cqeAddr) + (cqeIdx * sizeof(struct ionic_v1_cqe));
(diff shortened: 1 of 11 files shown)


src/application/transport/rdma/providers/ionic/ionic.cpp
```diff
@@ -40,6 +44,45 @@ namespace application {
 /*                                        Device Attributes                                       */
 /* ---------------------------------------------------------------------------------------------- */
 
+namespace {
+
+// Minimum firmware build number required for CCQE support (e.g. "1.117.5-a58" → 58).
```
jhchouuu: 可能需要判断一下1.117.5，因为似乎以前就有1.117.1这种版本，只判断58可能不够golden

include/mori/core/transport/rdma/providers/ionic/ionic_fw.h
```diff
@@ -207,7 +243,7 @@ enum ionic_v1_cqe_qtf_bits {
   IONIC_V1_CQE_TYPE_RECV = 1,
   IONIC_V1_CQE_TYPE_SEND_MSN = 2,
   IONIC_V1_CQE_TYPE_SEND_NPG = 3,
-  IONIC_V1_CQE_TYPE_RECV_INDIR = 4,
+  c = 4,
```
jhchouuu: c = 4？ 是不是弄错了

include/mori/core/transport/rdma/providers/ionic/ionic_device_primitives.hpp
```diff
@@ -591,217 +512,13 @@ inline __device__ int PollCq<ProviderType::PSD>(void* cqAddr, uint32_t cqeNum, u
     const uint32_t msn = BE32TOH(cqe->send.msg_msn) & 0xFFFF;
     const uint8_t error = IonicHandleErrorCqe(status);
```
jhchouuu: 要不拿个else框一下，不然下面的实际上可能会编进去...

src/application/transport/rdma/providers/ionic/ionic.cpp
QizhouZhang97: done

include/mori/core/transport/rdma/providers/ionic/ionic_fw.h
QizhouZhang97: done

include/mori/core/transport/rdma/providers/ionic/ionic_device_primitives.hpp
QizhouZhang97: done

PR #308, opened 2026-05-08
feat(shmem): add MORI_MULTITHREAD_SUPPORT for SPMT

jhchouuu:
## Summary
Add single-process multi-thread (SPMT) support gated by `MORI_MULTITHREAD_SUPPORT`, enabling multiple GPU threads within one process to drive MORI EP independently for the JAX SPMD model.
## Key Changes
- **Per-GPU state isolation**: make ShmemStates, KernelRegistry, and XLA FFI handle cache per-GPU keyed by `hipGetDevice()`, with ScopedDevice RAII guard in FFI handlers
- **Same-process P2P memory**: detect same-process peers via PID allgather, skip `hipIpcOpenMemHandle` and use direct VA + `hipDeviceEnablePeerAccess`, with complete SDMA resource cleanup on deregistration
- **SDMA channel isolation**: key AnvilLib `sdma_channels_` by `(srcDeviceId, dstDeviceId)` pair with mutex for thread-safe access
- **AsyncLL kernel launch fix**: correct C++ FFI path to use split kernel names matching `ep_async_ll.hip` (previous combined names were dead code)
- **Python / pybind**: GIL release on blocking shmem APIs, remove duplicate pybind defs, replace torch import with ctypes `hipGetDevice` for JAX compatibility
- **Safety**: guard `~EpDispatchCombineHandle` against use-after-finalize, clear pending GPU errors before `FinalizeGpuStates`, centralize `kMaxGpusPerNode`
## Testing
- Torch multi-process IntraNode: passed (no regression)
- Torch multi-process AsyncLL IBGDA: 68 passed
- Torch multi-process AsyncLL SDMA: 68 passed

Co-Authored-By: @pemeliya


src/pybind/pybind_xla_ffi_ops.cpp
```diff
@@ -280,14 +331,27 @@ ErrorOr<std::unique_ptr<EpDispatchCombineState>> EpDispatchCombineInstantiate(
...
+  auto cfg = EpDispatchCombineConfig::FromPackedI32Array(key.data(), key.size());
+
+#ifdef MORI_MULTITHREAD_SUPPORT
+  // SPMT: XLA FFI handlers run on framework worker threads where
+  // hipGetDevice() does NOT match the rank's device. Look up the device
+  // recorded at ShmemInit time and bind it on this thread (RAII-restored on
+  // exit so XLA's other state isn't disturbed) before any HIP call. This
+  // ensures GetHandleCacheSlot() and ShmemStatesSingleton::GetInstance()
+  // (both keyed by hipGetDevice()) hit the right slot.
+  ScopedDevice _dev_guard(mori::shmem::ShmemStatesSingleton::GetDeviceByRank(cfg.rank));
```
draganmladjenovic, line 345: Has this been checked? I was under impression that xla ensures that the "context" device is set to the one the thread serves.

include/mori/utils/limits.hpp
```diff
@@ -0,0 +1,33 @@
...
+#pragma once
+
+// Tiny header so any TU (device or host) can size per-GPU arrays without
+// pulling in shmem/ops headers. Bump this when supporting nodes with > 8 GPUs
+// (e.g. future MI400 platforms) — every per-GPU array sized by this constant
+// will pick up the new value automatically.
+
+namespace mori {
+
+inline constexpr int kMaxGpusPerNode = 8;
```
draganmladjenovic, line 31: I think with CPX it is 32. Not sure if it can be split even more than that.

i-chaochen: Yes, we will have ~32~ 72 GPUs at rack level. It's best to not hardcode this or we should get this max number of GPU from the build script?

jhchouuu: Yes, CPX and rack level will have more than 8 GPUs. We have already tried this on CPX, but it was limited to a single card before... Additionally, the rack level is also included in our plan...  
So currently, kMaxGpusPerNode equals to 8...

src/pybind/pybind_xla_ffi_ops.cpp
jhchouuu: Yes, and could you please check it again? Maybe my understanding is incorrect.

pemeliya: I think the issue here is that, we cannot initialize MORI in a correct way (when SIMT is on), unless MORI is integrated to XLA. This is because XLA is managing GPU threads, so ShmemInit() shall be called by XLA intentally when GPU communicators are created

pemeliya: In my original branch, I added some hacky way for MORI initialization to see if it generally works

src/application/memory/symmetric_memory.cpp
pemeliya, line 280: I have tried running my XLA SPMT test - and getting the issue here:
```
[/data/mori/src/application/memory/symmetric_memory.cpp:251] hip failed with invalid device context
```
I think we shall guard this block by SameProcessP2P(i)

jhchouuu: your XLA SPMT test is related to SDMA? I think I should add a verification for the SDMA path to my TODO list... I haven't verified SDMA...

pemeliya: yep, I use simple SDMA-based collectives there. But I think RegisterSymmMemObj can be used also without SDMA transport

jhchouuu: OK. I'm continuing my work on SDMA. I will verify the SDMA path through the async_ll EP kernel.

include/mori/application/context/context.hpp
```diff
@@ -39,10 +39,10 @@ class Context {
   int LocalRank() const { return bootNet.GetLocalRank(); }
   int WorldSize() const { return bootNet.GetWorldSize(); }
   int LocalRankInNode() const { return rankInNode; }
-  std::string HostName() const;
+  std::string HostName() const { return myHostname; }
```
pemeliya: NIT: I think we can return 'const std::string&' here since myHostname is a class member

pemeliya:
I have tested it with XLA PoC branch, everything works fine ! Also a simple all-gather SDMA perf test works fine with SPMT:
```
tf-docker /data/mori > Rank 0 / 1 starting (chunk_bytes=339738624)
Rank 0: wrote UniqueId to /data/mori/allgather_test_uid.bin
---------------------- Global rank 0 deviceID 0 worldSize 4
---------------------- Global rank 1 deviceID 1 worldSize 4
---------------------- Global rank 2 deviceID 2 worldSize 4
---------------------- Global rank 3 deviceID 3 worldSize 4
allgather_test: 4 PEs rank: 0, 339738624 bytes/PE (331776 KB), 1358954496 bytes total
3: baseBuf: 0x737b35600000 baseBufObj: 0x737b35600000
1: baseBuf: 0x737a04800000 baseBufObj: 0x737a04800000
0: baseBuf: 0x737a69e00000 baseBufObj: 0x737a69e00000
2: baseBuf: 0x737acfa00000 baseBufObj: 0x737acfa00000
Rank 0: PASS (0 errors) | 5 warmup + 20 runs | avg 5.84 ms (232.577 GB/s) min 5.83 ms (233.010 GB/s) max 5.86 ms
Rank 3: PASS (0 errors) | 5 warmup + 20 runs | avg 5.84 ms (232.626 GB/s) min 5.83 ms (233.176 GB/s) max 5.85 ms
Rank 2: PASS (0 errors) | 5 warmup + 20 runs | avg 5.84 ms (232.651 GB/s) min 5.83 ms (233.224 GB/s) max 5.86 ms
Rank 1: PASS (0 errors) | 5 warmup + 20 runs | avg 5.84 ms (232.677 GB/s) min 5.83 ms (233.080 GB/s) max 5.85 ms
```

pemeliya (approved):

i-chaochen:
@pemeliya IIUC your XLA PoC branch is this one? https://github.com/ROCm/mori/tree/pemeliya/multi_threading_support

pemeliya:
> @pemeliya IIUC your XLA PoC branch is this one? https://github.com/ROCm/mori/tree/pemeliya/multi_threading_support

ah nope, this is my MORI PoC branch. My XLA PoC is: https://github.com/ROCm/xla/tree/pemeliya/mori_upstream2

PR #309, opened 2026-05-09
Refactor: decouple allocator from master and client

TianDi101:
## Submission Checklist


src/umbp/distributed/pool_client.cpp
```diff
@@ -317,1711 +350,887 @@ std::optional<std::pair<mori::io::MemoryDesc, size_t>> PoolClient::FindRegistere
...
-  // Hydrate + snapshot under a single peers_mutex_ window (mirror of
-  // PlanPeerForBatchPut).  Pages come from master via r.pages and are
-  // stored in plans[orig_idx].parsed_pages during Phase 1a.
-  struct PerItemSnapshot {
-    size_t orig_idx;
-    std::vector<mori::io::MemoryDesc> remote_mems;  // one per parsed_page
-  };
-  std::vector<PerItemSnapshot> snaps;
-  snaps.reserve(peer_item_indices.size());
+void PoolClient::ExecuteRemotePutTransfers(std::vector<RemotePutEntry>& entries,
```
isytwu, line 784: One BatchPut pipeline concern: the old path intentionally submitted RDMA first, then ran local memcpy / staging-backed work while RDMA was in flight, and waited at the end.

Here all staging entries are copied before BatchWrite is issued, so memcpy no longer overlaps with RDMA progress. Should we preserve that pipeline behavior?

src/umbp/distributed/pool_client.cpp
```diff
@@ -317,1711 +350,887 @@ std::optional<std::pair<mori::io::MemoryDesc, size_t>> PoolClient::FindRegistere
...
-              p.buffer_index, peer.peer_address, keys[orig_idx]);
-          ok = false;
-          break;
-        }
-        snap.remote_mems.push_back(peer.dram_memories[p.buffer_index]);
-      }
-      if (!ok) {
-        plans[orig_idx].kind = BatchPutItemPlan::Kind::REMOTE_ZC_PREP_FAILED;
-        continue;
+bool PoolClient::BuildRemotePutTransfers(std::vector<RemotePutEntry>& entries, PeerConnection& peer,
```
isytwu, line 719: If we copy all non-zero-copy entries into staging at once, the staging buffer may not be large enough for the whole peer group. Since staging memory is RDMA-registered, the previous serial fallback also let us keep this registered buffer small and reused, instead of sizing it for the whole batch. Should we preserve that fallback behavior?

src/umbp/distributed/pool_client.cpp
```diff
@@ -317,1711 +350,887 @@ std::optional<std::pair<mori::io::MemoryDesc, size_t>> PoolClient::FindRegistere
...
+  for (size_t i = 0; i < keys.size(); ++i) {
+    if (results[i]) continue;
+    results[i] = false;
   }
-  results = std::move(found);
   return results;
 }
 
-bool PoolClient::Get(const std::string& key, void* dst, size_t size) {
+std::vector<bool> PoolClient::BatchGet(const std::vector<std::string>& keys,
```
isytwu, line 534: BatchGet looks similar to BatchPut: the pipeline behavior and staging fallback may have the same issues here as well.

src/umbp/distributed/master/master_client.cpp
```diff
@@ -890,82 +297,112 @@ void MasterClient::HeartbeatLoop() {
...
+      }
+      // Mirror back so external callers get a consistent read.
+      current_capacities_ = caps;
+    } else {
+      std::lock_guard lock(caps_mutex_);
+      caps = current_capacities_;
     }
 
+    std::vector<KvEvent> events;
+    if (peer_alloc_ != nullptr) events = peer_alloc_->DrainPendingEvents();
```
isytwu, line 321: Heartbeat failure currently drops drained ADD/REMOVE events without retrying or re-queueing them, so a committed Put may never become visible to the master?

src/umbp/distributed/pool_client.cpp
TianDi101: Yes, good suggestion

PR #315, opened 2026-05-12
feat(EP): decouple fp8_blockwise combine scale_dim from user config

maning00:
## Summary

- Add a dedicated `fp8BlockwiseCombineScaleDim` on `EpDispatchCombineHandle`, plumbed through `EpDispatchCombineArgs<T>` / `ArgsRaw` to the kernel. Driven by a new env var `MORI_FP8_COMBINE_SCALE_DIM` (default `56`, matching `block_elems = 7168 / 56 = 128` for the `AccumNum=8 + VecBytes=8` dequant specialization).
- `EpDispatchCombineConfig.scaleDim` / `scaleTypeSize` keep their original semantics (caller-provided dispatch scales, e.g. FP4 input). fp8_blockwise combine no longer consults them.
- `shmemOutScalesMemObj` is allocated only when the caller provides user scales; `shmemInpScalesMemObj` is sized as `max(userScaleSize, fp8BlockwiseScaleSize)`. Pybind `get_dispatch_output_ptrs` guards with `IsValid()`.
- Python `EpDispatchCombineOp` reads the effective fp8 combine `scale_dim` from `handle_info` into a private attribute; `self.config.scale_dim` is no longer mutated by op construction.

## Test plan

Verified on MI300X, EP=8, `hidden_dim=7168`, `max-tokens=4096`, `zero-copy=0`:

- [x] `pytest tests/python/ops/test_dispatch_combine_intranode.py::test_dispatch_combine -k "fp8_blockwise and data_type0 and True"` — **16/16 pass**
- [x] Sanity: `MORI_FP8_COMBINE_SCALE_DIM=112` correctly drives the internal combine scale_dim to 112 while `config.scaleDim` (user value) stays untouched.
- [x] `bench_dispatch_combine.py --cmd bench --quant-type fp8_blockwise`: combine ~910 us (~19.5% faster than bf16 no-quant baseline ~1131 us), matching PR #311.
- [x] Same combine latency irrespective of caller-supplied `--scale-dim` ∈ {0, 32, 56}.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,6 +139,10 @@ jobs:
           $CT exec -e PYTHONPATH=$GITHUB_WORKSPACE $CONTAINER bash -c "
             cd $GITHUB_WORKSPACE
             timeout 120 python3 tests/python/ops/bench_dispatch_combine.py
+            timeout 120 python3 tests/python/ops/bench_dispatch_combine.py \
+              --cmd bench --dtype bf16 --quant-type fp8_blockwise \
+              --zero-copy 0 --max-tokens 128 \
+              --force-scale-active 1 --report-scale-stats 1
           "
 
       - name: MORI-IO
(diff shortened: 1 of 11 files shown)


maning00:
cc @billishyahao

PR #318, opened 2026-05-13
feat(jit): enable hip kernel debug info

kawhil-amd:
## Motivation

  Enable debugging of JIT-compiled kernels by adding debug symbol generation support. When troubleshooting kernel issues, developers need to inspect assembly with source-level debug info.

## Technical Details

  Add MORI_DEBUG_INFO environment variable to control hipcc debug flags in JIT compilation:

  - config.py: add is_debuginfo_enabled() to detect MORI_DEBUG_INFO env var
  - core.py: add _debuginfo_flags() returning ["-g", "-ggdb"] when enabled, integrate into _hipcc_genco()

  When MORI_DEBUG_INFO=1, hipcc jit compiles with -g -ggdb flags alongside -O2.

## Test Plan

MORI_PRECOMPILE=1 MORI_DEBUG_INFO=1 python -c "import mori"

## Test Result

- [x] The mori jit kernel pre-compilation test has been verified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,7 @@ python -c "import mori; print('OK')"
 ```bash
 cd /path/to/mori
 export PYTHONPATH=/path/to/mori:$PYTHONPATH
+python -c "import mori; print(mori.__file__)"
 
 # Test correctness (8 GPUs)
 pytest tests/python/ops/test_dispatch_combine_intranode.py -q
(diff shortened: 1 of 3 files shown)


jhchouuu (commented):
LGTM
Can the mode switching be achieved during runtime?

kawhil-amd:
> LGTM Can the mode switching be achieved during runtime?

The mode switching function requires clearing the kernel cache first, will consider adding a kernel cache clearing feature later.

PR #319, opened 2026-05-13
refactor(ccl): migrate collective kernels from AOT to JIT compilation

jhchouuu:
## Summary
Migrate CCL (collective) kernel compilation from build-time AOT (`-x hip` / `--hip-link`) to runtime JIT (`hipcc --genco`), consistent with the EP kernel architecture.
### Motivation
- **Fix `undefined symbol: atexit`** on glibc >= 2.34 (TheRock toolchain). AOT HIP compilation generates device registration code that references `atexit`, which newer glibc no longer exports as a dynamic symbol.
- **Architecture consistency**: all mori kernels (EP + CCL) now use the same JIT path — `.hip` → `hipcc --genco` → `.hsaco` → `HipModule` → `launch_struct`.
- **Clean wheel**: `libmori_collective.so` is now pure host C++ (no `--hip-link`, no device code embedded).
- **Runtime arch detection**: no build-time `--offload-arch` baked into the .so.
### Changes
- Split all CCL `__global__` kernels into `__device__ _body` + `__global__` wrapper in headers (backward compatible)
- Add `src/collective/kernels/ccl_kernels.hip` as JIT compilation unit with `extern "C"` entry points
- Add `CclAll2allArgs` / `CclAllgatherArgs` / `CclAllreduceArgs` POD args structs
- Refactor C++ host classes: remove `<<<>>>`, add `prepare_*/finish_*` methods for Python-side launch
- Rewrite `python/mori/ccl/collective.py` to use `compile_genco` + `HipModule` + `launch_struct`
- Remove `hip::device` from collective CMakeLists (host-only build)
- Add `ccl_kernels` to `MORI_PRECOMPILE` path
- Package `.cuh` files in wheel JIT sources
### Behavioral equivalence
All kernel names, grid/block sizes, launch order, and sync patterns are identical to the original AOT path. No runtime behavior change.
## Test plan
- [x] `pip install .` builds without `-x hip` or `--hip-link` for collective
- [x] `nm -D libmori_collective.so | grep "U atexit"` — no undefined atexit
- [x] `import mori.ops` / `import mori.ccl` — both work
- [x] `MORI_PRECOMPILE=1 python -c "import mori"` — ccl_kernels compiled and cached
- [x] `test_allgather --world-size 8 --elems 1048576` — passed
- [x] `test_all2all --world-size 8 --elems 1048576` — passed
- [x] AllReduce test

### DIFF
--- a/include/mori/collective/allgather/oneshot_allgather_sdma_class.hpp
+++ b/include/mori/collective/allgather/oneshot_allgather_sdma_class.hpp
@@ -32,6 +32,7 @@
 
 // Include necessary headers
 #include "mori/application/application.hpp"
+#include "mori/collective/ccl_kernel_args.hpp"
 #include "mori/collective/collective_pub.hpp"
 #include "mori/collective/core/wall_time.hpp"
 #include "mori/shmem/shmem.hpp"
@@ -88,6 +89,9 @@ class AllgatherSdma {
 
   // Range lookup helper: returns {SymmMemObjPtr, offset_from_base} or {invalid, 0}.
   std::pair<application::SymmMemObjPtr, size_t> find_registered(void* ptr) const;
+
+  // Reusable args struct for JIT kernel launch from Python
+  CclAllgatherArgs<T> jit_args_;
 
   // Disable copy constructor and assignment operator
   AllgatherSdma(const AllgatherSdma&) = delete;
@@ -231,6 +235,13 @@ class AllgatherSdma {
    * @brief Resets flags (sets to 0)
    */
   void resetFlags();
+
+  int64_t prepare_sync(T* input, T* output, size_t total_count, hipStream_t stream);
+  double finish_sync(T* output, size_t total_count, hipStream_t stream);
+  int64_t prepare_async_start(T* input, T* output, size_t total_count, hipStream_t stream);
+  void after_async_start();
+  int64_t prepare_async_wait(hipStream_t stream);
+  double finish_async_wait(hipStream_t stream);
 };
 
 }  // namespace collective
--- a/include/mori/collective/allgather/oneshot_allgather_sdma_class.hpp
+++ b/include/mori/collective/allgather/oneshot_allgather_sdma_class.hpp
@@ -90,7 +90,7 @@ class AllgatherSdma {
   // Range lookup helper: returns {SymmMemObjPtr, offset_from_base} or {invalid, 0}.
   std::pair<application::SymmMemObjPtr, size_t> find_registered(void* ptr) const;
 
-  // Reusable args struct for JIT kernel launch from Python
+  // Not reentrant: previous launch must complete before next prepare_*.
   CclAllgatherArgs<T> jit_args_;
 
   // Disable copy constructor and assignment operator
--- a/python/mori/ccl/__init__.py
+++ b/python/mori/ccl/__init__.py
@@ -20,27 +20,36 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-from .collective import All2allSdma
-from .collective import AllgatherSdma
-from .collective import AllreduceSdma
+try:
+    from .collective import All2allSdma
+    from .collective import AllgatherSdma
+    from .collective import AllreduceSdma
 
-# NCCL/RCCL-style C++ AllGather-into-tensor dispatcher.  The class and its
-# DataType enum are implemented entirely in C++ (see
-# ``include/mori/collective/allgather/allgather_into_tensor.hpp`` and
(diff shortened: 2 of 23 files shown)


python/mori/ccl/__init__.py
```diff
@@ -20,27 +20,36 @@
...
-]
+    __all__ = [
+        "All2allSdma",
+        "AllgatherSdma",
+        "AllreduceSdma",
+        "AllGatherIntoTensor",
+        "DataType",
+        "size_of",
+    ]
+except (ImportError, AttributeError):
```
jhchouuu, line 48: The `except (ImportError, AttributeError)` will be triggered because `_mori_cpp.AllGatherIntoTensor` was removed from pybind in this PR. The `__getattr__` error message tells the user to rebuild with `-DBUILD_COLLECTIVE=ON` -- but that flag is not the problem; `AllGatherIntoTensor` is simply not ported to the JIT path yet and no longer exists in pybind. A user following the error message will rebuild and still hit the same `AttributeError`.

Suggest either:
1. Keep the pybind registration for `AllGatherIntoTensor` (it does not suffer from the `atexit` issue and could stay in a separate AOT unit), or
2. Change the `__getattr__` message to explicitly say `AllGatherIntoTensor` is temporarily unavailable pending JIT porting.

src/collective/kernels/ccl_kernels.hip
```diff
@@ -0,0 +1,182 @@
...
+// ============================================================================
+
+// --- SdmaReduceScatterKernel ---
+CCL_ALLREDUCE_ALL_TYPES(CCL_WRAP_ALLREDUCE_RS, SdmaReduceScatterKernel)
+
+// --- AllGatherSdmaKernel ---
+CCL_ALLREDUCE_ALL_TYPES(CCL_WRAP_ALLREDUCE_AG, AllGatherSdmaKernel)
+
+// --- ReduceScatterAllGatherFusedKernel ---
+CCL_ALLREDUCE_ALL_TYPES(CCL_WRAP_ALLREDUCE_RS, ReduceScatterAllGatherFusedKernel)
```
jhchouuu: `ReduceScatterAllGatherFusedKernel` is compiled here (5 type instantiations via `CCL_ALLREDUCE_ALL_TYPES`) but the Python allreduce path never launches it -- it uses `SdmaReduceScatterKernel` + `AllGatherSdmaKernel` as two separate steps. This is dead code in the JIT path and adds unnecessary compile time. Either wire it up from Python or remove it from this file.

src/collective/core/oneshot_allgather_sdma_class.cpp
```diff
@@ -444,34 +363,172 @@ bool AllgatherSdma<T>::operator()(T* input, T* output, size_t total_count, hipSt
...
+
+  jit_args_.myPe = myPe_;
+  jit_args_.npes = npes_;
+  jit_args_.input = input;
+  jit_args_.srcMemObj = input_transit_buffer_obj_;
+  jit_args_.dstMemObj = dst_obj;
+  jit_args_.flagsMemObj = flagsObj_;
+  jit_args_.elementCount = total_count;
+  jit_args_.dstBaseOffset = direct ? byteOffset : 0;
+  jit_args_.flagVal = 0;
```
jhchouuu: `jit_args_.flagVal = 0` is not a functional bug -- `CCL_WRAP_ALLGATHER_ASYNC_PUT` does not forward `flagVal` to the put-kernel body. However, `jit_args_` is a shared member struct: `prepare_async_wait` will later overwrite this same field with `async_flag_token_`. Setting it to 0 here makes it look like the put phase intentionally signals generation 0, which is confusing.

Suggest setting `flagVal = async_flag_token_` here (already computed two lines above) so both phases consistently carry the correct generation number.

include/mori/collective/allgather/oneshot_allgather_sdma_class.hpp
```diff
@@ -89,6 +90,9 @@ class AllgatherSdma {
   // Range lookup helper: returns {SymmMemObjPtr, offset_from_base} or {invalid, 0}.
   std::pair<application::SymmMemObjPtr, size_t> find_registered(void* ptr) const;
 
+  // Reusable args struct for JIT kernel launch from Python
+  CclAllgatherArgs<T> jit_args_;
```
jhchouuu, line 94: `jit_args_` is a single instance shared across all `prepare_*` / `finish_*` calls on this handle. All three prepare methods write into it and return a raw pointer to the same struct. If a caller pipelines two operations on the same handle without waiting for the GPU to consume the previous launch, the struct can be overwritten mid-flight.

The same pattern exists in `All2allSdma` and `AllreduceSdma`. Current sequential usage is safe, but this should be documented explicitly (or enforced with an assertion) to guard against future misuse.

jhchouuu (commented):
Review comments on the AOT to JIT migration.

PR #321, opened 2026-05-14
fix(sdma): All2allSdma completion flags not reset between calls

ylerman-dn:
## Motivation

`resetFlags()` was commented out in `All2allSdma::start_async()`, `All2allSdma::operator()`, and `All2allSdma::cancel_async()` since the initial SDMA commit. Without resetting, completion flags remain set to 1 after the first call, causing all subsequent calls to return immediately with stale output.

The original test used fixed input values across iterations, so stale output from call 1 matched the expected values for call 2+, masking the bug entirely.

## Technical Details

Uncomment `resetFlags()` in the three call sites in `src/collective/core/oneshot_all2all_sdma_class.cpp`.

The test is also updated to use iter-dependent input values so the bug is reliably reproducible on call 2+, and async mode is covered via a new `--async` CLI flag.

## Test Plan

- Sync mode, no fix → FAIL (stale flags produce wrong output on iteration 2+)
- Sync mode, fix applied → PASS
- Async mode, fix applied → PASS

## Test Result

Tested on AMD MI355X (8 GPUs), `rocm/primus:v26.2`, 1M elements, 10 iterations — all 8 PEs passed in both sync and async modes.

### DIFF
--- a/src/collective/core/oneshot_all2all_sdma_class.cpp
+++ b/src/collective/core/oneshot_all2all_sdma_class.cpp
@@ -158,7 +158,7 @@ bool All2allSdma<T>::start_async(T* input, T* output, size_t total_count, hipStr
     // copy_input_to_transit(input, total_count * npes_, stream);
 
     // Step 2: Reset flags
-    // resetFlags();
+    resetFlags();
 
     // Step 3: Execute All2All kernel (PUT operation)
     printf("PE %d: Launching async PUT kernel...\n", myPe_);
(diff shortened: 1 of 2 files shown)


ylerman-dn:
Investigation concluded that the suspected flag reset bug is not real — the kernel already resets completion flags internally (flags[threadLinearId] = 0). The original test failure was caused by a verification bug in the test, not stale SDMA flags. Closing this PR.

PR #335, opened 2026-05-25
feat(ep): add cached routing replay for training dispatch/combine

sudhu2k:
## Motivation

MoE training needs dispatch/combine behavior that inference-only paths do not provide:

1. **Cached routing (mode-1 / mode-2)** — Forward dispatch records where each top-k slot lands; backward must **replay the same layout** without redoing atomics, dedup, and cross-rank slot assignment. That matches DeepEP’s “first call” vs “cached replay” split used in training stacks.

2. **`-1` sentinel experts** — Routing tensors can carry DeepEP-style sentinels for dropped or invalid top-k slots. Dispatch must skip those slots; combine must not count them toward PE fan-in.

3. **Multi-layer reuse of one op** — A single `EpDispatchCombineOp` may serve many layers if each layer keeps its own routing snapshot and combine uses the correct handle, without stale symmetric-buffer state bleeding across layers.

This PR adds that training-oriented surface on **IntraNode** and **InterNodeV1** only. Low-latency (LL) internode kernels are intentionally unchanged so inference/LL paths stay aligned with `main`.

### DeepEP-style `-1` sentinels

On the IntraNode path (`intranode.hpp`), any top-k slot with a negative expert ID is treated as a dropped route: the kernel does not send payload for that slot, writes the combine null sentinel into `dispDestTokIdMap` (encoded as `PE == worldSize`), and excludes negative peers when deduplicating sends to the same destination rank.

InterNodeV1 (`internode_v1.cpp`) applies the same sentinel rules on the internode send path. In replay mode, it reuses the maps recorded by an earlier mode-1 dispatch instead of rebuilding layout.

Coverage lives in `test_dispatch_combine_minus_one_sentinel` (IntraNode only). Test data generation accepts a configurable `sentinel_pattern` in `dispatch_combine_test_utils.py`, and the combine reference only counts experts with `idx >= 0` when computing unique PE fan-in.

### DeepEP-style routing handle

This PR introduces `EpDispatchRoutingHandle` in Python and `EpDispatchCombineRoutingPtrs` in C++. Both wrap caller-owned tensors—`disp_dest_tok_id_map`, internode routing maps, `total_recv_token_num`, and `disp_tok_id_to_src_tok_id_local`—that capture where each top-k slot was placed for a given forward pass.

In **mode-1**, the caller passes `return_routing=True` on `dispatch()`. Kernels run with `replayMode=false`, perform slot assignment and map population, and then `snapshot_disp_tok_id_to_src_tok_id_local` copies the symmetric inverse map into the handle so later no-P2P combine can read a stable local view.

In **mode-2**, the caller passes `routing=R` on a subsequent `dispatch()`. Kernels run with `replayMode=true`, reuse the frozen maps from mode-1, and may ship a new activation payload while expert indices and layout stay unchanged. The same handle is passed to `combine(..., routing=R)`, which reads those snapshot maps rather than relying solely on op-owned symmetric state.

The Python API exposes `routing` and `return_routing` as keyword-only arguments on `dispatch()` and `combine()`, and rejects passing both at once. This path is implemented for `IntraNode` and `InterNodeV1` only. Pybind adds `build_args_with_routing` and `snapshot_disp_tok_id_to_src_tok_id_local`, and threads routing pointers through `GetEpDispatchCombineArgsRaw`.

## Submission Checklist


python/mori/ops/dispatch_combine.py
isytwu, line 746: Should this pass routing.total_recv_token_num.data_ptr() when use_routing_handle is true? Otherwise call_local_expert_count still reads handle.totalRecvTokenNum.

src/ops/dispatch_combine/intranode.hpp
```diff
@@ -108,38 +108,64 @@ __device__ void EpDispatchIntraNodeKernel_body(EpDispatchCombineArgs<T> args) {
...
-                               config.numExpertPerRank);
-      }
-      if (__any(condition)) {
-        // Indicate that this token is already sent to the destination PE by setting an overflow
-        // token index
-        if (laneId == 0) args.dispDestTokIdMap[i] = FlatTokenIndex(config, config.worldSize, 0);
-        continue;
-      }
+      if (!args.replayMode) {
+        // Mode-1 (DeepEP-style cached-mode "first call"): decide where this
```
isytwu: Could we describe this as cached/replay routing instead of Mode-1 / Mode-2 / DeepEP-style?

isytwu (commented):
Thanks for working on this PR. I left a couple of minor comments. Overall this looks good from my side once those are addressed.

src/ops/dispatch_combine/intranode.hpp
sudhu2k: Changed it!

python/mori/ops/dispatch_combine.py
sudhu2k: You're right, I've fixed it. Thanks!

src/ops/dispatch_combine/intranode.hpp
```diff
@@ -108,38 +108,64 @@ __device__ void EpDispatchIntraNodeKernel_body(EpDispatchCombineArgs<T> args) {
...
-        condition = destPe == (args.tokenIndices[srcTokId * config.numExpertPerToken + laneId] /
-                               config.numExpertPerRank);
-      }
-      if (__any(condition)) {
-        // Indicate that this token is already sent to the destination PE by setting an overflow
-        // token index
-        if (laneId == 0) args.dispDestTokIdMap[i] = FlatTokenIndex(config, config.worldSize, 0);
-        continue;
-      }
+      if (!args.replayMode) {
```
isytwu, line 115: Thanks for addressing the earlier comments. One more performance-related point: to avoid any possible regression on the default non-replay path, would it make sense to specialize this as a static/template variant instead?

sudhu2k: replayMode is a plain scalar that's uniform across the entire launch hence there wouldn't be any warp divergence.
I ran a benchmark using the following command:
`python3 -m tests.python.ops.bench_dispatch_combine   --cmd bench --world-size 8 --max-tokens 4096 --hidden-dim 7168 --dtype bf16`

the two runs were essentially identical — the replayMode change shows no measurable regression

Phase | main | curr branch | Δ (run 2) | Δ (run 1)
-- | -- | -- | -- | --
Dispatch | 1035.8 µs | 1032.1 µs | −3.7 µs (−0.36%) | +1.5 µs (+0.15%)
Combine | 935.0 µs | 932.8 µs | −2.2 µs (−0.23%) | −3.6 µs (−0.39%)
E2E | ~1946.1 µs | ~1941.6 µs | −4.5 µs (−0.23%) | ~flat

replayMode is a launch-uniform, perfectly-predicted branch, so it costs effectively nothing on the hot path. The data confirms there's no need to template-specialize it for performance. Let me know what you think!

isytwu (approved):

src/ops/dispatch_combine/intranode.hpp
isytwu: Could you also check a small-token case, e.g. --max-tokens 1 or a few low-token settings?

sudhu2k: Seems like noise to me. 

### 1 token
| kernel | curr (µs) | main (µs) | Δ | Δ% |
|--------|-----------|-----------|------|------|
| Dispatch | 36.0 | 37.6 | -1.6 | -4.3 |
| Combine  | 20.9 | 21.2 | -0.3 | -1.4 |
| E2E      | 32.2 | 30.6 | +1.6 | +5.2 |
### 8 tokens
| kernel | curr (µs) | main (µs) | Δ | Δ% |
|--------|-----------|-----------|------|------|
| Dispatch | 39.1 | 38.4 | +0.7 | +1.8 |
| Combine  | 27.9 | 27.5 | +0.4 | +1.5 |
| E2E      | 41.7 | 41.4 | +0.3 | +0.7 |
### 64 tokens
| kernel | curr (µs) | main (µs) | Δ | Δ% |
|--------|-----------|-----------|------|------|
| Dispatch | 44.3 | 43.0 | +1.3 | +3.0 |
| Combine  | 35.1 | 35.1 | 0.0 | 0.0 |
| E2E      | 55.3 | 54.9 | +0.4 | +0.7 |
### 512 tokens
| kernel | curr (µs) | main (µs) | Δ | Δ% |
|--------|-----------|-----------|------|------|
| Dispatch | 161.9 | 160.5 | +1.4 | +0.9 |
| Combine  | 137.2 | 135.7 | +1.5 | +1.1 |
| E2E      | 274.6 | 272.4 | +2.2 | +0.8 |

PR #363, opened 2026-06-05
fix(ep): guard dispatch kernels against out-of-range expert id

inkcherry:
## Submission Checklist


python/mori/ops/dispatch_combine.py
```diff
@@ -66,6 +65,38 @@ def _current_stream():
     return torch.cuda.current_stream().cuda_stream
 
 
+def validate_dispatch_indices(min_index, max_index, num_experts_per_rank, world_size):
```
isytwu: Can we skip the host-side validator? The real protection should be the device-side destPe guard. This adds an opt-in env var path plus GPU->CPU sync / graph-capture handling, but it is default-off and does not protect normal dispatch.

tests/python/ops/test_dispatch_index_guard_all_kernels.py
```diff
@@ -0,0 +1,154 @@
+# Copyright © Advanced Micro Devices, Inc. All rights reserved.
```
isytwu: I’d prefer not to add this test. It mostly tests Python models of the kernel predicates instead of the real kernels, which feels inconsistent with the existing dispatch tests. A small unit test for the simple range check should be enough if the validator stays.

tests/python/ops/test_internode_dispatch_destpe_guard.py
```diff
@@ -0,0 +1,146 @@
+# Copyright © Advanced Micro Devices, Inc. All rights reserved.
```
isytwu: Same here: this duplicates kernel logic in Python rather than exercising the actual dispatch path. I think we can drop it and keep the PR focused on the device-side guard.

python/mori/ops/dispatch_combine.py
inkcherry: thanks, modified

tests/python/ops/test_dispatch_index_guard_all_kernels.py
inkcherry: modified

tests/python/ops/test_internode_dispatch_destpe_guard.py
inkcherry: modified

isytwu (approved):

PR #366, opened 2026-06-05
fix(ep): add mlx5 collapsed CQ + dedicated dispatch staging buffer for internode-v1

jhchouuu:
## Summary

Two related changes to the mlx5 internode-v1 dispatch/combine path: a collapsed-CQ completion path for the IBGDA quiet, and a fix for a silent data-corruption bug that only showed up on mlx5 under high pressure.

## 1. Collapsed CQ (cc=1 / oi=1) completion tracking

The NIC collapses every completion into a single CQE slot (overrun ignored, so there is no per-CQE consumer ring and no CQ consumer doorbell to maintain). The device advances `wq.doneIdx` purely from `CQE[0].wqe_counter`, reconstructing the 32-bit completion from the 16-bit counter against `doneIdx` (a guaranteed lower bound, since outstanding < sqWqeNum < 65536). `atomicMax` on `doneIdx` keeps it lock-free and multi-warp safe.

- Final per-pe quiet waits for the live `postIdx` (every reserved WQE completed); the SQ recycle gate keeps a cheap snapshot drain (a live drain there would self-deadlock, since the gated warp already holds its own un-doorbelled reservation).
- CQ buffer is initialized to `0xff` so `wqe_counter` reads `0xffff` ("nothing completed") until the NIC writes a real completion.
- The GPU-resident control structures (CQ, both doorbell records, QP, atomic ibuf) are allocated uncached for NIC↔GPU coherence.
- Device-scope fences in the drain (it only touches the uncached CQ and the GPU-only `doneIdx`); dead `outstandingWqe` writes removed on the mlx5/psd providers.

## 2. Fix silent data corruption

The inter-node dispatch send is non-blocking and reads its payload from `staging`. The same iteration's combine phase then writes its accumulation results back into `staging` (an overlapping region). Under load the NIC could still be DMA-reading `staging` for the dispatch send when combine overwrote it, so combine's activation data was sent into a peer's `dispatchInp` — silent corruption (zero CQE errors, no loss/reorder), reproducible only on mlx5-RoCE under high pressure (qp1 and qp4), independent of congestion (persisted even at zero ECN).

Fix: dispatch gets its own send-source buffer `dispatchStaging`, filled by `EpDispatchCopyToStaging` and read by the send, which combine never writes — removing the reuse hazard entirely.

## Validation

- 2-node mlx5-RoCE, EP16 bench/stress mode, SL=0 / TC=41: qp1 / qp2 / qp4 all PASS at 50000 iterations

### DIFF
--- a/include/mori/ops/dispatch_combine/dispatch_combine.hpp
+++ b/include/mori/ops/dispatch_combine/dispatch_combine.hpp
@@ -182,6 +182,8 @@ struct ShmemBufsInterNodeV1 {
   mori::application::SymmMemObjPtr dispatchOut;
   mori::application::SymmMemObjPtr combineOut;
   mori::application::SymmMemObjPtr staging;
+  // Dispatch send source, separate from `staging` so combine can't overwrite it.
+  mori::application::SymmMemObjPtr dispatchStaging;
 };
 
 // InterNode / AsyncLL: full 5-buffer set used by the non-V1 RDMA paths.
--- a/include/mori/shmem/shmem_ibgda_kernels.hpp
+++ b/include/mori/shmem/shmem_ibgda_kernels.hpp
@@ -285,98 +285,73 @@ inline __device__ void ShmemQuietThreadKernelPsdImpl(int pe, int qpId) {
   }
 }
 
+// Drain a collapsed CQ (cc=1, oi=1) and advance wq.doneIdx. The NIC keeps the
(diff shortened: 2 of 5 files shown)


jhchouuu:
EP16 benchmark passed:
image

jhchouuu:
EP16 1/2/4 QP stress test passed at 50000 iterations:
QP=4:
image
QP=2:
image
qp=1:
image

PR #370, opened 2026-06-08
feat(io): add rail affinity for RDMA endpoint connection

staryxchen:
## Summary

Add rail affinity support to mori-io's RDMA backend. When enabled, the sender includes its NIC index (`railId`) in the `RegEndpoint` handshake message, and the receiver uses the same index to select its local NIC — ensuring both ends connect through the same leaf switch rail rather than crossing spine links.

This optimization targets rail-optimized fat-tree networks (common in homogeneous GPU clusters) where NIC-i on every node attaches to the same leaf switch. On rail-isolated networks (no spine switch), cross-rail traffic is completely unreachable, making this fix critical for correctness.

## Motivation: Why `nicRank` is insufficient

The existing `nicRank` field represents a **relative index** into each side's topology-sorted NIC candidate list. When sender and receiver memory reside on different NUMA nodes, `MatchCpuNics()` produces different NIC orderings based on NUMA affinity:

```
Sender (memory on NUMA-0):   Search(numaNode=0) → [NIC-0, NIC-1, NIC-2, NIC-3, ...]
Receiver (memory on NUMA-1): Search(numaNode=1) → [NIC-4, NIC-5, NIC-6, NIC-7, ...]

nicRank=0 → Sender picks NIC-0, Receiver picks NIC-4 → Cross-rail!
```

**Verified on MI335X with 8×ionic NICs**: cross-NUMA CPU transfer without rail affinity results in `Connection timed out` on rail-isolated networks.

`railId` solves this by transmitting the **absolute** `availDevices` index, bypassing the receiver's local topology sort entirely.

## Changes

- **`src/io/rdma/protocol.hpp`** — Add `railId` field to `MessageRegEndpoint` (default `-1` for backward compatibility)
- **`src/io/rdma/backend_impl.hpp`** — Add `RdmaManager::NumAvailDevices()` accessor
- **`src/io/rdma/backend_impl.cpp`** — Sender passes `devId` as `railId`; receiver matches `railId` when `MORI_IO_RAIL_AFFINITY=1` is set, otherwise falls back to the existing topology-based selection; added trace logging for NIC selection
- **`tests/cpp/io/test_protocol.cpp`** — Unit tests covering railId serialization, msgpack round-trip, backward compat defaults, and decision logic
- **`tests/cpp/io/test_rail_affinity.cpp`** — Integration test: verifies cross-NUMA rail affinity on a single machine (requires libnuma + ≥2 NUMA nodes)
- **`tests/cpp/CMakeLists.txt`** — Register `test_protocol` and `test_rail_affinity` build targets

## Design

- **Rail ID** = index in `availDevices` (from `ibv_get_device_list`). Same-SKU nodes enumerate NICs in the same order, so the index is a zero-config rail identifier — same approach as NCCL's rail alignment.
- **Backward compatible**: msgpack array append; old receivers ignore the extra field, old senders produce `railId=-1` which triggers the fallback path.
- **Opt-in**: disabled by default. Set `MORI_IO_RAIL_AFFINITY=1` to enable. Heterogeneous clusters (different NIC counts/ordering) should leave it off.
- **Trade-off**: receiver may use a non-NUMA-local NIC (extra PCIe hop, ~ns penalty) to guarantee rail alignment (avoids ~μs cross-spine penalty or total unreachability on rail-isolated fabrics).

## Test plan

- [x] `test_protocol` unit tests pass (TCP round-trip, msgpack, decision logic)
- [x] `test_rail_affinity` integration test: cross-NUMA transfer succeeds with `MORI_IO_RAIL_AFFINITY=1`, times out without it on rail-isolated network (verified on MI335X, 8×ionic)
- [ ] Multi-node integration: 2-node cluster with `MORI_IO_RAIL_AFFINITY=1`, grep logs for `"Rail affinity: using railId="` confirming both ends select matching devId

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -628,6 +628,36 @@ jobs:
             --num-initiator-dev 8 --num-target-dev 8
           wait
 
+      - name: MORI-IO internode cross-rail write (rail affinity)
+        run: |
+          SCRIPT="$GITHUB_WORKSPACE/tools/run_internode_io_benchmark.sh"
+          PORT=29125
+          # Cross-rail validation: --target-dev-offset 5 makes GPU-0 pair with GPU-5
+          # (different NUMA → different NIC preference). With MORI_IO_RAIL_AFFINITY=1
+          # the receiver must follow the sender's railId, keeping traffic on-rail.
+          # On a rail-isolated fabric this would fail without rail affinity.
+          echo "=== MORI-IO internode: cross-rail GPU write (offset=5, rail affinity ON) ==="
+          NODE2_CMD=(
+            $CT exec -e MORI_IO_RAIL_AFFINITY=1 $CONTAINER bash $SCRIPT
+            --rank 1 --master-addr $NODE1_HOST --master-port $PORT
+            --ifname $NODE2_IFNAME
+            -- --op-type write --buffer-size 4096 --transfer-batch-size 64
+            --iters 4 --enable-sess --enable-batch-transfer
+            --num-qp-per-transfer 2
+            --num-initiator-dev 8 --num-target-dev 8
+            --target-dev-offset 5
+          )
+          ssh $SSH_OPTS -p $NODE2_PORT $(whoami)@$NODE2_HOST "${NODE2_CMD[*]}" &
+          $CT exec -e MORI_IO_RAIL_AFFINITY=1 $CONTAINER bash $SCRIPT \
+            --rank 0 --master-addr $NODE1_HOST --master-port $PORT \
+            --ifname $NODE1_IFNAME \
+            -- --op-type write --buffer-size 4096 --transfer-batch-size 64 \
+            --iters 4 --enable-sess --enable-batch-transfer \
+            --num-qp-per-transfer 2 \
+            --num-initiator-dev 8 --num-target-dev 8 \
+            --target-dev-offset 5
+          wait
+
       - name: MORI-EP internode normal kernel bench
         run: |
           SCRIPT="$GITHUB_WORKSPACE/tools/run_internode_test.sh"
--- a/src/io/rdma/backend_impl.cpp
+++ b/src/io/rdma/backend_impl.cpp
@@ -905,7 +905,7 @@ void ControlPlaneServer::BuildRdmaConn(EngineKey ekey, TopoKeyPair topo, int nic
   application::RdmaEndpoint lep = rdma->CreateEndpoint(devId);
 
   Protocol p(tcph);
-  p.WriteMessageRegEndpoint({myEngKey, topo, devId, lep.handle, rank});
+  p.WriteMessageRegEndpoint({myEngKey, topo, devId, lep.handle, rank, devId});
   MessageHeader hdr = p.ReadMessageHeader();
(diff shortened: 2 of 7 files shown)


maning00:
Hi @staryxchen, our **`mori internode test (MI355X_AINIC)`** CI runs on a rail-only fabric, so it's a great place to actually validate this. Would you mind adding a cross-rail test case there? FYI #379 adds a `--target-dev-offset` benchmark flag that makes it easy to set up a cross-rail (e.g. GPU0→GPU5) transfer.

staryxchen:
@maning00 Thanks for the suggestion! I've added a cross-rail test step to the MI355X_AINIC internode CI in the latest push:

**"MORI-IO internode cross-rail write (rail affinity)"**
- Uses `--target-dev-offset 5` (from #379) to pair GPU-0 with GPU-5, forcing each side to prefer a different NUMA-local NIC
- Sets `MORI_IO_RAIL_AFFINITY=1` so the receiver follows the sender's `railId` instead of picking its own NUMA-local NIC
- CI just passed on the rail-only fabric ✅

Note: the current test only validates the positive case (affinity ON → success). I didn't add a negative case (affinity OFF + cross-rail) since it would timeout/fail and slow down CI. That said, we confirmed the failure locally on our MI335X rail-isolated setup (`Connection timed out` when sender NIC-0 pairs with receiver NIC-4).

Do you think it's worth adding a negative test on MI355X_AINIC as well — e.g. a short run with `MORI_IO_RAIL_AFFINITY=0` and `--target-dev-offset 5`, expecting failure? That would provide stronger end-to-end proof on the actual rail-only fabric, but at the cost of CI time (~5s timeout per attempt).

maning00:
> @maning00 Thanks for the suggestion! I've added a cross-rail test step to the MI355X_AINIC internode CI in the latest push:
> 
> **"MORI-IO internode cross-rail write (rail affinity)"**
> 
> * Uses `--target-dev-offset 5` (from [test(io): add --target-dev-offset to MORI-IO benchmark #379](https://github.com/ROCm/mori/pull/379)) to pair GPU-0 with GPU-5, forcing each side to prefer a different NUMA-local NIC
> * Sets `MORI_IO_RAIL_AFFINITY=1` so the receiver follows the sender's `railId` instead of picking its own NUMA-local NIC
> * CI just passed on the rail-only fabric ✅
> 
> Note: the current test only validates the positive case (affinity ON → success). I didn't add a negative case (affinity OFF + cross-rail) since it would timeout/fail and slow down CI. That said, we confirmed the failure locally on our MI335X rail-isolated setup (`Connection timed out` when sender NIC-0 pairs with receiver NIC-4).
> 
> Do you think it's worth adding a negative test on MI355X_AINIC as well — e.g. a short run with `MORI_IO_RAIL_AFFINITY=0` and `--target-dev-offset 5`, expecting failure? That would provide stronger end-to-end proof on the actual rail-only fabric, but at the cost of CI time (~5s timeout per attempt).

Thanks @staryxchen, the positive case looks good to me. I'd skip the negative test — on a genuinely rail-only fabric the positive case is already strong proof, and not all our runners are rail-only (MI355X_AINIC runner for now), a negative case may add some flakiness to every CI run for limited extra value.

maning00 (approved):

PR #377, opened 2026-06-10
feat(cco): BNXT GDA support + split scale-out header + move LSA tests

jhchouuu:
## Summary
Three related CCO changes (3 commits):

1. **BNXT GDA support** — make the GDA (cross-node IBGDA RDMA) device layer work on
   Broadcom NICs, plus runtime provider selection in the GDA tests.
2. **Header split** — factor the GDA device layer out of `cco.hpp` into
   `cco_scale_out.hpp` so host-only / LSA-only translation units no longer pull in
   the provider RDMA core.
3. **LSA tests** — move the 3 LSA examples into the cco test suite as ctests.

### 1. BNXT GDA support (`dc5126c9`)
GDA hung at 3+ ranks on BNXT. BNXT dedups the UAR doorbell page across endpoints, so
distinct per-peer QPs in one warp share a `dbrAddr`; a per-thread put then has each
active lane ring a different QP's doorbell in the same SIMT step, and same-address
store coalescing drops all but one doorbell — the dropped WQE never completes and
`quietUntil` spins forever (2 ranks pass = single active lane; 3+ hang).

- `ringDoorbellOrdered`: lane-serialize the BNXT doorbell MMIO (mirrors the PSD path).
- `quietUntil`: real BNXT completion tracking for the `cqeNum==1` CQ model — map the
  CQE's SQ slot back to the monotonic logical WQE id via `outstandingWqe[]` under
  `pollCqLock` and advance `doneIdx` with `fetch_max` (replaces a broken MLX5-style copy).
- add `recordOutstandingWqe` helper (BNXT `% sqWqeNum`, PSD `% table`).
- `ccoIbgdaContext` exposes a host-readable `providerType`, resolved at DevComm
  creation from the first connected endpoint's vendor id.
- GDA tests drop the hardcoded provider; `CCO_GDA_DISPATCH` selects the
  `ccoGda<PrvdType>` instantiation at runtime from `devComm.ibgda.providerType`.

### 2. Split GDA device layer into `cco_scale_out.hpp` (`f5c37032`)
- `cco.hpp`: shared types (incl. the GDA POD structs embedded in `ccoDevComm`),
  cooperative groups, teams, the LSA barrier session, and the host API. No longer
  includes any RDMA headers.
- `cco_scale_out.hpp` (new): `ccoGda<PrvdType>` + the `mori::cco::impl` provider
  primitive layer; the sole consumer of `rdma.hpp`. Includes `cco.hpp` and is
  self-contained — GDA users include just this header. One-directional dependency,
  no cycle.

### 3. Move LSA examples into the test suite (`9583d3ca`)
- `examples/cco/lsa_{barrier,allreduce,memcheck}.cpp` →
  `tests/cpp/cco/test_cco_lsa_*.cpp`; the auto-discovery harness builds them
  (HIP + MPI) and registers `cco_lsa_<x>` (np=1) + `cco_lsa_<x>_mpi` (np=8).
- replace `assert()` (which drops side-effecting setup calls under the Release
  `-DNDEBUG` build) with an always-evaluated `CCO_MUST()` macro; drop a dead
  `args_parser.hpp` include; remove the old example registrations.

## Testing
All run in-container (rocm/mori:ci, MI300X + BNXT):
- GDA: `put`/`get`/`signal`/`flush_async` pass at 2/3/4/6/8 ranks.
- `test_cco_host`, `test_cco_gda_modes` pass.
- `ctest -R cco_lsa` → 6/6 passed (np=1 and np=8 each).


include/mori/cco/cco_scale_out.hpp
```diff
@@ -0,0 +1,1136 @@
...
+    // across endpoints (see BnxtCqContainer / TryRegisterUar), so distinct QPs
+    // in one warp can share the same dbrAddr. A per-thread GDA put has each
+    // active lane targeting a *different* peer's QP in the same SIMT step; if
+    // those QPs share a UAR, same-address store coalescing drops all but one
+    // lane's doorbell — the dropped WQE is never fetched and quiet hangs. Walk
+    // the warp's active lanes one at a time so every doorbell store survives.
+    core::UpdateSendDbrRecord<PrvdType>(wq->dbrRecAddr, myPostIdx + numWqes);
+    __threadfence_system();
+    uint64_t mask = core::GetActiveLaneMask();
+    while (mask) {
```
kawhil-amd, line 379: Maybe we could replace this code with ringDoorbellWarp func?  Expand this function to more providers by cpp template specialization.

jhchouuu: sure, in our TODOs

PR #384, opened 2026-06-11
Fix: umbp qp per transfer default value & add logs

TianDi101:
- Change default qpPerTransfer = 4 to avoid AINIC resource overflow
- Add logs for rdma transport implementation


TianDi101:
Leave the review solved by the next PR, we need to merge this one to main for an urgent test

PR #412, opened 2026-06-22
perf(ep): improve intranode dispatch kernel load balancing and eliminate repeated loads

kawhil-amd:
## Problem

   * The existing intranode dispatch kernel reads each token's hidden-dim payload from HBM once per destination PE, causing up to 8× redundant HBM reads at max fan-out. 
   * Routing and data copy run serially within the same warp.

  ## Change
  *  Load-once-write-many: each warp loads its hidden-dim slice from HBM once into registers, then writes the same data to all destination PEs via v_readlane_b32 pointer broadcast.
  *  Pipelined control/data plane: warp 0 computes routing for the next token while data-plane warps copy the current token, using a double-buffered TokenRoute in shared memory.

### Dispatch Latency  (MI355X)
* topk = 8, ep=8,  dtype=bf16, hidden_dim=7168
* **per_block_warp_num=9**  (decouple the control plane and data plane onto different warps)

| block_num | topk | max_token | Baseline(us) | Current(us) | Speedup |
|-----------|------|-----------|----------|------------|---------|
| 1        | 8    | 1        |    15.8      |    11.6        |   +26.5%      |
| 32        | 8    | 32       |    19.16      |   15.03          |   +20.8%      |
| 64        | 8    | 64       |    24.04      |     23       |   +4.1%      |


src/ops/dispatch_combine/intranode_ll.hpp
```diff
@@ -0,0 +1,458 @@
...
+        FlatTokenIndex(args.config, laneId, warpSlot);
+  }
+}
+
+/* ──────────────────────────────────────────────────────────────────────────────
+ * EpDispatchIntraNodeCoopKernel
+ *
+ * Token-centric cooperative dispatch. Compared to the per-(token,expert)
+ * EpDispatchIntraNodeKernel and the 2-warp-group EpDispatchIntraNodeLLKernel:
+ *
```
jhchouuu, line 232: 注释还是coop kernel，顺手改一下？

src/ops/dispatch_combine/intranode_ll.hpp
```diff
@@ -0,0 +1,458 @@
...
+        }
+      }
+
+      // ── Data plane: warps 2..N-1 broadcast-copy token payload ────────
+      if (warpId > 1) {
+        const TokenRoute& route = routeBuf[pingPong];
+
+        if (route.numDests > 0) {
+          const int copyWarpRank = warpId - 2;
+          const int numCopyWarps = warpNum - 2;
```
jhchouuu, line 373: warpNum 好像要大于等于3，是不是可以保护一下？host端launch的时候？

jhchouuu (approved):
LGTM

src/ops/dispatch_combine/intranode_ll.hpp
kawhil-amd: The next pr will fix it.

src/ops/dispatch_combine/intranode_ll.hpp
kawhil-amd: The next pr will do it.

PR #414, opened 2026-06-22
(bugfix) mori bug on RoCM714

QizhouZhang97:
# Fix MORI build + multi-process topology crash on ROCm 7.14

## Summary
Two fixes that let MORI build and run on ROCm 7.14 (no regression on 7.2).

## 1. CMake: missing `numa::numa` target
ROCm 7.14's `hsakmt-config.cmake` declares that `hsakmt::hsakmt` links `numa::numa`, but ships no `find_package`/finder to define it, so configuration aborts:

> The link interface of target "hsakmt::hsakmt" contains: numa::numa but the target was not found.

**Fix:** define a `numa::numa` IMPORTED target from the system `libnuma` before `find_package(hsakmt)`.

## 2. Runtime: `rsmi_init()` crash under multi-process (rsmi symbol split-brain)
On the **ROCm 7.14** docker, the `rsmi_*` symbols resolve to **two different shared objects in the same process**: some calls (e.g. `rsmi_init`, `rsmi_is_P2P_accessible`) bind to `libamd_smi.so` (amdsmi), while the rest (`rsmi_num_monitor_devices`, `rsmi_dev_pci_id_get`, …) bind to `librocm_smi64`. amdsmi sits in the global symbol scope and interposes the subset of the rsmi ABI it re-exports.

The two libraries keep **separate global singletons**, so `rsmi_init()` initializes one while `rsmi_num_monitor_devices()` queries the other (uninitialized) → `RSMI_STATUS_INIT_ERROR`, and the check's `exit(-1)` kills every rank. This does **not** happen on ROCm 7.2, where all symbols resolve to a single library.

**Fix:** pin the library ourselves. `TopoSystemGpu::Load()` now `dlopen`s `librocm_smi64` with `RTLD_NOW | RTLD_LOCAL` and resolves **every** `rsmi_*` symbol from that one handle, so all calls hit the same instance and amdsmi can no longer interpose. The link-time dependency on `librocm_smi64` is removed from CMake, which also means the library is loaded **lazily, after torch has initialized HIP** — its `NEEDED libamdhip64` then resolves to the already-loaded copy by SONAME instead of pulling in a second HIP runtime.

An optional `MORI_ROCM_SMI_PATH` env var overrides the library path.

## Test plan
- `pip install . --no-build-isolation` builds on ROCm 7.14.
- 8-rank `test_dispatch_combine_internode.py --cmd stress` completes 5000/5000 dispatch + combine with no `RSMI_STATUS_INIT_ERROR` (no `LD_PRELOAD` needed).
- No regression on ROCm 7.2.

### DIFF
--- a/src/application/CMakeLists.txt
+++ b/src/application/CMakeLists.txt
@@ -9,6 +9,29 @@ if(WITH_MPI)
   endif()
 endif()
 find_package(hsa-runtime64 REQUIRED)
+
+# ROCm 7.14's hsakmt cmake config declares a link dependency on the imported
+# target `numa::numa` but never provides it (no FindNUMA / find_package(NUMA)).
+# Without it, find_package(hsakmt) fails at generate time: "The link interface
+# of target hsakmt::hsakmt contains: numa::numa but the target was not found."
+# Define it ourselves from the system libnuma before pulling in hsakmt.
+if(NOT TARGET numa::numa)
+  find_library(NUMA_LIBRARY NAMES numa)
+  find_path(NUMA_INCLUDE_DIR NAMES numa.h)
+  if(NUMA_LIBRARY)
+    add_library(numa::numa UNKNOWN IMPORTED)
+    set_target_properties(numa::numa PROPERTIES IMPORTED_LOCATION
+                                                "${NUMA_LIBRARY}")
+    if(NUMA_INCLUDE_DIR)
+      set_target_properties(numa::numa PROPERTIES INTERFACE_INCLUDE_DIRECTORIES
+                                                  "${NUMA_INCLUDE_DIR}")
+    endif()
+  else()
+    message(FATAL_ERROR "libnuma not found but required by hsakmt::hsakmt. "
+                        "Install it (e.g. 'apt install libnuma-dev').")
+  endif()
+endif()
+
 find_package(hsakmt REQUIRED)
 # find_library(IONIC_LIBRARY NAMES ionic HINTS /lib/x86_64-linux-gnu REQUIRED )
 
--- a/src/application/CMakeLists.txt
+++ b/src/application/CMakeLists.txt
@@ -83,17 +83,9 @@ add_library(mori_application SHARED ${MORI_APP_SOURCES})
 
 target_include_directories(mori_application PUBLIC ${CMAKE_SOURCE_DIR}/include)
 target_include_directories(mori_application PUBLIC ${CMAKE_SOURCE_DIR})
-find_library(ROCM_SMI_LIB rocm_smi64 HINTS /opt/rocm/lib REQUIRED)
 find_library(HSA_RUNTIME_LIB hsa-runtime64 HINTS /opt/rocm/lib REQUIRED)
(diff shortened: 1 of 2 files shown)


jhchouuu (commented):
**Overall**: Right direction, should land — but please don't remove the P2P topology capability in this PR.

**✅ CMake numa::numa fix** — LGTM. Defining the imported target from system libnuma to work around ROCm 7.14's missing target is correct.

**✅ rsmi → HIP for bus id** — correct. This avoids rocm-smi's cross-process mutex contention under `/dev/shm/rocm_smi_*` (concurrent `rsmi_init()` races return `RSMI_STATUS_INIT_ERROR`, and `ROCM_SMI_CHECK`'s `exit(-1)` kills every rank). HIP's query APIs are read-only and multi-process safe.

**⚠️ Don't drop the GPU↔GPU P2P topology.** The crash root cause is rsmi's process safety, not P2P collection itself. The rack solution will need this P2P graph (link type/hops/weight) for scale-up routing and cost modeling. Suggest either:
1. **(Preferred)** keep it, swap the source — rebuild via `hipDeviceCanAccessPeer` + `hipDeviceGetP2PAttribute`; or
2. **(Minimum)** keep the `TopoNodeGpuP2pLink` struct and `p2ps` fields with a TODO, drop only the rsmi collection.

Removing the capability is a rack-roadmap decision and shouldn't ride along in a bugfix.

src/application/CMakeLists.txt
jhchouuu: 还需要静态链接吗？看到用了dlopen来打开rocm-smi

src/application/topology/gpu.cpp
```diff
@@ -44,30 +92,54 @@ PciBusId RsmiBusId2PciBusId(uint64_t rsmiBusId) {
...
+#define RSMI_CHECK(stmt)                                                       \
+  do {                                                                         \
+    rsmi_status_t _r = (stmt);                                                 \
+    if (_r != RSMI_STATUS_SUCCESS) {                                           \
+      const char* _msg = nullptr;                                              \
+      rsmi_status_string(_r, &_msg);                                           \
+      fprintf(stderr, "[ROCm-SMI] %s failed: %s\n", #stmt, _msg ? _msg : "?"); \
+      exit(-1);                                                                \
+    }                                                                          \
+  } while (0)
```
jhchouuu: 这个宏只能定义在这里？不能放在check.hpp里嘛

jhchouuu (approved):
LGTM need minor modification

src/application/CMakeLists.txt
jhchouuu: 指的是find_library(ROCM_SMI_LIB rocm_smi64 HINTS /opt/rocm/lib REQUIRED)

src/application/topology/gpu.cpp
QizhouZhang97: 复用之前的 ROCM_SMI_CHECK

src/application/CMakeLists.txt
QizhouZhang97: done,已去掉

PR #427, opened 2026-06-25
fix(cco): bind P2P FD-exchange sockets up front to avoid ENOENT in multi-thread mode

QizhouZhang97:
# Root cause
`ExchangeFileDescriptors` bound each server socket lazily inside a per-peer
loop and called a blocking `accept()` before binding later peers. A client
could only connect once the server's loop reached it. In fork mode ranks enter
near-simultaneously and win the race; in thread mode they serialize on HIP
locks, so a server stuck in `accept()` hadn't yet bound the socket a later
client needed → `ENOENT`, 5s timeout, then a cascading bail.

# Fix
- `local_bootstrap.cpp`: bind+listen **all** server sockets up front, then
  accept/connect; widen the connect retry window.
- `cco_init.cpp`: drop the leader-side blanket `unlink` (could delete a live
  socket) and add a per-clique discriminator to the socket path.

### Testing
`test_gda_modes`: **8 passed, 0 failed**. `test_gda_put` still passes.

### DIFF
--- a/src/cco/cco_init.cpp
+++ b/src/cco/cco_init.cpp
@@ -583,22 +583,27 @@ int ccoWindowRegister(ccoComm* comm, void* ptr, size_t size, ccoWindow_t* outWin
     int p2pWorldSize = static_cast<int>(sortedGroup.size());
 
     // Socket path must agree across the group but be unique per (group, window).
-    // groupId = rank 0's pid; slotOffset identifies the window.
-    std::string socketPath =
-        "/tmp/mori_cco_" + std::to_string(comm->groupId) + "_" + std::to_string(slotOffset) + "_";
+    // groupId = rank 0's pid; slotOffset identifies the window. The clique
+    // leader (smallest GLOBAL rank in the group) must also be part of the path:
+    // a single node can host several disjoint P2P cliques (e.g. GPUs {0,1,2,3}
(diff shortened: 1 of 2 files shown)


src/cco/cco_init.cpp
```diff
@@ -583,22 +583,27 @@ int ccoWindowRegister(ccoComm* comm, void* ptr, size_t size, ccoWindow_t* outWin
...
-        "/tmp/mori_cco_" + std::to_string(comm->groupId) + "_" + std::to_string(slotOffset) + "_";
-
-    // Best-effort cleanup of stale sockets from crashed runs.
-    if (myPeerRank == 0) {
-      for (int i = 0; i < p2pWorldSize; i++) {
-        for (int j = 0; j < p2pWorldSize; j++) {
-          std::string stale = socketPath + std::to_string(i) + "_" + std::to_string(j);
-          unlink(stale.c_str());
-        }
-        unlink((socketPath + "barrier_arrive_" + std::to_string(i)).c_str());
```
jhchouuu, line 597: 为啥这个clean-up要删掉？确认一下？

jhchouuu (approved):
确认一下clean up的细节我觉得就可以了
另外能知道为啥只有mlx的机器上暴露了这个问题？

PR #431, opened 2026-06-26
bugfix: shm spike bandwidth

QizhouZhang97:
### Problem
`Mlx5CollapsedCqDrain` (and the cco `quietUntil` equivalent) reconstructed the
32-bit completion counter from the CQE's low 16 bits by patching the high bits and
adding `0x10000` on an apparent wrap. On a shared QP, a stale/concurrent CQE read
(counter just behind `doneIdx`) looks like a 64K wrap, so `doneIdx` gets shoved
past `postIdx` — after which every `quiet` early-exits and stops waiting for real
completions.

### Fix
Reconstruct as `cons + forward 16-bit distance`, accepting the advance only if it
falls inside the in-flight window `(cons, dbTouchIdx]`:

- Handles 16-bit wrap automatically (small delta added to the full 32-bit `cons`).
- Rejects stale/torn reads (they produce a near-65536 delta, outside the window).
- `window` is **signed**: once `doneIdx` catches up to `dbTouchIdx` (`window <= 0`)
  every delta is rejected — avoids the unsigned underflow that defeated the bound
  when `cons` races ahead of a stale `dbTouchIdx` read.

### Changes
- `shmem/shmem_ibgda_kernels.hpp`, `cco/cco_scale_out.hpp`: windowed signed
  reconstruction (+ low-volume `[DRAIN-BUG]` tripwire).
- Bench default `iters` 10 → 100 (`shmem` & `cco`) to average out a hardware-level
  small-transfer timing jitter.

### Testing
`shmem`/`cco` `p2p_get_bw` and `p2p_put_bw` stable across many runs; tripwire never
fires; data verified correct.

### DIFF
--- a/benchmark/cco/util.hpp
+++ b/benchmark/cco/util.hpp
@@ -55,7 +55,7 @@ enum class Transport { kLsa, kIbgda };
 inline constexpr std::size_t kDefaultMinSize = 8;
 inline constexpr std::size_t kDefaultMaxSize = 64ULL * 1024ULL * 1024ULL;
 inline constexpr std::size_t kDefaultStepFactor = 2;
-inline constexpr std::size_t kDefaultIters = 10;
+inline constexpr std::size_t kDefaultIters = 100;
 inline constexpr std::size_t kDefaultWarmup = 5;
 inline constexpr int kDefaultNumBlocks = 32;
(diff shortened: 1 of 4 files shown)


jhchouuu (approved):
LGTM，用滑窗将race异常情况变得无害

PR #433, opened 2026-06-26
feat(ccl): add param-contiguous SDMA allgather

wuyl1:
## Summary

This PR adds a param-contiguous SDMA allgather path for FSDP zero-copy output.

The new path lets MORI write allgather output directly in the layout expected by FSDP flattened parameter views, while preserving the existing rank-major allgather path.

## Changes

- Add param-contiguous sync and async SDMA allgather kernels.
- Extend allgather JIT args and Python/pybind APIs for split sizes and offsets.
- Add Python wrappers for `enqueue_param_contiguous()` and `start_async_param_contiguous()`.
- Harden SDMA symmetric memory teardown after SHMEM finalize.
- Add an SDMA-gated CI correctness test for the param-contiguous allgather layout.

## Test Plan

- `MORI_ENABLE_SDMA=1 pytest tests/python/ccl/test_allgather_param_contiguous.py -v`
- Added CI step: `MORI-CCL (param-contiguous allgather SDMA)`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,6 +157,12 @@ jobs:
             cd $GITHUB_WORKSPACE && MORI_ENABLE_SDMA=1 timeout 300 pytest tests/python/ops/test_dispatch_combine_async_ll.py -v
           "
 
+      - name: MORI-CCL (param-contiguous allgather SDMA)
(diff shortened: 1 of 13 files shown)


inkcherry (approved):
LGTM

PR #441, opened 2026-07-01
ccl: hierarchical cross-node AllGather (intra-node SDMA + inter-node RDMA)

inkcherry:
## Summary

Adds `mori.ccl.HierAllGather` — a **hierarchical cross-node AllGather primitive**, drop-in compatible with `torch.distributed.all_gather_into_tensor`. Intra-node traffic rides the **SDMA copy engines (XGMI)** and inter-node traffic goes over **RDMA (NIC)**, keeping the collective's data movement off the compute units. Single-node degenerates to the pure intra-node SDMA path; the same object handles multi-node with no signature change.

## Technical details

- Two-leg hierarchical exchange: intra-node SDMA sub-group gather (XGMI) + inter-node RDMA ring over node-blocks.
- Inter-node leg selectable: **host-proxy** (CPU-posted RDMA) or **device IBGDA** (GPU-posted); intra-node leg selectable SDMA or NCCL.
- **Bit-exact** vs `torch.distributed.all_gather_into_tensor` (zero tolerance, `{bf16, fp16, fp32, int32}`).
- Includes an example FSDP2 backend (`MoriAllGather`) that wires the primitive in via `FSDPModule.set_custom_all_gather`.

## Test plan / results

Validated on **two platforms** — **MI300X + Mellanox mlx5** and **MI355X + AINIC (ionic)** RoCEv2.

Full numbers, figures, raw logs, and one-key reproduce scripts are under `examples/fsdp_sdma/bench/` (`results/mi300x_mlx5/`, `results/mi355x_ainic/`; see `examples/fsdp_sdma/README.md`).


wuyl1 (approved):

include/mori/application/context/context.hpp
```diff
@@ -98,6 +98,14 @@ class Context {
   RdmaContext* GetRdmaContext() const { return rdmaContext.get(); }
   RdmaDeviceContext* GetRdmaDeviceContext() const { return rdmaDeviceContext.get(); }
   bool RdmaTransportEnabled() const { return GetRdmaDeviceContext() != nullptr; }
+  // DUAL-RAIL (idle-NIC fan-out): the second RDMA device context (an otherwise-
+  // idle NIC) on which the upper half of each peer's QPs are created. nullptr
+  // unless MORI_HIER_DUAL_RAIL is enabled.
+  RdmaDeviceContext* GetRdmaDeviceContext2() const { return rdmaDeviceContext2.get(); }
```
jhchouuu: why need rdmaDeviceContext2?What is his role?
It seems that I didn't see it being used.

include/mori/application/context/context.hpp
```diff
@@ -98,6 +98,14 @@ class Context {
...
+  bool DualRailEnabled() const { return rdmaDeviceContext2.get() != nullptr; }
+  // Local QP index (within a peer's [0,numQpPerPe) block) at/after which QPs live
+  // on the second rail. Equals numQpPerPe when dual-rail is off (no rail-2 QP).
+  int GetRail2QpStart() const { return rail2QpStart; }
```
jhchouuu, line 101: DualRailEnabled and GetRail2QpStart is same as rdmaDeviceContext2

include/mori/application/transport/rdma/providers/mlx5/mlx5.hpp
```diff
@@ -96,7 +96,7 @@ class Mlx5QpContainer {
 
  private:
   void ComputeQueueAttrs(const RdmaEndpointConfig& config);
-  void CreateQueuePair(uint32_t cqn, uint32_t pdn);
+  void CreateQueuePair(uint32_t cqn, uint32_t pdn, uint32_t cqnRcv);
   void DestroyQueuePair();
```
jhchouuu, line 100: Regarding the recv queue, only the implementations of mlx5 and ionic, while there is no implementation for bnxt.

include/mori/application/application_device_types.hpp
```diff
@@ -103,6 +111,14 @@ struct SymmMemObj {
...
+  // Dual-rail (idle-NIC fan-out): when this symmetric buffer is ALSO registered
+  // on a second RDMA device (an otherwise-idle NIC), lkey2/peerRkeys2 carry that
+  // second MR's local/remote keys. The device put selects these for QP ids that
+  // live on rail 2 (see ShmemPutMemNbiThreadKernelImpl). hasRail2==false (default)
+  // => single-rail, these are never read and the byte path is unchanged.
+  bool hasRail2{false};
+  uint32_t lkey2{0};
+  uint32_t* peerRkeys2{nullptr};
 
   // For VMM allocations: chunk key information (nvshmem-style)
```
jhchouuu, line 107: have you ever test vmm heap？,  the vmm heap is temporarily unavailable currently.
If not, this part should not be modified.

include/mori/core/transport/rdma/providers/mlx5/mlx5_device_primitives.hpp
```diff
@@ -243,6 +243,54 @@ inline __device__ uint64_t PostReadWrite<ProviderType::MLX5, true>(WorkQueueHand
...
+}
+
+template <>
+inline __device__ uint64_t PostWriteImm<ProviderType::MLX5>(
+    WorkQueueHandle& wq, uint32_t curPostIdx, uint32_t curMsntblSlotIdx, uint32_t curPsnIdx,
+    bool cqeSignal, uint32_t qpn, uintptr_t laddr, uint64_t lkey, uintptr_t raddr, uint64_t rkey,
+    uint32_t imm, size_t bytes) {
+  return Mlx5PostWriteImmImpl(wq, curPostIdx, cqeSignal, qpn, laddr, lkey, raddr, rkey, imm, bytes);
+}
+
```
jhchouuu: the same, only mlx5 and ionic impl

include/mori/core/transport/sdma/anvil_device.hpp
```diff
@@ -66,6 +78,11 @@ __device__ __forceinline__ SDMA_PKT_COPY_LINEAR CreateCopyPacket(void* srcBuf, v
   copy_packet.DST_ADDR_LO_UNION.dst_addr_31_0 = (uint32_t)(uintptr_t)dstBuf;
   copy_packet.DST_ADDR_HI_UNION.dst_addr_63_32 = (uint32_t)((uintptr_t)dstBuf >> 32);
 
+  if (cacheHint != 0) {
+    copy_packet.PARAMETER_UNION.dst_ha = (cacheHint & 0x1) ? 1u : 0u;
+    copy_packet.PARAMETER_UNION.src_ha = (cacheHint & 0x2) ? 1u : 0u;
+  }
+
```
jhchouuu: Where did this basis come from?

include/mori/core/transport/sdma/device_primitives.hpp
```diff
@@ -45,23 +46,176 @@ inline __device__ void SdmaPutThread(void* srcBuf, void* dstBuf, size_t copy_siz
...
+// flag op rides the copy engine. Distinct from SdmaPutFencedFlagThread
+// (COPY+FENCE-to-peerFlagAddr). peerFlagAddr/flagVal are unused here.
+inline __device__ void SdmaPutCopySignalThread(void* srcBuf, void* dstBuf, size_t copy_size,
+                                               anvil::SdmaQueueDeviceHandle** deviceHandles,
+                                               HSAuint64* signals, HSAuint64* expectedSignals,
+                                               uint32_t queNum, uint32_t qId, void* peerFlagAddr,
+                                               uint32_t flagVal) {
+  (void)queNum;
+  (void)flagVal;
+  (void)peerFlagAddr;
```
jhchouuu: why declare these parameters？

include/mori/core/transport/sdma/device_primitives.hpp
```diff
@@ -45,23 +46,176 @@ inline __device__ void SdmaPutThread(void* srcBuf, void* dstBuf, size_t copy_siz
...
+  char* dstC = reinterpret_cast<char*>(dstBuf);
+  uint64_t placeOffset = offset;
+  for (size_t i = 0; i < nCopy; ++i) {
+    size_t o = i * kMaxCopy;
+    size_t len = copy_size - o;
+    if (len > kMaxCopy) len = kMaxCopy;
+    auto copy_packet = anvil::CreateCopyPacket(srcC + o, dstC + o, len);
+    handle.template placePacket<SDMA_PKT_COPY_LINEAR>(copy_packet, pendingWptr, placeOffset);
+    placeOffset = 0;  // wrap padding applies only to the first placement
+  }
```
jhchouuu: SdmaPutCopyRemoteAddThread chunks COPY_LINEAR at 8MB ("max single COPY_LINEAR byte count") but SdmaPutFencedFlagThread/SdmaPutCopySignalThread (and the ndesc<=1 SdmaPutThread path) emit a single un-chunked copy. Is 8MB a hard per-descriptor limit? If so, the fenced/signal variants will emit an invalid descriptor for bytesPerPeer > 8MB. Either apply the same cap everywhere, or drop it from RemoteAdd.

include/mori/core/transport/sdma/device_primitives.hpp
```diff
@@ -36,7 +36,8 @@ namespace core {
 inline __device__ void SdmaPutThread(void* srcBuf, void* dstBuf, size_t copy_size,
                                      anvil::SdmaQueueDeviceHandle** deviceHandles,
                                      HSAuint64* signals, HSAuint64* expectedSignals,
-                                     uint32_t queNum, uint32_t qId) {
+                                     uint32_t queNum, uint32_t qId, int ndesc = 1,
```
jhchouuu: What does "ndesc" mean?

include/mori/core/transport/sdma/device_primitives.hpp
```diff
@@ -45,23 +46,176 @@ inline __device__ void SdmaPutThread(void* srcBuf, void* dstBuf, size_t copy_siz
...
+    ++nReal;
+  }
+  if (nReal < 1) nReal = 1;
+
+  const size_t reserveBytes =
+      static_cast<size_t>(nReal) * sizeof(SDMA_PKT_COPY_LINEAR) + sizeof(SDMA_PKT_ATOMIC);
+  base = handle.ReserveQueueSpace(reserveBytes, offset);
   pendingWptr = base;
-  HSAuint64* signal = signals + qId;
+  startBase = base;
```
jhchouuu: Where did this basis come from? any referenced specification?

include/mori/shmem/internal.hpp
```diff
@@ -109,14 +109,52 @@ struct MemoryStates {
...
+  // -1 (default) => single-rail, no QP is on rail 2 and the device put path is
+  // byte-for-byte unchanged. Set from Context by GpuStateInit when dual-rail is on.
+  int rail2QpStart{-1};
+  // this work (transport in-flight depth): when >0, the fast-path (non-VMM) RDMA
+  // put in ShmemPutMemNbiThreadKernelImpl splits a large transfer into multiple
+  // WQEs of at most this many bytes, so several writes stay in flight per QP per
+  // put (the WQ free-entry backpressure at lines ~520-532 already supports many
+  // outstanding WQEs). 0 (default) = one WQE for the whole transfer = the proven
+  // the prior behavior. Set from env MORI_RDMA_PUT_CHUNK_BYTES in GpuStateInit.
+  size_t putChunkBytes{0};
```
jhchouuu: Has putChunkBytes brought any benefits?

include/mori/shmem/internal.hpp
```diff
@@ -109,14 +109,52 @@ struct MemoryStates {
...
+  // this work (drain-free landing fence): mlx5 ctrl-segment fence-mode value OR'd
+  // into fm_ce_se on the fused signal ATOMIC WQE. On RoCE RC a WRITE is NOT ordered
+  // before a following ATOMIC on the same QP, so the DEEP_PIPE put-with-signal AMO
+  // can be executed by the responder before its own large payload WRITE lands
+  // (>=64MB flag-beats-data race). A strong-ordering fence on the atomic WQE makes
+  // it wait for all prior WQEs (incl. the payload write) => the flag never precedes
+  // the landing, WITHOUT the send-CQ quiet-drain round-trip (which serializes the
+  // pipeline back to ~0.88x). 0 (default) = no fence bit = byte-identical shipped
+  // path. Set from env MORI_HIER_SIGNAL_FENCE in GpuStateInit (MLX5 only).
+  int signalFenceMode{0};
```
jhchouuu: In my understanding, the WRITE within the same qp is guaranteed to occur before the subsequent ATOMIC operation. This is the fundamental principle.

include/mori/shmem/internal.hpp
```diff
@@ -109,14 +109,47 @@ struct MemoryStates {
...
+  bool batchPutDoorbell{false};
+  // Drain-free landing fence: mlx5 ctrl-segment fence-mode value OR'd into
+  // fm_ce_se on the fused signal ATOMIC WQE. On RoCE RC a WRITE is NOT ordered
+  // before a following ATOMIC on the same QP, so the DEEP_PIPE put-with-signal AMO
+  // can be executed by the responder before its own large payload WRITE lands
+  // (>=64MB flag-beats-data race). A strong-ordering fence on the atomic WQE makes
+  // it wait for all prior WQEs (incl. the payload write) => the flag never precedes
+  // the landing, without the send-CQ quiet-drain round-trip. 0 (default) = no
+  // fence bit. Set from env MORI_HIER_SIGNAL_FENCE in GpuStateInit (MLX5 only).
+  int signalFenceMode{0};
```
jhchouuu: In my understanding, the WRITE within the same qp is guaranteed to occur before the subsequent ATOMIC operation. This is the fundamental principle.

inkcherry:
todo: remove raw log.

src/application/transport/rdma/providers/ionic/ionic.cpp
```diff
@@ -276,6 +285,21 @@ IonicQpContainer::IonicQpContainer(ibv_context* context, const RdmaEndpointConfi
...
+  if (recv_cq != nullptr && recv_cq != cq) {
+    IonicDvApi::Instance().get_cq(&dvrcq, recv_cq, udma_idx);
+    recv_cq_dbval = dvrcq.q.db_val;
+    recv_cq_mask = dvrcq.q.mask;
+    ionic_recv_cq_buf = reinterpret_cast<ionic_v1_cqe*>(dvrcq.q.ptr);
+    MORI_APP_TRACE("recv cq ptr:0x{:x}, recv cq size:{}, recv cq mask:0x{:x}",
+                   reinterpret_cast<uintptr_t>(dvrcq.q.ptr), dvrcq.q.size, dvrcq.q.mask);
+  }
+
   ionic_dv_qp dvqp;
```
jhchouuu: Ionic also has the "collapsed cqe" mode. However, it determines whether to use the collapsed cqe mode based on the FW version. But for recvCq, the collapsed cqe mode cannot be used, right?

jhchouuu (commented):
I think it's necessary to remove excessive comments and dead code before conducting another review.

include/mori/application/context/context.hpp
inkcherry: removed.

include/mori/application/context/context.hpp
inkcherry: removed.

include/mori/application/application_device_types.hpp
inkcherry: reverted, matches main now.

include/mori/core/transport/sdma/device_primitives.hpp
inkcherry: It was for the CDNA3 ~8MB COPY_LINEAR limit. Tested up to 32MB/copy and full Qwen-7B e2e, bit-exact, no fault — so it's not needed. Removed; single COPY_LINEAR now.

include/mori/application/transport/rdma/providers/mlx5/mlx5.hpp
inkcherry: Dropped the whole recv-CQ/WRITE_WITH_IMM path instead of adding bnxt currently.

include/mori/core/transport/rdma/providers/mlx5/mlx5_device_primitives.hpp
inkcherry: Removed the whole path, same reason

include/mori/core/transport/sdma/anvil_device.hpp
inkcherry: It was an experimental DW2 coherency hint, removed currently.

include/mori/core/transport/sdma/device_primitives.hpp
inkcherry: ndesc split was for the CDNA3 8MB limit; verified unnecessary, removed.

include/mori/core/transport/sdma/device_primitives.hpp
inkcherry: Good catch — dropped the 8MB chunking too, single COPY_LINEAR everywhere now.

include/mori/core/transport/sdma/device_primitives.hpp
inkcherry: 16B was the COPY_LINEAR alignment for the split; gone now.

include/mori/shmem/internal.hpp
inkcherry: Agreed, it's guaranteed — knob was redundant. Removed.

include/mori/shmem/internal.hpp
inkcherry: Same — removed.

src/application/transport/rdma/providers/ionic/ionic.cpp
inkcherry: recv-CQ path removed entirely, and reverted the ionic forceClassic arg it needed.

include/mori/shmem/internal.hpp
inkcherry: removed currently

include/mori/shmem/shmem_sdma_kernels.hpp
```diff
@@ -40,7 +40,8 @@ template <>
 inline __device__ void ShmemPutMemNbiThreadKernel<application::TransportType::SDMA>(
     const application::SymmMemObjPtr dest, size_t destOffset,
     const application::SymmMemObjPtr source, size_t sourceOffset, size_t bytes, int pe, int qpId) {
-  int intraNodePe = pe % 8;
+  int intraNodePe = pe;  // index by GLOBAL pe: deviceHandles_d/signalPtrs/expectSignalsPtr are
+                         // host-populated per global pe (symmetric_memory.cpp)
```
jhchouuu: if you use a global pe id, may be rename intraNodePe  to peSlot?

src/application/context/context.cpp
```diff
@@ -349,10 +349,19 @@ void Context::EnsureSdmaTransport() {
...
   for (int i = 0; i < WorldSize(); i++) {
     if (!peerCaps[i].canSDMA) continue;
-    if (i != LocalRank()) anvil::EnablePeerAccess(LocalRank() % 8, i % 8);
-    anvil::anvil.connect(LocalRank() % 8, i % 8, sdmaNumChannels);
+    // Peer within-node device id: count of same-host peers before it.
+    int peerDevId = 0;
+    for (int j = 0; j < i; j++)
+      if (peerInfos[j].sameHost) peerDevId++;
+    if (i != LocalRank()) anvil::EnablePeerAccess(localDevId, peerDevId);
+    anvil::anvil.connect(localDevId, peerDevId, sdmaNumChannels);
```
jhchouuu, line 370: The same logic as L177? Perhaps the logic can be unified.

src/application/transport/sdma/anvil.cpp
```diff
@@ -146,6 +149,44 @@ static const std::string getBusId(int deviceId) {
...
+  static std::unordered_map<int, int> hipToAgent;
+  std::lock_guard<std::mutex> lock(mapMutex);
+  auto it = hipToAgent.find(hipDeviceId);
+  if (it != hipToAgent.end()) return it->second;
+
+  // BDF of the HIP device, parsed from its "domain:bus:device.function" string.
+  std::string busId = getBusId(hipDeviceId);
+  unsigned domain = 0, bus = 0, dev = 0, func = 0;
+  std::sscanf(busId.c_str(), "%x:%x:%x.%x", &domain, &bus, &dev, &func);
+  uint32_t hipBdf = ((bus & 0xFF) << 8) | ((dev & 0x1F) << 3) | (func & 0x7);
```
jhchouuu, line 173: Has the domain not been used? On multiple PCIe segments, two GPUs on the same machine may share the same 16-bit BDFID → resulting in a mismatch with the correct agent.

include/mori/core/transport/sdma/anvil_device.hpp
```diff
@@ -53,6 +53,7 @@ constexpr bool BREAK_ON_RETRIES = true;
 
 #if defined(__HIPCC__) || defined(__CUDACC__)
 
+// SDMA_PKT_COPY_LINEAR DW2 (PARAMETER_UNION) stays 0: peak D2D/XGMI copy BW.
```
jhchouuu, line 55: clean this comment?

jhchouuu (commented):
For shmem section, I think is acceptable.

jhchouuu:
and agent review FYR:
## Review — hierarchical cross-node AllGather

### Blocking

**1. `deepPipe` is computed independently in Python and C++, and they disagree → OOB write + read of uninitialised memory**

`python/mori/ccl/hier_allgather.py:864-865` applies a 32 MiB coherence gate (sub-chunk ≥ 32 MiB
→ depth 1 → a single flag slot) and the comment states it "Mirrors the C++ HierDeepPipe gate".
`include/mori/collective/ccl_kernel_args.hpp:43` (`HierDeepPipe()`) has **no such gate** and
returns 2 by default. With `chunkBytes ≥ 64 MiB` the ring publishes both `chunkReadyFlags[0]`
and `[1]` (`inter_node/kernels/all_gather.hpp:289`, `:398`) into a one-element tensor: an 8-byte
write past the allocation, and slot 1 is never cleared by `flags.zero_()`, so the reassembly
worker reads uninitialised memory as "landed" and SDMA-pushes ring bytes the NIC has not written
yet — silent wrong data.

Please make the depth single-sourced: either pass the Python-computed depth down, or replicate
the gate in `HierDeepPipe()`.

**2. Reassembly uses `deepPipe` unconditionally, but the ring only engages deep-pipe under six preconditions**

`src/collective/kernels/ccl_kernels.hip:371` does
`const int pipe = args.deepPipe > 1 ? args.deepPipe : rb;` with no check of whether the ring
actually took the deep-pipe path. `inter_node/kernels/all_gather.hpp:117` requires
`deepPipe > 1 && peerIsRdma && prevIsRdma && maxRounds == 1 && !multiBlock && chunkReadyFlags && useWarps >= 1`;
if any of those fails the ring publishes only `chunkReadyFlags[bid]`. A single-node / loopback
run (`!peerIsRdma`) or a P2P neighbour reaches this today, and partition `f ≥ 1` then waits on a
flag nobody ever sets.

Related: the ring clamps `P` by `reqPmax` (`all_gather.hpp:126`) but the reassembly side does
not, so the two also diverge when `chunkBytes < deepPipe MiB`. Suggest deriving both from one
place (engage-decision + clamped `P`).

**3. Bounded spin fallbacks turn a hang into silent corruption**

`src/collective/kernels/ccl_kernels.hip:327` and `:437`:
`if (++spin > 100000000) break;  // bounded fallback`. After the break the code proceeds to push
/ consume the data anyway; at `:327` that means pushing ring bytes over XGMI before the NIC wrote
them. The deep-pipe paths in the same file use `__builtin_trap()` — please do the same here.
Failing loudly is much better than producing wrong results.

**4. The sender-side fence exists at only one of five drain → AMO sites**

`include/mori/collective/allgather/oneshot_sdma_kernel.hpp:201-204` inserts
`__threadfence_system()` between `SubGroupSdmaDrainPe` and the completion-flag AMO, with a
comment explaining that otherwise "the receiver can observe the flag ahead of the data".
Lines **89, 372, 439, 513** have the identical drain → AMO sequence with no fence in between.
(The fences at `:118` / `:231` belong to the receiver-side tail and serve a different purpose.)

Either the fence at `:201` is unnecessary and should go, or these four sites have a
flag-beats-data race. Please make it consistent.

---

### Should fix

**5. `waitForSignal` can give up silently, and polls at the wrong scope** —
`include/mori/core/transport/sdma/anvil_device.hpp:102-116`

It returns `false` after `MAX_RETRIES` (2^30), but `SdmaQueitThread` and `SubGroupSdmaDrainPe`
ignore the return value and go on to fence and set the completion flag. Also the poll is
`__ATOMIC_RELAXED` at AGENT scope on a value written by a separate DMA agent; SYSTEM scope is
what's warranted here.

**6. Ring flag array can overflow with `DEEP_PIPE=auto`** —
`include/mori/collective/inter_node/inter_node_ring_class.hpp:121`

Flags are sized `numBlocks_ * npes_`. Deep-pipe requires `!multiBlock` (so `numBlocks == 1`) and
writes `flagBase + p` for `p ∈ [0, P)` with `P` up to 16. Any configuration where `P > npes`
overflows. Sizing by `max(npes_, kMaxDeepPipe)` would close this.

**7. Importing `mori.ccl` rewrites a process-global env var and prints to stdout** —
`python/mori/ccl/hier_allgather.py:61`, called at import time on `:70`

`_clamp_sdma_channels()` overwrites `MORI_SDMA_NUM_CHANNELS`. That variable is read by the core
SDMA transport (`anvil.hpp:161`, `context.cpp:254`), i.e. by subsystems that have nothing to do
with `mori.ccl` — so merely importing the package silently changes their behaviour. The cap
`MORI_SDMA_CH_HW_MAX = 8` (`:49`) is also hardcoded as an MI300X value rather than probed.

Since `#501` already clamps to `kMaxSdmaChannelsPerPair = 8` on the C++ side, I'd drop this block
entirely after the rebase.

**8. Constructor mutates `os.environ`** — `python/mori/ccl/hier_allgather.py:364`

`os.environ["MORI_HIER_DEEP_PIPE_QUIET"] = "0"` is a process-wide side effect whose value depends
on the *first* handle's `ranks_per_node`; handles constructed later in the same process inherit
it. An instance attribute would be equivalent and order-independent. (Commit `d0a389a5` claims
this class of mutation was removed — this one survived.)

**9. A host-proxy error leaves peers hung in the barrier** —
`python/mori/ccl/host_proxy_ag.py:446-452`

On `wait_all` failure the rank raises, but the other ranks are blocked in
`dist.barrier(self._pair_barrier)` and will wait forever. Consider signalling the failure to the
pair before raising, or bringing the barrier into the error path. `_timeout_ms = 60000` (`:101`)
is also hardcoded — worth making configurable.

include/mori/shmem/shmem_sdma_kernels.hpp
inkcherry: yes, renamed

src/application/context/context.cpp
inkcherry: Unified into a ```SameHostPeersBefore()``` helper.

include/mori/core/transport/sdma/anvil_device.hpp
inkcherry: cleand up

src/application/transport/sdma/anvil.cpp
inkcherry: HSA BDFID has no domain, so I match on the 16-bit BDF only when it's unique, else fall back to identity. No change on single-segment; avoids the wrong pick on multi-segment.

jhchouuu (approved):
LGTM
add some CI test if possible?

inkcherry:
thanks @jhchouuu , added to ci and passed, verified E2E again locally

PR #445, opened 2026-07-02
feat(shmem/sdma): implement address-based device putmem_nbi_signal on…

zjing14:
… the SDMA transport

The device-side ShmemPutMemNbiSignalBlockKernel<SDMA> address-based overloads were stubs (TODO), and DISPATCH_TRANSPORT_TYPE_WITH_BOOL asserted on SDMA, so a device putmem_nbi_signal targeting an SDMA peer hit assert(false).

This wires SDMA through the bool dispatch and implements the address-based signal-put as a COPY_LINEAR(source -> peer dest) followed by an ATOMIC on the peer flag, enqueued on the SAME SDMA queue so the DMA engine executes the flag update strictly after the copy completes (in-order queue). Result: a CU-free per-tile push+signal usable to drive an in-kernel wait_until gate (e.g. a fused all-gather + GEMM consumer) without consuming compute units.

Notes:
- Thread scope resolves the heap object from globalGpuStates and computes peer offsets from heapBaseAddr, matching the existing address-based SDMA paths.
- Uses atomic INCREMENT (monotonic-generation semantics); consumers wait `flag >= gen`. signalValue/signalOp are accepted for API parity but the SDMA path currently only wraps increment.
- onlyOneSignal=false forwards to the =true path.

Validated on 4x gfx950 (MI350) with a FlyDSL SDMA-signal probe and a fused AG+GEMM PoC.

## Submission Checklist

### DIFF
--- a/include/mori/shmem/shmem_device_api.hpp
+++ b/include/mori/shmem/shmem_device_api.hpp
@@ -54,6 +54,8 @@ namespace shmem {
     func<application::TransportType::RDMA, boolParam>(__VA_ARGS__);               \
   } else if (transportType == application::TransportType::P2P) {                  \
     func<application::TransportType::P2P, boolParam>(__VA_ARGS__);                \
+  } else if (transportType == application::TransportType::SDMA) {                 \
+    func<application::TransportType::SDMA, boolParam>(__VA_ARGS__);               \
   } else {                                                                        \
     assert(false);                                                                \
   }
--- a/include/mori/shmem/shmem_sdma_kernels.hpp
+++ b/include/mori/shmem/shmem_sdma_kernels.hpp
@@ -358,30 +358,54 @@ template <>
 inline __device__ void ShmemPutMemNbiSignalBlockKernel<application::TransportType::SDMA, true>(
     const void* dest, const void* source, size_t bytes, const void* signalDest,
     uint64_t signalValue, core::atomicType signalOp, int pe, int qpId) {
-  // TODO: add SDMA PutMemNbiSignal (address-based)
-  (void)dest;
-  (void)source;
-  (void)bytes;
-  (void)signalDest;
-  (void)signalValue;
-  (void)signalOp;
-  (void)pe;
-  (void)qpId;
+  // SDMA push + signal: COPY_LINEAR(source -> peer dest) then an ATOMIC on the
+  // PEER flag, enqueued on the SAME SDMA queue so the DMA engine executes the
+  // flag update strictly after the copy completes (in-order queue) -- a CU-free
+  // per-tile signal. NOTE: uses atomic INCREMENT (monotonic-generation
+  // semantics); consumers should wait `flag >= gen`. signalValue/signalOp are
+  // accepted for API parity but the SDMA path only wraps inc today.
+  if (bytes == 0) return;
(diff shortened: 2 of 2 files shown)


include/mori/shmem/shmem_sdma_kernels.hpp
```diff
@@ -358,30 +358,54 @@ template <>
 inline __device__ void ShmemPutMemNbiSignalBlockKernel<application::TransportType::SDMA, true>(
     const void* dest, const void* source, size_t bytes, const void* signalDest,
     uint64_t signalValue, core::atomicType signalOp, int pe, int qpId) {
```
jhchouuu, line 360: Only the block-level version is implemented; the thread and warp versions aren't yet. The macro expansion probably expands the thread/warp paths too, so there might be a small issue here. But it's fine to merge for now, we'll add the complete functionality in a follow-up.

include/mori/shmem/shmem_device_api.hpp
```diff
@@ -54,6 +54,8 @@ namespace shmem {
     func<application::TransportType::RDMA, boolParam>(__VA_ARGS__);               \
   } else if (transportType == application::TransportType::P2P) {                  \
     func<application::TransportType::P2P, boolParam>(__VA_ARGS__);                \
+  } else if (transportType == application::TransportType::SDMA) {                 \
+    func<application::TransportType::SDMA, boolParam>(__VA_ARGS__);               \
```
jhchouuu, line 58: This adds a bit more device-side runtime overhead to this API. Previously the branch was just between P2P and IBGDA; now there's an extra SDMA branch, which also grows the code size. There's not much we can do about it for now, it's inherent to runtime dispatch. We'll look into addressing this in a follow-up.

include/mori/shmem/shmem_sdma_kernels.hpp
```diff
@@ -358,30 +358,54 @@ template <>
...
+
+    // 1) COPY + queue completion atomic (for quiet), thread-scoped.
+    core::SdmaPutThread(srcPtr, dstPtr, bytes, devicehandles, signalAddr, expectedSignals,
+                        heapObj->sdmaNumQueue, qpId);
+    // 2) Peer-flag atomic on the SAME queue -> ordered after the copy.
+    anvil::SdmaQueueDeviceHandle handle = **(devicehandles + qpId);
+    uint64_t off = 0;
+    uint64_t base = handle.ReserveQueueSpace(sizeof(SDMA_PKT_ATOMIC), off);
+    uint64_t wptr = base;
+    auto pkt = anvil::CreateAtomicIncPacket(sigPtr);
```
jhchouuu, line 393: Only the increment semantics are implemented here, signalValue/signalOp are ignored (always atomic += 1). Fine for now; we'll complete the remaining ops in a follow-up.

jhchouuu (approved):
LGTM.

Merging is fine.  Follow-ups for us to complete later:
- Implement the thread/warp scopes (only block-level is done now; the macro also expands those, so they currently hit a silent no-op stub).
- Support the full signal semantics (signalValue/signalOp are currently ignored, always += 1).
- Reduce the runtime dispatch overhead from the added SDMA branch (extra cost and code size).

jhchouuu:
And now CI system has some issues, is being addressed.

PR #457, opened 2026-07-07
fix: only define true/false for C in bnxt_re_hsi.h

pemeliya:
This issue has popped up during MORI bazelification for XLA


jhchouuu (approved):
LGTM

PR #459, opened 2026-07-09
test(ep): add configurable CLI knobs to dispatch/combine benchmark

amd-arozanov:
## Motivation

The dispatch/combine benchmark script only supported a fixed set of scenarios. This makes it hard to explore different kernel implementations, routing patterns, or tune benchmark timing without editing the script. This PR adds CLI flags to cover these cases.

## Technical Details

Adds the following flags to `--cmd bench` (all optional, defaulting to prior behavior):
- `--warmup` / `--iters` – number of warmup and timing iteration.
- `--routing` – select the token-to-expert routing pattern; adds a new `remote_round_robin` pattern that always routes to a remote rank.
- `--verify` – skip the host-side correctness check during warmup.
- `--graph-replay-iters` – control how many times each captured CUDA graph is replayed per timed sample.

## Test Plan

Run bench_dispatch_combine.py with new command line options

## Test Result

bench_dispatch_combine.py works, default values were not changed, new command line arguments work as expected

## Submission Checklist

### DIFF
--- a/tests/python/ops/bench_dispatch_combine.py
+++ b/tests/python/ops/bench_dispatch_combine.py
@@ -52,8 +52,10 @@ class EpDispatchCombineBenchmark(EpDispatchCombineTestCase):
         force_scale_active=False,
         report_scale_stats=False,
         combine_scale_dim=None,
+        routing="random",
     ):
         super().__init__(config)
+        self.routing = routing
         self.combine_data_type = (
             combine_data_type if combine_data_type is not None else config.data_type
         )
@@ -83,6 +85,7 @@ class EpDispatchCombineBenchmark(EpDispatchCombineTestCase):
         self.config.hidden_dim = self.dispatch_hidden_dim
         result = super().gen_test_data(
             use_max_token_num=True,
+            routing=self.routing,
             input_dist=self.input_dist,
             input_scale=self.input_scale,
             input_shift=self.input_shift,
@@ -908,6 +911,7 @@ def _bench_dispatch_combine(
     warmup=None,
     iters=None,
     verify=True,
+    routing="random",
 ):
     if combine_data_type is None:
         combine_data_type = data_type
@@ -973,6 +977,7 @@ def _bench_dispatch_combine(
             force_scale_active=force_scale_active,
             report_scale_stats=report_scale_stats,
             combine_scale_dim=bench_combine_scale_dim,
+            routing=routing,
         )
 
         (
@@ -1252,6 +1257,7 @@ def bench_dispatch_combine(
     warmup=None,
     iters=None,
     verify=True,
+    routing="random",
 ):
     if combine_data_type is None:
         combine_data_type = dtype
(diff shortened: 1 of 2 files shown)


amd-arozanov:
Hello, can someone take a look and review this PR? 
@QizhouZhang97

tests/python/ops/bench_dispatch_combine.py
```diff
@@ -1537,4 +1652,10 @@ def bench_dispatch_combine(
         input_shift=args.input_shift,
         force_scale_active=bool(args.force_scale_active),
         report_scale_stats=bool(args.report_scale_stats),
+        warmup=args.warmup,
+        iters=args.iters,
+        verify=bool(args.verify),
+        routing=args.routing,
+        kernel_type=args.kernel_type,
```
QizhouZhang97: Thanks for the PR. One request on the `--kernel-type` 

`examples/ops/dispatch_combine/test_dispatch_combine_internode.py` is the script for inter-node kernels, and it already supports `--kernel-type` (`v0`/`v1`/`v1_ll`/`async_ll`). `bench_dispatch_combine.py` is written for intranode only, so it shouldn't take this argument — please drop that commit.

The others look good.

QizhouZhang97 (changes requested):

tests/python/ops/bench_dispatch_combine.py
amd-arozanov: Thanks for the review! I see there are two intra node kernels. The four internode values are gone.
--kernel-type now accepts only IntraNode and IntraNodeLL

Also found, that _save_intranode_tuning_result hardcoded "IntraNode" in the metadata and in the auto filename, so --cmd tune --kernel-type IntraNodeLL would have written LL numbers into the IntraNode config file. The kernel type is now threaded through to both.

tests/python/ops/bench_dispatch_combine.py
```diff
@@ -1033,6 +1071,7 @@ def _bench_dispatch_combine(
                 combine_block_num=combine_block_num,
                 combine_warp_per_block=combine_warp_per_block,
                 call_local_expert_count=call_local_expert_count,
+                **_optional_kwargs(warmup=warmup, capture_iters=iters),
```
QizhouZhang97, line 1065: stress() accepts neither warmup nor capture_iters, so --cmd stress --iters N raises TypeError,  suggest just dropping this line

amd-arozanov: This line is in the `elif cmd == "profile"` branch, not the stress one, and profile() does declare both warmup=5 and capture_iters=3 (lines 626-627). benchmark.stress() on line 1051 is called without those kwargs, so --cmd stress --iters N doesn't raise, it just ignores the flags, which matches the --warmup/--iters help text ("Applies to --cmd bench and --cmd profile")

QizhouZhang97: I see, thanks for your contributing

QizhouZhang97:
the PR will be merged after ci passing

QizhouZhang97:
please resolve the conflict @amd-arozanov

amd-arozanov:
@QizhouZhang97, rebase is done

PR #462, opened 2026-07-10
ci: recover podman state before internode build on MI355X_AINIC

QizhouZhang97:
The MI355X_AINIC internode job uses podman and occasionally hits "invalid internal status" at `podman build`, caused by stale pause-process state from a previous run. 
FIX: add a `podman system migrate || true` step before Build CI image to self-heal without manual intervention.


jhchouuu (approved):

QizhouZhang97:
image

PR #463, opened 2026-07-10
feat(cco): add fabric handle support for gfx1250

kawhil-amd:
## Summary

  - Add fabric handle support to CCO P2P memory sharing, following RCCL's pattern
  - When the HIP runtime supports fabric handles (hipMemHandleTypeFabric, 0x8), CCO uses 64-byte opaque tokens exchanged via Allgather instead of dma-buf
  FDs passed through Unix sockets + SCM_RIGHTS
  - Falls back transparently to the existing FD-based LocalBootstrapNetwork path when fabric is unavailable (runtime probe at ccoCommCreate)

  ## Motivation

  The current dma-buf FD exchange requires per-pair Unix domain socket setup (socket path construction, leader disambiguation, SCM_RIGHTS ancillary data).
  Fabric handles reduce this to a single bootNet->Allgather of 64-byte tokens — simpler, faster, and required for upcoming gfx1250 / UALoE interconnect.

  ## Changes

  - include/mori/utils/hip_compat.hpp: Add hipMemFabricHandle_compat_t and hipMemHandleTypeFabricCompat compat shim (uses native types when HIP_FABRIC_API
  is defined)
  - include/mori/cco/cco.hpp: Add ccoFabricHandle_t, change AllocMeta to union of shareFd/fabricHandle with isFabric flag, add handleType to ccoComm
  - src/cco/cco_init.cpp:
    - ccoCommCreate: probe fabric support (alloc + export test), set handleType
    - ccoMemAlloc: use probed handle type for allocation and export
    - ccoWindowRegister: fabric path uses Allgather + hipMemImportFromShareableHandle; FD path unchanged as fallback. RDMA MR uses
  RegisterRdmaMemoryRegionAuto (ibv_reg_mr) in fabric mode
    - ccoMemFree/ccoCommDestroy: skip close(fd) when using fabric handles

### DIFF
--- a/include/mori/cco/cco.hpp
+++ b/include/mori/cco/cco.hpp
@@ -80,6 +80,13 @@ struct SdmaQueueDeviceHandle;  // ccoSdmaContext / ccoComm (pointer)
 // Opaque HIP typedef, replicated so ccoComm can name it without the ROCm header.
 struct ihipMemGenericAllocationHandle;
 typedef struct ihipMemGenericAllocationHandle* hipMemGenericAllocationHandle_t;
+// Fabric handle — 64-byte opaque token for cross-process sharing.
+// Intentionally a separate type from hipMemFabricHandle_compat_t (hip_compat.hpp)
+// so this header stays self-contained; the two are layout-compatible and
+// cco_init.cpp reinterpret_casts between them.
+struct ccoFabricHandle_t {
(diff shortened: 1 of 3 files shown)


kawhil-amd:
This feature has supported in pr https://github.com/ROCm/mori/pull/466, so close it.

PR #469, opened 2026-07-13
fix(jit): invalidate kernel cache on #included source changes

Oseltamivir:
## Problem
The JIT content hash for a kernel covers only its top-level `.hip` file plus `include/mori`. But the ops kernel `.hip` translation units `#include` sibling sources — `src/ops/dispatch_combine/*.cpp` (e.g. `internode_v1.cpp`, `low_latency_async.cpp`) and `src/ops/kernels/ep_common.hip`. Editing one of those included files leaves the hash unchanged, so a stale `.hsaco` is reused and the change silently has no effect. Today the only workaround is deleting `~/.mori/jit/<arch>_<nic>/` by hand.

## Fix
- `compile_genco`: hash the whole subsystem source tree (parent of the kernel dir) instead of only the top-level `.hip`.
- `_hash_tree`: include `.hip` files when hashing directories, so `#include`d translation units (e.g. `ep_common.hip`) are covered.

## Verification
Recomputing the hash over the source tree while editing an `#include`d file: before the change the hash is unchanged (stale reuse); after, it changes for both `dispatch_combine/*.cpp` and `ep_common.hip`. No tests reference the hashing functions.

### DIFF
--- a/python/mori/jit/cache.py
+++ b/python/mori/jit/cache.py
@@ -42,7 +42,8 @@ def get_cache_root() -> Path:
 def _hash_tree(paths: list[Path]) -> str:
     """Compute a short content hash over files and directories.
 
-    For directories, all ``.hpp``, ``.h``, and ``.cpp`` files are included.
+    For directories, all ``.hpp``, ``.h``, ``.cpp``, and ``.hip`` files are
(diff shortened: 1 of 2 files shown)


jhchouuu (approved):
LGTM
thanks for catching this issue

PR #478, opened 2026-07-15
cco ops-v2: tuned geometry by default + gfx1250 EP4 retune + warmup hang fix

jhchouuu:
## Summary
- **Tuned geometry by default**: `EpDispatchCombineConfig.__post_init__` auto-pulls the per-device tuned schedule (block/warp) when geometry isn't pinned; bench gains an `AUTO` mode to exercise it.
- **gfx1250 EP4 retune**: finer block×warp sweep for EP4 bf16 (combine warp2 at small tok, block96 mid, vec4-aware); dedup merged blocks.
- **combine gather perf**: inner-unroll load-first scheduling.
- **EP8 graph-smoke hang fix** (`bench_dispatch_combine.py`): the dispatch/combine cross-device (xdb) barrier matches per-call flags across ranks; if one rank gets a full call ahead it overwrites the flag a slower rank is still waiting on → deadlock. The warmup loops fired the kernel back-to-back with no per-iter cross-rank sync, so at small token counts (short kernels, e.g. ct=128) rank drift **intermittently hung the EP8 smoke**. Fix: `sync()` + `comm.barrier()` inside the warmup loops so no rank starts call N+1 before all ranks finished N. Warmup is untimed → measured replay bandwidth unchanged.

## Verification (MI355X, 8 ranks, `MODE=graph SWEEP=128,512,4096,8192`)
- Hang: **8/8 runs clean** with the fix (was ~1-in-2 hang before).
- Perf: identical to baseline (warmup is untimed).

### DIFF
--- a/python/mori/ops/dispatch_combine_v2/dispatch_combine_op.py
+++ b/python/mori/ops/dispatch_combine_v2/dispatch_combine_op.py
@@ -125,16 +125,18 @@ class EpDispatchCombineConfig:
     # "gather" (UseP2PRead) or "scatter" (mori _nop2p, fp8 compression home).
     combine_mode: str = "gather"
     quant_type: str = "none"  # none | fp8_direct_cast | fp8_blockwise
-    dispatch_block_num: int = 64
-    combine_block_num: int = 80
-    warp_num_per_block: int = 16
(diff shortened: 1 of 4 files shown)


kawhil-amd (approved):

kawhil-amd:
This pr has file conflict.

PR #484, opened 2026-07-16
feat(EPv2): combine gather inner-unroll + tuned geometry by default + gfx1250 EP4 retune

jhchouuu:
Copy of #478's branch (`jiahzhou/cco-ep-tuned`, kept untouched for others to consume), merged up to current `main` with conflicts resolved. Net change vs `main` is exactly the EP-tuning work below.

## Summary
- **Tuned geometry by default**: `EpDispatchCombineConfig.__post_init__` auto-pulls the per-device tuned schedule (block/warp) when geometry isn't pinned; bench gains an `AUTO` mode to exercise it.
- **gfx1250 EP4 retune**: finer block×warp sweep for EP4 bf16 (combine warp2 at small tok, block96 mid, vec4-aware); dedup merged blocks.
- **combine gather perf**: inner-unroll load-first scheduling.
- **EP8 graph-smoke hang fix** (`bench_dispatch_combine.py`): the dispatch/combine cross-device (xdb) barrier matches per-call flags across ranks; if one rank gets a full call ahead it overwrites the flag a slower rank is still waiting on → deadlock. The warmup loops fired the kernel back-to-back with no per-iter cross-rank sync, so at small token counts (short kernels, e.g. ct=128) rank drift **intermittently hung the EP8 smoke**. Fix: `sync()` + `comm.barrier()` inside the warmup loops so no rank starts call N+1 before all ranks finished N. Warmup is untimed → measured replay bandwidth unchanged.

## Merge conflict resolution
- `dispatch_combine_op.py` `EpDispatchCombineConfig.tuned`: kept this branch's tuned-by-default docstring and adopted main's relative `.tuning_configs` import (from the v2-packaging change on main, #482). Also converted the tuned-by-default `__post_init__` lookup to a relative import so it works from the now-packaged module.

## Verification (MI355X, 8 ranks, `MODE=graph SWEEP=128,512,4096,8192`)
- Hang: **8/8 runs clean** with the fix (was ~1-in-2 hang before).
- Perf: identical to baseline (warmup is untimed).

Co-authored-by: kawhil-amd <kaiclian@amd.com>

### DIFF
--- a/python/mori/ops/dispatch_combine_v2/dispatch_combine_op.py
+++ b/python/mori/ops/dispatch_combine_v2/dispatch_combine_op.py
@@ -125,16 +125,18 @@ class EpDispatchCombineConfig:
     # "gather" (UseP2PRead) or "scatter" (mori _nop2p, fp8 compression home).
     combine_mode: str = "gather"
     quant_type: str = "none"  # none | fp8_direct_cast | fp8_blockwise
-    dispatch_block_num: int = 64
-    combine_block_num: int = 80
-    warp_num_per_block: int = 16
-    # Combine wants few warps (K-deep per-lane MLP already saturates); kept
-    # separate from dispatch's warp count.
-    combine_warp_num_per_block: int = 4
+    # Geometry: None => auto (pull the tuned schedule for this device/shape/dtype
+    # from tuning_configs in __post_init__). Pin any of these to opt out and use a
+    # fixed geometry instead. Combine wants few warps (K-deep per-lane MLP already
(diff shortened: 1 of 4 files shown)


jhchouuu:
Copied from #478 (branch `jiahzhou/cco-ep-tuned`, which is left untouched for others to consume) and merged up to current `main` with conflicts resolved. This PR is the mergeable version; #478 stays as the shared branch.

jhchouuu:
cc @kawhil-amd

PR #493, opened 2026-07-21
fix(bootstrap): make rendezvous timeout configurable via MORI_BOOTSTRAP_TIMEOUT

Lzy17:
## Summary                                                                                                                                
                                                                                                                                            
  The socket bootstrap rendezvous has four hardcoded timeouts that together cap              
  each rank's wait for the root at roughly **10 s** (the retry-based sites are                                                              
  `50 retries x 200 ms`, despite in-code comments claiming `30 s`). When the root            
  rank is slow to reach `accept()` — e.g. it is busy with model initialization in                                                           
  a large MoE launch — non-root ranks exhaust their retries and die with               
  `failed to connect to root after 50 retries` / `PhoneHomeProtocol failed`, even                                                           
  though the root is healthy and would have been reachable moments later.                    
                                                                                                                                            
  This makes intra-node bootstrap a coin-flip under load: on a busy node the                                                                
  workers can lose the race against a slow-starting root.                              
                                                                                                                                            
  ## Change                                                                                  
                                                                                       
  All four bootstrap timeout sites in `socket_bootstrap.cpp` now derive from a                                                              
  single, env-configurable budget `MORI_BOOTSTRAP_TIMEOUT` (seconds, default           
  **300 s**):                                                                                                                               
                                                                                                                                            
  - `SetupCommunicationRing()` ring-connect retries                                                                                         
  - `SetupCommunicationRing()` ring-accept timeout                                                                                          
  - `PhoneHomeProtocol()` root-side per-rank accept timeout                                                                                 
  - `PhoneHomeProtocol()` non-root connect-to-root retries                                                                                  
                                                                                                                                            
  A small lazily-initialized helper reads the env var once, validates it (positive                                                          
  integer, else falls back to the default), and the retry-based sites compute their                                                         
  retry count from the budget. This also fixes the stale `30 s` comments, which                                                             
  never matched the actual `~10 s` retry behavior.                                
                                                                                                                                            
  The default of 300 s keeps behavior safe (bounded, no infinite hang) while giving          
  a slow-starting root plenty of headroom; unset env = default, so existing                                                                 
  deployments only see the larger budget, not a behavior change in the success path.         
                                                                                                                                            
  ## Validation                                                                              
                                                                                                                                            
  Built into `libmori_application.so` and hardware-tested on an 8-GPU-per-node                                                              
  MI35x MoE disaggregation launch (DeepSeek wide-EP16, mori for both MoE               
  all-to-all and KV): with the stock ~10 s budget, prefill engines intermittently                                                           
  lost the bootstrap race (both engines failed on a busy node in one run); with              
  `MORI_BOOTSTRAP_TIMEOUT=300` the same topology bootstrapped cleanly with zero                                                             
  races across all ranks and passed the end-to-end accuracy gate (GSM8K 0.927).

### DIFF
--- a/src/application/bootstrap/socket_bootstrap.cpp
+++ b/src/application/bootstrap/socket_bootstrap.cpp
@@ -33,6 +33,7 @@
 #include <algorithm>
 #include <cassert>
 #include <chrono>
+#include <cstdlib>
 #include <cstring>
 #include <iostream>
 #include <random>
@@ -43,6 +44,30 @@
 
 namespace mori {
 namespace application {
+
+namespace {
+// Total wall-clock budget (ms) for each bootstrap rendezvous step (non-root
+// connect-to-root, ring connect, and the root-side accepts). Historically these
+// were fixed at 50 retries x 200 ms ~= 10 s, which is too tight when the root
+// rank is slow to reach accept() (e.g. busy with model init in a large MoE
+// launch), producing spurious "failed to connect to root" bootstrap failures.
+// Configurable via MORI_BOOTSTRAP_TIMEOUT (seconds); defaults to 300 s to
+// tolerate slow-starting roots while still bounding the wait to prevent hangs.
+int BootstrapTimeoutMs() {
+  static const int kMs = [] {
+    int secs = 300;
+    if (const char* s = std::getenv("MORI_BOOTSTRAP_TIMEOUT")) {
+      try {
+        int v = std::stoi(s);
+        if (v > 0) secs = v;
+      } catch (...) {
+      }
+    }
+    return secs * 1000;
+  }();
+  return kMs;
+}
+}  // namespace
 
 SocketBootstrapNetwork::SocketBootstrapNetwork(const UniqueId& unique_id, int rank, int world_size)
     : unique_id_(unique_id), initialized_(false), unexpected_connections_(nullptr) {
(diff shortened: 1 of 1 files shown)


src/application/bootstrap/socket_bootstrap.cpp
```diff
@@ -44,6 +45,30 @@
...
+  static const int kMs = [] {
+    int secs = 300;
+    if (const char* s = std::getenv("MORI_BOOTSTRAP_TIMEOUT")) {
+      try {
+        int v = std::stoi(s);
+        if (v > 0) secs = v;
+      } catch (...) {
+      }
+    }
+    return secs * 1000;
```
jhchouuu, line 64: Integer overflow risk

src/application/bootstrap/socket_bootstrap.cpp
```diff
@@ -44,6 +45,30 @@
...
 namespace application {
 
+namespace {
+// Total wall-clock budget (ms) for each bootstrap rendezvous step (non-root
+// connect-to-root, ring connect, and the root-side accepts). Historically these
+// were fixed at 50 retries x 200 ms ~= 10 s, which is too tight when the root
+// rank is slow to reach accept() (e.g. busy with model init in a large MoE
+// launch), producing spurious "failed to connect to root" bootstrap failures.
+// Configurable via MORI_BOOTSTRAP_TIMEOUT (seconds); defaults to 300 s to
+// tolerate slow-starting roots while still bounding the wait to prevent hangs.
```
jhchouuu: pls help to reduce some of the comments.

jhchouuu (approved):
LGTM, some minor  suggestions
thanks!

src/application/bootstrap/socket_bootstrap.cpp
Lzy17: done, thank you!!

src/application/bootstrap/socket_bootstrap.cpp
Lzy17: done, thank you!!

PR #499, opened 2026-07-24
feat(tools): make Ionic QoS setup configurable

QizhouZhang97:
## Summary

- Move Ionic/Pollara PFC, DSCP, scheduling, and DCQCN parameters into an editable `.mori.conf`.
- Bundle the default configuration with the Python package and copy it to `~/.mori.conf` on first `mori setup`.
- Add `mori setup --config FILE` and `MORI_QOS_CONFIG` overrides.
- Reset stale Ionic QoS state before applying the configured mappings, allowing priority changes such as DSCP 48 from priority 5 to 6.
- Avoid requiring `sudo` when running as root inside a container.

BNXT and MLX5 configuration paths are unchanged.

## Details

Configuration values are parsed using a strict key whitelist and validated before modifying the NIC. The same `IONIC_CNP_DSCP` value is used for both DSCP-to-priority mapping and DCQCN CNP generation to prevent inconsistent configurations.

Default lookup order:

1. `mori setup --config FILE`
2. `MORI_QOS_CONFIG`
3. `~/.mori.conf`


python/mori/cli.py
```diff
@@ -136,6 +141,14 @@ def _cmd_setup(rest: List[str]) -> int:
             "(use with `source $(mori setup --path)` to export env vars)"
         ),
     )
+    parser.add_argument(
+        "--config",
+        metavar="FILE",
+        help=(
+            "Ionic QoS configuration path (default: ~/.mori.conf; "
+            "created from the bundled template on first use)"
```
jhchouuu: 我建议放在 ~/.mori/下面 我们jit kernel也是放在这个下面的

jhchouuu (approved):
LGTM 
one minor suggestion

PR #502, opened 2026-07-27
feat(cco): fix the SDMA ring, rework signals and multi-lane thread scope

jhchouuu:
Reworks the CCO SDMA device path: two ring-buffer correctness fixes, a completion/signal rework, safe multi-lane thread scope, and a ~2x faster trigger.

## Correctness

- **Slots could be written past the end of the ring.** Slots were 64B/128B and placed linearly from the wrapped offset, so once a queue carried mixed slot sizes a slot could straddle the end of the queue buffer. A slot is now a whole number of 32B units and each unit start is wrapped individually, so a packet is never split. This survives mixed slot sizes, which "slot size divides the ring" would not.
- **The read-index hint was never published.** `cachedHwReadIndex` is a by-value field, so a refreshed `rptr` was thrown away and every put re-read it — an uncached SYSTEM load, paid once the monotonic wptr passes the ring size.
- **Several lanes of one wavefront on one queue deadlocked.** Lock-step prevents the lane that must publish `committedWptr` from leaving the spin loop until the lane waiting on it does. Lanes are now grouped per queue, and the group leader makes one reservation and rings once.
- **The Aggregate solo post never issued `s_waitcnt`**, so a `commit()` from another wave could ring over a half-written packet.

## API changes (breaking)

- `ccoSdmaOptFlags` is a template parameter instead of a runtime argument. The C ABI is unchanged — `cco_device_wrapper.cpp` selects the instantiation — so the Python/FlyDSL `aggregate` kwarg still works.
- `expectSignals` is removed from `ccoSdmaContext`. Completion no longer depends on signals: `quiet`/`quietQueue` drain via `rptr`/`committedWptr` and work whatever a put's signal targeted.
- Signal slots are indexed by **sender**: `signalBuf[srcLsaRank * numQueue + queueId]`. `put`/`get` take independent `localSignal` and `remoteSignal` template flags; the receiver polls with `waitSignal`.
- New `ccoSdmaThreadMode`. `SameQueue` promises every active lane shares `(peer, queueId)` and skips the runtime matching; `Independent` (default) groups lanes at runtime.
- `quiet` and `commit` are templated on coop scope.

## Performance (MI355X, gfx950)

Put trigger cost and register pressure, measured against `origin/main` with the same harness:

| scope | trigger, no signal | trigger, one signal | SGPR, no signal | SGPR, one signal |
|---|---|---|---|---|
| thread | 6500 -> 2500 cyc | 9400 -> 2720 cyc | 48 -> 37 | 50 -> 42 |
| warp | 6900 -> 2500 cyc | 12800 -> 2695 cyc | 64 -> 37 | 70 -> 42 |
| block | 9400 -> 2500 cyc | 12900 -> 2695 cyc | 64 -> 37 | 70 -> 42 |

The trigger is now flat across scopes and nearly flat in the number of signals; on main it grew with both. VGPR is unchanged without a signal (25 -> 26) and higher with one (27 -> 36), which buys the 32B-unit slot fill.

With `ccoSdmaThreadSameQueue`, 8 lanes of one wave on one queue cost 2612 cyc against 3377 spread over 8 queues -- a whole wave's worth of packets for about what one costs.

End-to-end put latency is unchanged at ~5.9 us: that floor is the engine, not the trigger. Bandwidth is a function of bytes per op only -- 52 GB/s at 1 MB, 60 GB/s at 8 MB, identical across scopes and directions.

## Tests and docs

- New `test_sdma_signal` (local/remote signal, `waitSignal`, slot non-contamination) and `test_sdma_warp_issue` (several thread-scope issuers per wavefront). `test_sdma_edge` gains a mixed-slot-size ring-wrap phase that faults the GPU without the straddle fix.
- CI runs both new tests and a benchmark smoke over all three coop scopes, both completion modes, and put/get latency and bandwidth.
- `.claude/skills/cco-sdma-api/` documents which template arguments and scope to pick, the build flags and env vars, and the failure modes.

## Note

The commit chain assumes the `wptr` / doorbell / `committedWptr` stores land in order; upstream `anvil_device.hpp` separates them with `s_waitcnt`. Adding one back costs ~42% of the trigger, and 19200 puts contending on a single queue showed no reordering, so this is deliberate. Called out in the code.

### DIFF
--- a/include/mori/cco/cco.hpp
+++ b/include/mori/cco/cco.hpp
@@ -1158,9 +1158,8 @@ struct ccoSdmaQueueDeviceHandle {
 
   __device__ __forceinline__ void submitPacket(uint64_t base, uint64_t pendingWptr) {
     // In-order commit chain (free when uncontended: one matching load). The
-    // s_waitcnt(0) publishes the packet dwords before the doorbell; the
-    // spin/stores stay RELAXED because visibility comes from it, not from
-    // fences, which would cause a coherence storm on the committedWptr line.
+    // s_waitcnt(0) publishes the packet dwords before the doorbell; the three
+    // stores below rely on landing in order, unlike upstream anvil_device.hpp.
     [[maybe_unused]] int retries = 0;
     while (__hip_atomic_load(committedWptr, __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_AGENT) != base) {
       __builtin_amdgcn_s_sleep(1);
--- a/include/mori/cco/cco.hpp
+++ b/include/mori/cco/cco.hpp
@@ -1194,6 +1194,12 @@ struct ccoSdmaQueueDeviceHandle {
 /* ---------------------------------------------------------------------------------------------- */
 /*                                           Post Tasks                                           */
 /* ---------------------------------------------------------------------------------------------- */
+// Aggregate: post without ringing the doorbell; commit() rings once (like GDA).
+enum ccoSdmaOptFlags : uint32_t {
+  ccoSdmaOptFlagsDefault = 0,
+  ccoSdmaOptFlagsAggregate = (1u << 0),
+};
+
 // Post one COPY on queue `qId`, plus 0/1/2 trailing ATOMIC signal packets.
 // localTarget/remoteTarget are precomputed signalBuf slot addresses, each filled
 // only when its template flag is set.
@@ -1208,7 +1214,8 @@ struct ccoSdmaQueueDeviceHandle {
 //
 // Reserving the slot in one go keeps concurrent same-queue issuers from
 // interleaving. Plain stores, published by submitPacket's s_waitcnt(0).
-template <bool localSignal = false, bool remoteSignal = false, bool Ring = true>
+template <bool localSignal = false, bool remoteSignal = false,
+          uint32_t optFlags = ccoSdmaOptFlagsDefault>
 inline __device__ void ccoSdmaPostCopy(ccoSdmaQueueDeviceHandle** deviceHandles, uint32_t qId,
                                        HSAuint64* localTarget, HSAuint64* remoteTarget,
                                        void* srcPtr, void* dstPtr, size_t size) {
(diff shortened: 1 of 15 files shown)


include/mori/cco/cco.hpp
```diff
@@ -1117,110 +1121,48 @@ __device__ __forceinline__ CCO_SDMA_PKT_ATOMIC ccoCreateAtomicIncPacket(HSAuint6
...
+                                                  ccoSdmaQueueDeviceHandle* shared) {
+    uint64_t base =
+        __hip_atomic_fetch_add(cachedWptr, slotBytes, __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_AGENT);
+    if ((base + slotBytes) - cachedHwReadIndex > CCO_SDMA_QUEUE_SIZE) {
+      [[maybe_unused]] int retries = 0;
+      do {
+        cachedHwReadIndex = __hip_atomic_load(rptr, __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_SYSTEM);
+        if constexpr (CCO_SDMA_BREAK_ON_RETRIES) {
+          if (retries++ == CCO_SDMA_MAX_RETRIES) {
+            __builtin_trap();
```
kawhil-amd, line 1207: Use assert maybe could provide more info for user when in trap?

jhchouuu: it is disable defaulted, and it is indeed possible to provide more info when enable debug mode

include/mori/cco/cco.hpp
```diff
@@ -1229,25 +1235,11 @@ struct ccoSdmaQueueDeviceHandle {
...
-    __builtin_amdgcn_wave_barrier();
     __atomic_signal_fence(__ATOMIC_SEQ_CST);
-
     __hip_atomic_store(wptr, pendingWptr, __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_AGENT);
-
-    __builtin_amdgcn_s_waitcnt(0);
-    __builtin_amdgcn_wave_barrier();
-    __atomic_signal_fence(__ATOMIC_SEQ_CST);
-
     __hip_atomic_store(doorbell, pendingWptr, __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_SYSTEM);
```
kawhil-amd, line 1235: don’t we need to ensure that the write to wptr is visible before the write to the doorbell?

kawhil-amd (approved):

include/mori/cco/cco.hpp
jhchouuu: The s_waitcnt(0) above these stores publishes the packet body before the doorbell. And the order of wptr and doorbell seems not have a affect after I do a stress test of this.

PR #444, opened 2026-07-01
perf(io): add busy-wait completion mode and numpy batch-transfer mars…

TianDi101:
…haling

WaitBusy() spins on the completion flag instead of blocking on the condition variable, trading a CPU core for lower wakeup latency. BatchRead/BatchWrite now accept numpy arrays for offsets/sizes, avoiding per-element Python-list conversion. Benchmark gains --busy-wait, --post-batch-size, --warmup-iters flags and separate launch/transfer latency reporting.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #504, opened 2026-07-28
MORI-IO nixl-style cpp bench script

amirakb89:
# bench(io): nixlbench-matching C++ engine benchmark

Adds `tests/cpp/io/bench_engine.cpp`, a native C++ MORI-IO transfer benchmark
whose measurement loop matches **[nixlbench](https://github.com/ai-dynamo/nixl)**
(NVIDIA NIXL's `xferbench`), so MORI-IO and NIXL RDMA numbers are directly
comparable on the same fabric with no Python-interpreter overhead in the way.

Applies directly to `main`. The only files touched are the new benchmark, its
CMake target, and `docs/MORI-IO-BENCHMARK.md`; no existing source is modified,
and the Python benchmark is untouched.

## What "nixl-matching" means here

- **One whole-loop timer** around the entire iteration loop (nixlbench's
  `total_timer`), not the per-iteration `time.time()` sum the Python bench uses,
  which silently drops the inter-iteration gap.
- **Latency per single transfer**: `total_us / (iters x batch)`, matching nixl's
  `avg_latency = total_duration / (per_thread_iter x batch_size)`.
- **Bandwidth** `msg x batch x iters / 1e9 / (total_us / 1e6)` — GB = 10^9, as in
  nixl.
- **Inline spin-poll completion** in the benchmark thread, mirroring nixl's
  `getXferStatus` scan, so no condition-variable wakeup latency lands in the
  measurement.
- **Strict stop-and-wait**: one request outstanding at a time, equivalent to
  nixl `--pipeline_depth 1`.
- **Warmup excluded**, and memory registered **once** during setup — the buffer
  is sized to the largest request in the whole sweep and reused across every
  size point, so no allocation or registration happens inside the timed loop.

## Also included

- **`--backend rdma|xgmi`**, so the same harness covers intra-node GPU-to-GPU
  transfers, sized by `--num-streams` / `--num-events`.
- **Built-in correctness check** after the sweep, matching the Python benchmark,
  which always validates. Both ranks seed a rank- and offset-dependent pattern
  and exchange one FNV-1a checksum per transferred slot, so the check covers
  every transferred byte while staying O(batch) on the wire. `--skip-validate`
  opts out and is safe to pass on a single rank, since the exchange is a
  collective and an opted-out rank contributes an empty vector rather than
  returning early.
- **Rendezvous over `mori::application::SocketBootstrapNetwork`** rather than a
  second hand-rolled socket setup, so connect/accept retries, timeouts and
  barriers follow the library's policy.
- **Python-parity flags** so runs can be matched arg-for-arg against
  `tests/python/io/benchmark.py`: size and batch sweeps (`--all`, `--all-batch`,
  `--buffer-size`, `--sweep-start`/`--sweep-max`/`--sweep-step`),
  `--enable-batch-transfer` / `--disable-batch-transfer`, chunking, session
  fast path, per-side memory type, `--target-dev-offset`, and the QP tuning
  knobs.
- The buffer and its memory registration are owned by a scope guard and failures
  throw rather than calling `std::exit()`, so the MR is dropped before the pages
  it covers are released on every exit path.
- Unusable arguments are rejected up front instead of being read out of bounds,
  looped on forever or reported as a NaN, and `-h/--help` prints a grouped flag
  list.

## Results

Broadcom Thor2 `bnxt_re` RoCE, 2 nodes, gfx950. VRAM->VRAM WRITE, **1 QP / 1
worker thread**, chunking off, `--batch-contiguous`, session enabled. Bandwidth
in GB/s. **Single run per point — small deltas are directional.**

| Harness | 1 MiB x 1 | 4 KiB x 128 | 512 B x 128 |
|---|---|---|---|
| Python `benchmark.py` | 22.59 | 14.79 | 2.60 |
| C++ `bench_engine` | **24.64** | **16.46** | **2.98** |

The C++ tool reads +9.1% / +11.3% / +14.6% higher, and the gap widening as
messages shrink is what the methodology difference predicts: at 1 MiB the
transfer dominates, while at 512 B each iteration is mostly harness, so the
whole-loop timer, the absent interpreter overhead and the spin-poll instead of a
condition-variable wakeup all take a larger share. These are measurement
differences, not a claim that the transport got faster.

All C++ runs above passed the built-in validation.

## Reproduce

```bash
cmake -B build -DBUILD_IO=ON -DBUILD_TESTS=ON -DBUILD_EXAMPLES=OFF
cmake --build build --target bench_engine -j
export LD_LIBRARY_PATH=$PWD/build/src/io:$PWD/build/src/application:$LD_LIBRARY_PATH
```

Start **rank 1 (target) first**, then rank 0, which drives the transfers and
prints the table. Both ranks take the same workload arguments:

```bash
# TARGET (rank 1) — first
build/tests/cpp/bench_engine --rank 1 --master-ip <RANK0_IP> --self-ip <RANK1_IP> \
    --port 18527 --mem-type gpu --op write --enable-sess \
    --num-qp-per-transfer 1 --num-worker-threads 1 --transfer-batch-size 1 \
    --all --sweep-start 1048576 --sweep-max 1048576 --iters 500 --warmup-iters 50

# INITIATOR (rank 0) — second, prints results
build/tests/cpp/bench_engine --rank 0 --master-ip <RANK0_IP> --self-ip <RANK0_IP> \
    --port 18527 --mem-type gpu --op write --enable-sess \
    --num-qp-per-transfer 1 --num-worker-threads 1 --transfer-batch-size 1 \
    --all --sweep-start 1048576 --sweep-max 1048576 --iters 500 --warmup-iters 50
```

`bench_engine --help` lists every flag, and `docs/MORI-IO-BENCHMARK.md` has the
full walkthrough plus the Python-to-C++ flag mapping.

## Test plan

- [x] Builds and links against `libmori_io` + `libmori_application`
      (`-DBUILD_EXAMPLES=OFF` skips the MPI-only examples target).
- [x] 2-node VRAM WRITE run, batch 1 and batched, sane BW and latency, built-in
      validation passing across read and write, contiguous and strided offsets,
      chunking on and off, with a zero-iteration negative control confirming the
      check fails when nothing moves.
- [ ] Reviewer sanity-check on a second fabric (CX7/mlx5).

### DIFF
--- a/tests/cpp/io/bench_engine.cpp
+++ b/tests/cpp/io/bench_engine.cpp
@@ -41,6 +41,7 @@
 #include <sys/socket.h>
 #include <unistd.h>
 
+#include <algorithm>
 #include <chrono>
 #include <cstdint>
 #include <cstring>
@@ -48,6 +49,7 @@
 #include <optional>
 #include <string>
 #include <thread>
+#include <utility>
 #include <vector>
 
 #include <hip/hip_runtime.h>
@@ -158,10 +160,17 @@ struct Args {
   std::string self_ip;           // this rank's own reachable IP (advertised in EngineDesc)
   uint16_t port = 18515;
   int gpu = 0;
+  int target_dev_offset = 0;     // --target-dev-offset (target GPU = (gpu+offset)%ndev)
   std::string op = "write";      // write|read
-  size_t sweep_start = 1u << 20; // 1 MiB
-  size_t sweep_max = 32u << 20;  // 32 MiB
-  size_t sweep_step = 1u << 20;  // 1 MiB
+
+  // Sweep control (Python parity). --all sweeps message size; --all-batch sweeps
+  // batch size. Neither set => single run at (buffer_size, batch).
+  bool sweep_all = false;        // --all
+  bool sweep_batch = false;      // --all-batch
+  size_t buffer_size = 32768;    // --buffer-size (single message size when not sweeping)
+  size_t sweep_start = 8;        // --sweep-start-size (Python default 8)
+  size_t sweep_max = 1u << 20;   // --sweep-max-size (Python default 2^20)
+  size_t sweep_step = 0;         // --sweep-step (0 = geometric x2; >0 = linear +step)
   int iters = 500;
   int warmup = 50;               // --warmup-iters
   int inflight = 1;              // == nixl pipeline_depth (transfers/requests in flight)
@@ -187,6 +196,10 @@ struct Args {
   bool busy_wait = false;        // --busy-wait (WaitBusy spin); Python default: off
 
   std::string mem_type = "gpu";  // --mem-type gpu|cpu
+  std::string init_mem_type;     // --initiator-mem-type (empty => mem_type)
+  std::string target_mem_type;   // --target-mem-type   (empty => mem_type)
+
+  std::string log_level = "info";  // --log-level trace|debug|info|warning|error|critical
 };
 
 static Args ParseArgs(int argc, char** argv) {
@@ -199,7 +212,11 @@ static Args ParseArgs(int argc, char** argv) {
     else if (k == "--self-ip") a.self_ip = next();
     else if (k == "--port") a.port = static_cast<uint16_t>(std::stoi(next()));
     else if (k == "--gpu") a.gpu = std::stoi(next());
+    else if (k == "--target-dev-offset") a.target_dev_offset = std::stoi(next());
     else if (k == "--op" || k == "--op-type") a.op = next();
+    else if (k == "--all") a.sweep_all = true;
+    else if (k == "--all-batch") a.sweep_batch = true;
+    else if (k == "--buffer-size") a.buffer_size = std::stoull(next());
     else if (k == "--sweep-start" || k == "--sweep-start-size") a.sweep_start = std::stoull(next());
     else if (k == "--sweep-max" || k == "--sweep-max-size") a.sweep_max = std::stoull(next());
     else if (k == "--sweep-step") a.sweep_step = std::stoull(next());
@@ -224,6 +241,9 @@ static Args ParseArgs(int argc, char** argv) {
     else if (k == "--busy-wait") a.busy_wait = true;
     else if (k == "--no-busy-wait") a.busy_wait = false;
     else if (k == "--mem-type") a.mem_type = next();
+    else if (k == "--initiator-mem-type") a.init_mem_type = next();
+    else if (k == "--target-mem-type") a.target_mem_type = next();
+    else if (k == "--log-level") a.log_level = next();
     else { std::cerr << "unknown arg " << k << std::endl; std::exit(1); }
   }
   return a;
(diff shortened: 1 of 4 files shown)


tests/cpp/io/bench_engine.cpp
```diff
@@ -0,0 +1,514 @@
...
+  addr.sin_port = htons(port);
+  if (bind(lfd, reinterpret_cast<sockaddr*>(&addr), sizeof(addr)) != 0) {
+    perror("bind");
+    std::exit(1);
+  }
+  listen(lfd, 1);
+  int fd = accept(lfd, nullptr, nullptr);
+  close(lfd);
+  return fd;
+}
```
maning00: Could we reuse MORI's existing networking abstractions instead of introducing another raw-socket rendezvous implementation here?
`mori::application::TCPContext` and `SocketBootstrapNetwork` already provide connection lifecycle management, send/receive primitives, barriers, timeouts, and cleanup. This benchmark also links `mori_application`, so duplicating these responsibilities with raw file descriptors creates a second implementation with different retry, timeout, and error-handling behavior.
Please consider using the existing bootstrap layer, or extracting a small reusable rendezvous helper from it. That would also avoid manual `close()/std::exit()` paths and make resource ownership explicit.

tests/cpp/io/bench_engine.cpp
```diff
@@ -0,0 +1,514 @@
...
+      if (batched) {
+        // ONE N-descriptor batch request (nixl / Python --enable-batch-transfer).
+        TransferUniqueId id =
+            useSess ? sessPtr->AllocateTransferUniqueId() : engine.AllocateTransferUniqueId();
+        if (useSess) {
+          if (isRead) sessPtr->BatchRead(offsets, offsets, sizes, &base[0], id);
+          else        sessPtr->BatchWrite(offsets, offsets, sizes, &base[0], id);
+        } else {
+          TransferStatusPtrVec sp{&base[0]};
+          TransferUniqueIdVec ids{id};
```
maning00: Both are ordinary `std::vector` types, so this normally introduces two heap allocations and two deallocations per iteration. This harness overhead is included in the reported latency and can be significant for small messages.
Could these vectors be allocated once outside the timed loop and reused by updating sp[0] and ids[0] before each submission? The mode dispatch could likewise be selected before timing, although the allocations are the more significant concern.

amirakb89: Fixed the issue in `5a81bfdc`.

maning00:
Before continuing the review, I’d like to request two structural changes:

1. Preserve the existing RDMA path
The current implementation refactors the regular `RdmaBatchReadWrite` path to share the new build/post logic. At this stage, prepared transfers should remain opt-in and isolated from the existing path, without adding overhead to normal `Read/Write/BatchReadWrite` calls.
The existing `RdmaBatchReadWrite` implementation should therefore remain unchanged, while the prepared API should use a separate build/post path invoked only by `--prepare-once`. Only after prepared transfers have been thoroughly validated and shown not to regress the existing RDMA path should we consider refactoring the regular path to share this implementation.

2. Split this PR into two focused PRs
    * **Baseline C++ benchmark**: benchmark executable, CMake integration, rendezvous, XGMI support, validation, and documentation, without prepared-transfer or RDMA core changes.
    * **Prepared transfers**: prepared-transfer API and implementation, dedicated tests, `--prepare-once` benchmark integration, and benchmark results. This implementation should remain isolated from the existing RDMA path.

Please also rebase after #444 is merged so its commits are not duplicated, and reorganize the commit history around the final logical changes. This will make the performance-sensitive core changes easier to review and verify independently.

PR #509, opened 2026-07-30
rdma: fix dmabuf offset for sub-allocated GPU buffers (LOC_PROT on ionic RoCE)

andyluo7:
## Problem

`RegisterRdmaMemoryRegionAuto` exports a dmabuf fd via `TryExportDmabufFd`
(`hipMemGetHandleForAddressRange`) and registers it with
`ibv_reg_dmabuf_mr(pd, /*offset=*/0, size, /*iova=*/ptr, fd, access)`.

`hipMemGetHandleForAddressRange` exports an fd for the **backing allocation** but
does not report where `ptr` sits inside it. Offset 0 is therefore only correct
when `ptr` is the base of its own allocation.

When `ptr` is a **sub-region of a larger allocation** — e.g. a pooled KV buffer
whose chunks are handed out from one big `hipMalloc`/VMM block, as SGLang does —
the exported dmabuf covers the whole backing block and `ptr` has a **nonzero
offset**. Registering with offset 0 yields an MR whose `lkey`/`rkey` resolve to
the base of the pool instead of `ptr`. On the wire this surfaces as
**`IBV_WC_LOC_PROT_ERR`** on every RDMA WR (or, worse, silent writes to the
wrong address).

This is what blocks MoRI KV transfer on AMD Pensando **ionic** RoCE NICs, which
require dmabuf registration for GPU memory (plain `ibv_reg_mr` returns `EINVAL`).

## Fix

Use `hsa_amd_portable_export_dmabuf`, which returns **both** the fd and the true
byte offset of `ptr` within the dmabuf, and pass that offset to
`ibv_reg_dmabuf_mr`. Fall back to the hip export (offset 0) when HSA export is
unavailable.

`hsa-runtime64` is already a link dependency of `mori_application`
(`src/application/CMakeLists.txt`), so **no build changes are required**.

The VMM symmetric-memory path (`RegisterRdmaMemoryRegionDmabuf`, offset 0) is
intentionally left unchanged: there each chunk's fd is exported for the chunk
base, so offset 0 is already correct.

## Validation

Reproduced and fixed end-to-end on **MI355X (gfx950) + ionic RoCE**, DeepSeek-V4
SGLang PD disaggregation (prefill TP8 / decode TP8, MoRI KV transfer):

- **Before:** every KV transfer fails — decode `KVTransferError`, prefill CQE
  `status=4 (local protection error) vendor_err=3` (`IBV_WC_LOC_PROT_ERR`).
- **After:** KV transfer completes cleanly; benchmark requests succeed (16/16 at
  concurrency 1, latencies matching the reference run). The exported offset is
  confirmed **nonzero** for the 512 MB KV buffers, exactly the case that offset 0
  got wrong.

(The fix was first proven via an `LD_PRELOAD` shim implementing the identical
`hsa_amd_portable_export_dmabuf` → `ibv_reg_dmabuf_mr(pd, offset, …)` logic, then
ported here.)

### DIFF
--- a/src/application/transport/rdma/rdma.cpp
+++ b/src/application/transport/rdma/rdma.cpp
@@ -33,6 +33,10 @@
 #include <unordered_set>
 
 #include "infiniband/verbs.h"
+
+#include <hsa/hsa.h>
+#include <hsa/hsa_ext_amd.h>
+
 #include "mori/application/transport/rdma/providers/bnxt/bnxt.hpp"
 #include "mori/application/transport/rdma/providers/dv_loader.hpp"
 #include "mori/application/transport/rdma/providers/ibverbs/ibverbs.hpp"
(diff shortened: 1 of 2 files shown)


maning00:
Thanks @andyluo7, could you fix the pre-commit format error?

maning00:
and if convenient, could you add a regression test to `tests/cpp/io/test_engine.cpp` that registers and transfers an interior GPU range (allocation_base + nonzero_offset) while forcing the dma-buf path with `MORI_ENABLE_DMABUF_REG=1`?

andyluo7:
Thanks @maning00 — both done:

1. **Pre-commit**: ran clang-format (v20.1.8, `-style=file`) on `rdma.cpp`; the reformat is in `aa284646`.

2. **Regression test** (`5fde29f2`): added `CaseRdmaDmabufInteriorRangeTransfer` to `tests/cpp/io/test_engine.cpp` (registered as `rdma_dmabuf_interior_range_transfer`). It forces the dma-buf path with `MORI_ENABLE_DMABUF_REG=1`, then:
   - `hipMalloc`s a 64 MiB backing allocation per side and registers only an **interior** sub-range (`base + 32 MiB`, 1 MiB) via `RegisterMemory`;
   - does a loopback RDMA `Write` into that sub-range;
   - verifies the payload (`0xCD`) lands in the **sub-range**, and that the **allocation base sentinel** (`0x11`) is untouched.

   With the pre-fix offset-0 behavior the write resolves to the allocation base, so both checks fire (missing payload in the sub-range and/or a clobbered base). It guards with `GetGpuCount() < 1 -> TestSkip` and falls back cleanly on non-dma-buf NICs (registration still succeeds via `ibv_reg_mr`, and the offset is 0 there anyway).

Let me know if you'd prefer the test to also assert the transfer status explicitly reports LOC_PROT on the buggy path, or any other tweak.

andyluo7:
Heads-up on CI: the one red check, **`mori intranode test (MI325X_BNXT)`**, is a spurious timeout unrelated to this change (it's already re-running).

- It failed with `exit 124` (timeout), and the log shows it hung in the **Python** suite (`tests/python/io/test_engine.py`, right after `... PASSED [71%]`) — this PR is C++-only and doesn't touch that path.
- The same job passed on the pre-fix commit, and **`mori intranode test (MI355X_AINIC)` passed** the identical suite on this run.
- Everything else is green: `pre-commit`, both `mori build` jobs (so the new `test_engine.cpp` case compiles), and **both `internode test` jobs (MI325X_BNXT + MI355X_AINIC) pass**.

Looks like a re-run should clear it — let me know if you'd like anything else.

PR #510, opened 2026-07-30
roctx ranges

AakarshAMD:
## Summary

Adds optional ROCTX instrumentation to the MORI-IO RDMA host send path for profiling with `rocprofv3`/rocprofiler-sdk. This change is **purely additive** — no existing logic is modified, removed, or renamed. When disabled (the default), it has zero runtime cost and introduces no new dependencies.

## What's added

- `src/io/roctx_mori.hpp` (new) — all ROCTX helper infrastructure: env-var gating, lazy `dlopen`/`dlsym` of `librocprofiler-sdk-roctx.so`, and thread-safe range/keyed-event handling.
- `src/io/engine.cpp` — wraps `IOEngine[Session]::BatchWrite` with sync host-post ranges.
- `src/io/rdma/common.cpp` — wraps `RdmaBatchReadWrite`'s host WR-posting loop; starts async post→CQE transfer ranges for signaled WRs.
- `src/io/rdma/backend_impl.cpp` — stops async transfer ranges when the CQ poller reaps the corresponding completion.

## Env vars (both OFF by default)

| Env Var | Default | Purpose |
|---|---|---|
| `MORI_ROCTX` | off | Synchronous push/pop ranges around the host `ibv_post_send` path. Measures **host post cost only** (building WRs + ringing the NIC doorbell). |
| `MORI_ROCTX_TRANSFER` | off | Asynchronous post→CQE ranges measuring the **real wire/transfer duration** — started at WR post time, stopped when the completion is reaped on the CQ thread. |

## Behavior when disabled

- No `dlopen` of the ROCTX library occurs.
- All marker call sites are no-ops, guarded by the resolved gate flags checked once at process start.
- No rebuild required to toggle — purely a runtime switch via env vars.
- No risk of missing-`.so` errors when disabled, since the library is never loaded in that path.

## Behavior when enabled

- `MORI_ROCTX=1` and/or `MORI_ROCTX_TRANSFER=1` trigger a lazy `dlopen("librocprofiler-sdk-roctx.so")` (falls back to `.so.1`) and resolve ROCTX function pointers via `dlsym`.
- Markers are visible in `rocprofv3 --marker-trace` (or any ROCTX-consuming tool) alongside GPU/kernel traces.
- Async transfer ranges are keyed by `(ledger, recordId)` to correctly pair post and completion events across threads (posting thread vs. CQ poller thread).

## Testing

- Verified marker emission and pairing with `rocprofv3 --marker-trace` under both env vars individually and together.

### DIFF
--- a/docs/MORI-IO-GUIDE.md
+++ b/docs/MORI-IO-GUIDE.md
@@ -10,6 +10,7 @@ MORI-IO is AMD's point-to-point communication library that leverages GDR (GPU Di
 - [Python API Quick Reference](#python-api-quick-reference)
 - [Example: Basic Read/Write](#example-basic-readwrite)
 - [Environment & Configuration](#environment--configuration)
+- [Profiling MORI-IO with ROCTX Markers](#profiling-mori-io-with-roctx-markers)
 - [Source Files](#source-files)
 
 ## Design & Concepts
@@ -181,6 +182,38 @@ See `examples/io/example.py` for more complete examples including batch transfer
 UMBP (the upper-layer cache pool) exposes a separate set of runtime-tunable
 env vars for distributed master / pool client / SPDK proxy timing. Those are
 out of scope for MORI-IO; see [`src/umbp/doc/runtime-env-vars.md`](../src/umbp/doc/runtime-env-vars.md).
+
+## Profiling MORI-IO with ROCTX Markers
+
+MORI-IO provides optional, runtime-gated ROCTX timeline markers for use with
+`rocprofv3 --marker-trace`. The following variables are disabled by default and
+operate independently:
+
+- `MORI_ROCTX` enables same-thread host submission ranges around MORI-IO
+  batch-write dispatch and RDMA posting paths. They measure host-side submission
+  and post work only, not completion latency.
+- `MORI_ROCTX_TRANSFER` enables process-wide ranges that begin immediately
+  before posting a signaled RDMA work request and end on the CQ polling thread
+  after its CQE is observed. They measure post-to-completion latency.
+
+Marker fields `bytes=` and `id=` identify application payload bytes (not
+physical RoCE wire bytes) and MORI's logical transfer ID, respectively. Key host
+marker families are `mori.io.*` and `mori.rdma.batch_post.*`; transfer markers
+use `mori.rdma.io_transfer[.read]`.
+
+Although generic to MORI-IO, these markers can help observe KV-cache
+post-to-completion latency in prefill/decode-disaggregated vLLM or SGLang
+deployments configured to use MORI-IO.
+
+```bash
+export MORI_ROCTX=1
+export MORI_ROCTX_TRANSFER=1
+rocprofv3 --marker-trace -- <your MORI command>
+```
+
+Either variable can be enabled alone. Enabled values include `1`, `on`, and
+`true`; disable a variable by unsetting it or setting it to `0`, `off`, or
+`false`.
 
 ## Source Files
 
--- a/src/io/engine.cpp
+++ b/src/io/engine.cpp
@@ -41,6 +41,7 @@
 #include "src/io/fabric/backend_impl.hpp"
 #include "src/io/rdma/backend_impl.hpp"
 #include "src/io/xgmi/backend_impl.hpp"
+#include "src/io/roctx_mori.hpp"  // ADDITIVE: MORI_ROCTX-gated host-send roctx markers
 
 namespace mori {
 namespace io {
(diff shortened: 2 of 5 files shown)


maning00:
Thanks @AakarshAMD, ROCTX instrumentation is useful for correlating host-side RDMA activity with GPU execution in a rocprofv3 timeline. Before merging, I think the following points should be addressed:

* The disabled path should have genuinely negligible overhead. Currently, `std::accumulate`, `epBytesSinceSignal` allocation, and per-WR byte accounting are performed even when both ROCTX options are disabled. Please gate or eliminate this work from the default hot path.
* A post-to-CQE range is not the actual wire duration. It also includes NIC/SQ queueing, CQ polling latency, and software scheduling/processing. Please describe it as post-to-completion latency rather than “real wire/transfer duration.”
* MORI-IO is a generic transfer API and does not necessarily carry KV cache data. Please rename `mori.rdma.kv_transfer` and related comments to a generic name such as `mori.rdma.io_transfer`.

With these changes, the optional instrumentation would be useful without affecting the normal MORI-IO data path

maning00:
and, pls fix pre-commit format error

AakarshAMD:
> and, pls fix pre-commit format error

hey! I fixed all the issues you had mentioned

maning00:
Thanks for the update. ROCTX timeline support is useful and complements the snapshot telemetry in #345. I think the following points still need to be addressed:
1. Please integrate the async tracking with #345. It already adds per-EP byte accumulation and stores completion-related data in `SubmissionLedger`. We should use one submission/completion lifecycle with optional ROCTX state instead of maintaining a second process-wide `(ledger, recordId)` map and byte-accounting path.
Please rebase your branch onto `feat/io-telemetry` and change this PR’s base branch from `main` to `feat/io-telemetry`. This PR can then be merged into #345, with the combined telemetry changes subsequently merged into `main` through #345.

2. Please provide an A/B benchmark with both ROCTX variables unset, comparing #345 before and after these changes. The disabled path avoids allocation and `dlopen`, but it still adds gate checks to the submission and CQE paths, so the “zero runtime cost” claim needs measurement.
3. Please replace `GateOn()` with `mori::env::IsEnvVarEnabled()`. The current implementation treats `MORI_ROCTX=off` as enabled.

AakarshAMD:
Hello, 
I have made the changes you asked for. 
I don't think we need to merge this with pr #345 since it has separate functionality. Plus the (submission_ledger, record_id), is not maintained with the environment variables are off. 

I have also made the changes to switch from GateOn to IsEnvVarEnabled

## Tests
I ran the benchmark tests 
| Operation | Variant | Median Avg BW (GB/s) | Median Avg Lat (us) |
|---|---:|---:|---:|
| Write | BASE | 19.700 | 53.235 |
| Write | OFF | 19.695 | 53.250 |
| Write | ON | 19.450 | 53.915 |
| Read | BASE | 19.635 | 53.415 |
| Read | OFF | 19.660 | 53.325 |
| Read | ON | 19.450 | 53.910 |

20 paired runs per operation/comparison; 95% bootstrap CI.

| Comparison | Operation | Metric | Mean delta | 95% CI |
|---|---|---|---:|---:|
| BASE → OFF | Write | Avg BW | +0.0305% | [-0.2498%, +0.3201%] |
| BASE → OFF | Write | Avg Lat | -0.0322% | [-0.3235%, +0.2511%] |
| BASE → OFF | Read | Avg BW | +0.4675% | [-0.0602%, +1.2448%] |
| BASE → OFF | Read | Avg Lat | -0.4506% | [-1.1824%, +0.0566%] |
| OFF → ON | Write | Avg BW | -1.2705% | [-1.5551%, -0.9940%] |
| OFF → ON | Write | Avg Lat | +1.2938% | [+1.0058%, +1.5950%] |
| OFF → ON | Read | Avg BW | -1.0467% | [-1.2715%, -0.8188%] |
| OFF → ON | Read | Avg Lat | +1.0586% | [+0.8271%, +1.2823%] |

## Conclusion
- With variables unset, no statistically distinguishable average BW/latency impact was detected; this result does not establish absence of cost.



@raviguptaamd @shikamd123 @amirakb89

shikamd123:
The PR looks good to me and we should get this merged.

maning00:
I re-reviewed the current diff against `main` in detail. The feature can land before #345 and does not need to depend on it. The disabled-path A/B already provided shows no statistically distinguishable regression, so this looks good to merge.

One non-blocking suggestion: the enabled transfer-profiling path currently uses a process-wide side-table map/mutex to associate range IDs with ledger records. Moving that state into `SubmissionLedger` could reduce contention and keep completion state in one place. If pursued, preserve the pre-post start point, return the range ID during record release, and call `roctxRangeStop` outside the ledger lock. Since the current side table is gated and no disabled-path regression has been shown, this refactor is not required for this merge.

Please keep the feature fully gated. If a later change modifies the disabled hot path or the ledger record layout, repeat the same A/B methodology to confirm no regression relative to `main`.

maning00 (approved):

PR #511, opened 2026-07-31
perf(EP): tune EPv1 MI355X EP8 for DeepSeek-V4-Pro / Kimi-K3, key configs by topk

Duyi-Wang:
## Summary

Tunes **EPv1** IntraNode EP8 on MI355X (gfx950) at hidden 7168 for two model shapes — **DeepSeek-V4-Pro** (top-6, 384 routed experts) and **Kimi-K3** (top-16, 896 routed experts) — across bf16 / fp8 / fp4 dispatch, and adds `topk` to the tuning config key so the two shapes can coexist.

- **Schema**: rules gain an optional `topk` field that participates in matching. Fully backward compatible — existing rules have no `topk` and keep working as wildcards.
- **Configs**: 42 measured combos (2 shapes × 3 dispatch dtypes × 7 token counts, 64–4096) → 42 new dispatch rules + 28 new combine rules.

## Motivation

Rules are keyed by `(dtype, num_tokens, hidden_dim)`. Both target models are hidden 7168, so tuning the second shape **silently overwrote the first** — there was no way to hold configs for both.

That would be harmless if geometry tracked token count alone, but it doesn't. Top-16 routes ~2.7× the tokens per rank of top-6, and the measured optima differ accordingly (bf16, GB/s at the chosen geometry):

| tok/rank | top-6 dispatch | top-16 dispatch | top-6 combine | top-16 combine |
|---:|---|---|---|---|
| 256 | blk 192, **warp 4** | blk 128, **warp 16** | blk 128, warp 8 | blk 128, warp 16 |
| 1024 | blk 256, **warp 6** | blk 256, **warp 16** | **blk 64**, warp 8 | **blk 128**, warp 16 |
| 4096 | blk 208, **warp 12** | blk 216, **warp 16** | **blk 64**, warp 16 | **blk 128**, warp 16 |

Top-16 wants the maximum warp count everywhere on dispatch, and twice the combine block at mid/large token counts, where top-6 is still latency-bound and prefers a small block.

### `topk` in the rule key

- `topk` is an **optional** rule field. Rules written before this change have none and act as wildcards, so existing configs resolve exactly as they did.
- `lookup()` takes `topk=None` and tries progressively less specific passes: `(hidden ✓, topk ✓) → (hidden ✓, topk *) → (hidden *, topk ✓) → (hidden *, topk *)`. **`hidden_dim` is relaxed last on purpose**: it sets per-token transport volume, so a rule measured at the right hidden and a different top-k is a better guess than the reverse.
- `save_tuning_result()` includes `topk` in the merge key, so tuning a shape at a new top-k adds rules instead of replacing existing ones.
- `EpDispatchCombineOp` passes `config.num_experts_per_token`; the tuning bench records it in the rules it writes.

### Measured configs

`gfx950_mi355x_IntraNode_ep8_{dispatch,combine}.json`, full-scope block×warp sweep (~360 geometries per combo), zero-copy, 0 failures. fp4 dispatch pairs with the fp8 direct-cast combine wire, matching `tools/run_all_intranode_tuning.sh`. The 14 dispatch / 28 combine rules that predate the `topk` key are untouched and keep serving as wildcard fallbacks.

## Not covered

- **tok 8192** — a full-scope sweep there exceeds a 40 min/combo budget; `lookup()` clamps above-range counts to the largest rule.
- **`zero_copy=0`** — combine rules are keyed by `zero_copy`; only the default zero-copy path was swept.
- **EPv2** — measured EPv2 schedules were dropped from this PR: the cco-LSA kernels have since changed, which invalidates those numbers. They will be re-measured against the new kernels and land separately.
- Other devices, EP sizes, and hidden dims are untouched.

## Test plan

- `MORI_EP_LAUNCH_CONFIG_MODE=AUTO` resolves the new rules on hardware for all 6 shape×dtype combinations, and an EP8 bench run at the Kimi-K3 shape completes cleanly.
- All 42 combos completed with 0 failures.
- Backward compatibility of the schema change was checked against configs with no `topk` field: exact top-k wins where present, legacy rules still resolve for untuned top-k, `hidden_dim` outranks `topk`, and the num_tokens ceiling/clamp still applies within a top-k group.
- Full GPU CI (intranode + internode, MI355X and MI325X) passed on the pre-trim revision of this branch, which contained these same EPv1 changes.


jhchouuu (approved):
LGTM

PR #515, opened 2026-07-31
MORI-IO XGMI KV-Transfer Session Failure

amd-wsung102:
## Motivation
A bug was reported by Dell under the following environment.
- **Hardware**: Single node, 8x AMD MI300X GPUs, no RDMA NIC present (XGMI-interconnected, `rocm-smi --showtopo` confirms full XGMI mesh between all 8 GPUs).
- **Image**: `vllm/vllm-openai-rocm:nightly-dcfebf93f4eccf30f71872283331eee757915daf`
- **Model**: `Qwen/Qwen3-30B-A3B` (MoE, expert-parallel enabled)
- **Topology**: Disaggregated prefill/decode serving via vLLM's `MoRIIOConnector` (KV cache transfer library `mori`, version `amd_mori 1.0.0`, bundled in the image).
  - Proxy: routes client requests, handles push-based registration/discovery.
  - Prefill: GPUs 0-3, TP=4, `kv_role=kv_producer`.
  - Decode: GPUs 4-7, TP=4, `kv_role=kv_consumer`, `read_mode=true` (decode pulls KV blocks from prefill).
  - Transport backend: `xgmi` (no RDMA hardware available, so RDMA is not usable; XGMI fallback is intentional).

## Error Log
```
(Worker_TP2_EP2 pid=969) INFO [moriio_connector.py:1291] MoRIIO handshake: get metadata took: 0.0016591940002399497
(Worker_TP2_EP2 pid=969) INFO [moriio_connector.py:1357] MoRIIO handshake done for request chatcmpl-...
(Worker_TP2_EP2 pid=969) ERROR [multiproc_executor.py:1007] WorkerProc hit an exception.
(Worker_TP2_EP2 pid=969) ERROR [multiproc_executor.py:1007] Traceback (most recent call last):
  File ".../vllm/v1/executor/multiproc_executor.py", line 999, in worker_busy_loop
    output = func(*args, **kwargs)
  File ".../vllm/v1/worker/worker_base.py", line 351, in execute_model
    return self.worker.execute_model(scheduler_output)
  File ".../vllm/v1/worker/gpu_worker.py", line 1147, in execute_model
    output = self.model_runner.execute_model(...)
  File ".../vllm/v1/worker/gpu_model_runner.py", line 4389, in execute_model
    self.maybe_get_kv_connector_output(...)
  File ".../vllm/v1/worker/kv_connector_model_runner_mixin.py", line 95, in _get_kv_connector_output
    kv_connector.start_load_kv(get_forward_context())
  File ".../vllm/distributed/kv_transfer/kv_connector/v1/moriio/moriio_connector.py", line 296, in start_load_kv
    self.connector_worker.start_load_kv(self._connector_metadata)
  File ".../vllm/distributed/kv_transfer/kv_connector/v1/moriio/moriio_connector.py", line 1799, in start_load_kv
    self._read_blocks_for_req(*self._ready_requests.get_nowait())
  File ".../vllm/distributed/kv_transfer/kv_connector/v1/moriio/moriio_connector.py", line 1818, in _read_blocks_for_req
    self._read_blocks(...)
  File ".../vllm/distributed/kv_transfer/kv_connector/v1/moriio/moriio_connector.py", line 2012, in _read_blocks
    transfer_status = self.moriio_wrapper.read_remote_data(...)
  File ".../vllm/distributed/kv_transfer/kv_connector/v1/moriio/moriio_engine.py", line 605, in read_remote_data
    transfer_status = session.batch_read(...)
AttributeError: 'NoneType' object has no attribute 'batch_read'

(EngineCore pid=673) ERROR [core.py:1332] RuntimeError: Worker failed with error
  ''NoneType' object has no attribute 'batch_read'', please check the stack trace above for the root cause
(APIServer pid=1) ERROR [async_llm.py:704] vllm.v1.engine.exceptions.EngineDeadError: EngineCore encountered an issue.
(APIServer pid=1) INFO: 127.0.0.1:xxxxx - "POST /v1/chat/completions HTTP/1.1" 500 Internal Server Error
```

## Problem

`IOEngine::RegisterRemoteEngine` and `IOEngine::RegisterMemory` only forward to the backends that exist at call time, and nothing back-fills a backend created later:

```c
void IOEngine::RegisterRemoteEngine(const EngineDesc& remote) {
  for (auto& it : backends) {          // whatever happens to exist right now
    it.second->RegisterRemoteEngine(remote);
  }
}
```

A backend installed after those calls is permanently blind to them: `XgmiBackend::CanHandle` rejects every pair whose `remote.engineKey` it never learned about, `SelectBackend` returns `nullptr`, and `CreateSession` yields `std::nullopt` forever.

This needs no misuse by the caller, because MORI creates XGMI backends late on its own — `EnsureXgmiBackendCreatedIfSupported` runs *after* RDMA init, and the XGMI-only fallback constructs one when no RDMA NIC is active. Registering peers before `CreateBackend`, the natural order when discovery precedes transport setup, yields an XGMI backend that can never route.

Neither `nullopt` branch of `CreateSession` logged anything, so the failure reached Python as a bare `None` and surfaced later on an unrelated line as `AttributeError: 'NoneType' object has no attribute 'batch_read'`. It is the one silent entry point; `Read`/`Write`/`BatchRead`/`BatchWrite` all go through `SELECT_BACKEND_AND_RETURN_IF_NONE`, which sets `ERR_BAD_STATE` and logs.

The XGMI hidden-device path itself is sound — I ran the split-visibility topology (disjoint `HIP_VISIBLE_DEVICES`, 1/8/48/128 KV-layer registrations, same- and cross-container) against `main` and it works. This is an ordering defect, not a capability gap.

## Fix

Install every backend through a new `IOEngine::InsertBackend`, which replays what was registered before it existed:

```c
Backend* IOEngine::InsertBackend(BackendType type, std::unique_ptr<Backend> backend) {
  auto [it, inserted] = backends.insert({type, std::move(backend)});
  Backend* raw = it->second.get();
  InvalidateRouteCache();
  if (!inserted) return raw;

  for (const auto& [key, remoteDesc] : remoteEngines) raw->RegisterRemoteEngine(remoteDesc);
  for (auto& [id, memDesc] : memPool) raw->RegisterMemory(memDesc);
  return raw;
}
```

`IOEngine` now tracks `remoteEngines` for this (memory was already in `memPool`). All four construction sites — RDMA, XGMI, FABRIC, XGMI-only fallback — route through it, so ordering stops mattering. Replay is idempotent and a no-op when nothing has been registered yet.

The rest makes the failures that remain explain themselves:

- **`Backend::ExplainCannotHandle(local, remote)`** — new virtual returning why a pair was declined, empty meaning accepted. Defaults to deriving from `CanHandle`, so other backends need no change.
- **`XgmiBackend`** implements it, and `CanHandle` becomes `ExplainCannotHandle(...).empty()` to keep one decision path. All seven rejection conditions now name themselves.
- **`IOEngine::CreateSession`** logs at `ERROR` on both `nullopt` branches via the new `ExplainSessionUnavailable`, which is also exported to Python.
- **Python `create_session`** raises `SessionUnavailableError` instead of returning `None`.
- **`RegisterMemory`** warns when no backend exists yet. That case is unrecoverable rather than merely late: backends supply the credentials a peer needs (IPC handle for XGMI, MR keys for RDMA), and a descriptor already returned to the caller cannot be back-filled.

## Test

`tests/python/io/test_xgmi_split_visibility.py` (new) runs two processes with disjoint `HIP_VISIBLE_DEVICES`, mirroring a disaggregated prefill/decode split, and adds three cases:

- `test_xgmi_read_across_disjoint_visible_devices` — an XGMI read completes across the split, guarding the hidden-device path.
- `test_xgmi_session_when_remote_engine_registered_before_backend` — the regression itself: a remote engine registered before the backend exists.
- `test_create_session_names_reason_when_peer_memory_has_no_ipc_handle` — the unrecoverable case raises and the message names its cause.

Verified on 8x MI300X. With the replay loop disabled the regression test fails; with the fix all three pass. The full `tests/python/io/` suite reports 71 failed, 91 passed, 3 skipped — the same 71 as the pre-change baseline on this node, which has no usable RDMA NIC (`ibv_create_qp` errno=22).

`test_engine.py::test_no_backend` asserted the old `None` contract and now asserts the raise; `benchmark.py` loses a dead `None` check.

## Submission Checklist

### DIFF
--- a/include/mori/io/backend.hpp
+++ b/include/mori/io/backend.hpp
@@ -22,6 +22,7 @@
 #pragma once
 
 #include <cstddef>
+#include <string>
 
 #include "mori/io/common.hpp"
 #include "mori/io/enum.hpp"
@@ -194,6 +195,11 @@ class Backend {
                                         TransferStatus* status) = 0;
 
   virtual bool CanHandle(const MemoryDesc& local, const MemoryDesc& remote) const { return true; }
+
+  // Why CanHandle() rejected this pair; empty when supported. Callers have no other error channel.
+  virtual std::string ExplainCannotHandle(const MemoryDesc& local, const MemoryDesc& remote) const {
+    return CanHandle(local, remote) ? std::string{} : std::string{"backend declined this pair"};
+  }
 };
 
 }  // namespace io
--- a/include/mori/io/engine.hpp
+++ b/include/mori/io/engine.hpp
@@ -108,6 +108,8 @@ class IOEngine {
   StatusCode WaitAll(const std::vector<TransferStatus*>& statuses, int timeoutMs = -1);
 
   std::optional<IOEngineSession> CreateSession(const MemoryDesc& local, const MemoryDesc& remote);
+  // Why CreateSession() could not build a session for this pair, empty when it can.
+  std::string ExplainSessionUnavailable(const MemoryDesc& local, const MemoryDesc& remote) const;
   void LoadScatterGatherModule(const std::string& hsacoPath);
   void LoadFabricCopyModule(const std::string& hsacoPath);
 
@@ -150,6 +152,8 @@ class IOEngine {
   };
 
   Backend* SelectBackend(const MemoryDesc& local, const MemoryDesc& remote);
+  // Install a backend and replay the registrations it missed by coming up late.
+  Backend* InsertBackend(BackendType type, std::unique_ptr<Backend> backend);
   bool SupportsXgmiBackendByP2P() const;
   void EnsureXgmiBackendCreatedIfSupported();
   void InvalidateRouteCache();
@@ -164,6 +168,8 @@ class IOEngine {
   std::atomic<uint32_t> nextTransferUid{0};
   std::atomic<uint32_t> nextMemUid{0};
   std::unordered_map<MemoryUniqueId, MemoryDesc> memPool;
+  // Remote engines registered so far, replayed into backends created later.
+  std::unordered_map<EngineKey, EngineDesc> remoteEngines;
   std::unordered_map<BackendType, std::unique_ptr<Backend>> backends;
   mutable std::shared_mutex routeCacheMu;
   std::unordered_map<RouteCacheKey, BackendType, RouteCacheKeyHash> routeCache;
--- a/python/mori/io/engine.py
+++ b/python/mori/io/engine.py
@@ -34,6 +34,10 @@ TORCH_DEVICE_TYPE_MAP = {
 }
 
 
+class SessionUnavailableError(RuntimeError):
+    """No backend could build a transfer session for a local/remote memory pair."""
+
+
 class IOEngineSession:
     def __init__(self, mori_sess):
         self._sess = mori_sess
@@ -223,7 +227,10 @@ class IOEngine:
     def create_session(self, local_mem, remote_mem):
         mori_sess = self._engine.CreateSession(local_mem, remote_mem)
         if mori_sess is None:
-            return None
+            # Returning None here pushed the failure to the first transfer call, surfacing as an unattributable AttributeError on NoneType.
+            raise SessionUnavailableError(
+                self._engine.ExplainSessionUnavailable(local_mem, remote_mem)
+            )
         return IOEngineSession(mori_sess)
 
     def pop_inbound_transfer_status(self, remote_key, transfer_uid):
--- a/src/io/engine.cpp
+++ b/src/io/engine.cpp
@@ -30,6 +30,7 @@
 #include <chrono>
 #include <cstdlib>
 #include <limits>
+#include <sstream>
 #include <stdexcept>
 #include <string>
 #include <utility>
@@ -213,9 +214,7 @@ void IOEngine::CreateBackend(BackendType type, const BackendConfig& beConfig) {
                    internal::kXgmiOnlyFallbackPlaceholderPort);
 
       XgmiBackendConfig xgmiConfig{};
-      auto backend = std::make_unique<XgmiBackend>(desc.key, config, xgmiConfig);
-      backends.insert({BackendType::XGMI, std::move(backend)});
-      InvalidateRouteCache();
+      InsertBackend(BackendType::XGMI, std::make_unique<XgmiBackend>(desc.key, config, xgmiConfig));
       return;
     }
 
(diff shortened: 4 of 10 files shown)


maning00:
Thanks @amd-wsung102 for improving the session-creation diagnostics. Logging the rejection reason and raising a meaningful Python exception instead of returning `None` are valuable changes.

 However, I’m not convinced that the backend-registration replay is the root-cause fix for the reported vLLM failure.

 I checked the exact vLLM commit used by the reported image (`dcfebf93f4eccf30f71872283331eee757915daf`). Its ordering is:

 1. construct `IOEngine`;
 2. call `set_backend_type()`, which synchronously creates the RDMA/XGMI backend;
 3. register KV memory later in `register_kv_caches()`;
 4. register the remote engine later during the handshake;
 5. create transfer sessions.

 Relevant code:

 - [engine/backend initialization](https://github.com/vllm-project/vllm/blob/dcfebf93f4eccf30f71872283331eee757915daf/vllm/distributed/kv_transfer/kv_connector/v1/moriio/moriio_connector.py#L901-L948)
 - [remote-engine registration](https://github.com/vllm-project/vllm/blob/dcfebf93f4eccf30f71872283331eee757915daf/vllm/distributed/kv_transfer/kv_connector/v1/moriio/moriio_connector.py#L1257-L1299)
 - [memory registration](https://github.com/vllm-project/vllm/blob/dcfebf93f4eccf30f71872283331eee757915daf/vllm/distributed/kv_transfer/kv_connector/v1/moriio/moriio_connector.py#L1415-L1483)

 Therefore, the new MORI regression test proves that registration-before-backend is unsupported by the existing API, but it does not establish that the reported vLLM workload actually exercised that ordering.

 To establish causality, I think we need an A/B result using the same vLLM workload and environment:

 - logs showing that `RegisterRemoteEngine` ran before XGMI backend creation;
 - replay disabled: failure is reproducible and caused by the XGMI backend not knowing the remote engine;
 - replay enabled, with no other change: the failure disappears.

 Without that evidence, I suggest keeping the diagnostic improvements but removing or splitting out `InsertBackend`, `remoteEngines`, and registration replay.

amirakb89:
The root-cause of the failure is PID namespaces, not registration ordering. Passing  `--pid host` on both containers fixes it. 
It also works with a single container.

amd-wsung102:
As Amir mentioned, the root cause is not having unique PIDs, which can be fixed by passing `--pid host` to containers. Therefore, this PR is being closed.

PR #517, opened 2026-08-03
feat(EPv2): enable fp4 dispatch feature in EPv2

kawhil-amd:
## Summary

* **fp4 dispatch + bf16 combine (`fp4_disp_bf16_comb`)** — unblock the EPv2 fp4
     path and give it its own tuning key, plus measured MI355X schedules.

### DIFF
--- a/python/mori/ops/dispatch_combine_v2/dispatch_combine_op.py
+++ b/python/mori/ops/dispatch_combine_v2/dispatch_combine_op.py
@@ -190,9 +190,14 @@ class EpDispatchCombineConfig:
                     "combine_data_type (asymmetric dtype) requires combine_mode=gather, "
                     "quant_type=none, enable_std_moe=False"
                 )
-            if torch.float4_e2m1fn_x2 in (self.dispatch_dtype, self.combine_dtype):
+            # fp4 DISPATCH + bf16 COMBINE (the SGLang/aiter fp4-asym path) is
+            # supported: dispatch is a plain byte-mover (hidden/2 B) with e8m0
+            # scales forwarded, combine stays bf16. fp4 on the combine side is
+            # not implemented for the asymmetric path.
+            if self.combine_dtype == torch.float4_e2m1fn_x2:
                 raise ValueError(
-                    "asymmetric dispatch/combine dtype does not support fp4"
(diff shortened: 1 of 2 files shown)


jhchouuu (approved):
LGTM

kawhil-amd:
Some works refer to https://github.com/kkHuang-amd/mori/commits/feat/epv2-sglang-stability/, thanks to @kkHuang-amd for the contribution.

python/mori/ops/dispatch_combine_v2/dispatch_combine_op.py
```diff
@@ -402,12 +412,14 @@ def __init__(
...
         *,
         reverse_src_view=None,
+        recv_cap=None,
     ):
         self.disp_dest_tok_id_map = disp_dest_tok_id_map
         self.inter_node_disp_dest_tok_id_map = inter_node_disp_dest_tok_id_map
         self.inter_node_disp_send_map = inter_node_disp_send_map
         self.total_recv_token_num = total_recv_token_num
         self.cur_rank_num_token = cur_rank_num_token
+        self.recv_cap = recv_cap
```
kawhil-amd: maybe change name to expect_recv_token will be better

kawhil-amd (commented):
## Review Comments

Combine side is fine (both use 2-tuples).

### API simplification suggestions

1. **`clone_routing` can be dropped** — it saves one small tensor alloc (`n_tok * topk * 4` bytes) but exposes a fragile "must combine before next dispatch" constraint to the caller. `return_routing=True` should always clone; simpler and safer.

2. **Four dispatch variant dicts → fewer** — `_dispatch_replay_variants` and `_dynamic_dispatch_replay_variants` serve the same purpose (lazy-built replay kernels), just keyed by different recv_cap values. Merging them into one dict with `recv_cap` in the key removes complexity without losing anything. The pre-built `_dispatch_variants` has a reason to stay separate (O(1) init-time lookup), but the replay side has no pre-building, so two dicts add no value.

python/mori/ops/dispatch_combine_v2/dispatch_combine_op.py
kawhil-amd: outdate

PR #519, opened 2026-08-04
Fix flaky data corruption in CCO SDMA tests

QizhouZhang97:
hipMemset is async on the host, but ccoBarrierAll is a host-only socket barrier. The pending fill can therefore land after a peer's SDMA put and overwrite it (chunk 0 word 0: got ffffffff, 16/200).

Fix: hipDeviceSynchronize() after each hipMemset, before the barrier


kawhil-amd (approved):
LGTM

PR #528, opened 2026-08-05
bugfix: guard ModuleLogger's shared state with a mutex

QizhouZhang97:
ModuleLogger's loggers_/envOverrides_ maps had no synchronization. When multiple threads first call GetLogger() for the same module concurrently (e.g. CCO ranks entering ccoCommCreate together), the unconditional loggers_[moduleName] = logger write races on the same unordered_map, corrupting it — reproduced via disassembly as a SIGSEGV either inside operator[] or when dereferencing a torn shared_ptr. No MORI_*_LOG_LEVEL env var is required to trigger it. Fixed by adding a std::recursive_mutex (recursive since these methods call each other), locked in place with no restructuring. Verified: test_sdma_stress went from ~5/300 crashes to 300/300 pass.

### DIFF
--- a/include/mori/utils/mori_log.hpp
+++ b/include/mori/utils/mori_log.hpp
@@ -26,6 +26,7 @@
 #include <chrono>
 #include <cstdlib>
 #include <memory>
+#include <mutex>
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
@@ -58,6 +59,7 @@ class ModuleLogger {
  public:
   // Initialize a module-specific logger
   void InitModule(const std::string& moduleName, Level level = Level::ERROR) {
+    std::lock_guard<std::recursive_mutex> lock(mutex_);
     // Check if logger already exists
     auto existing_logger = spdlog::get(moduleName);
     std::shared_ptr<spdlog::logger> logger;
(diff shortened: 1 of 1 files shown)


kawhil-amd (approved):

include/mori/utils/mori_log.hpp
```diff
@@ -196,6 +216,7 @@ class ModuleLogger {
 
   // Allow access to loggers for advanced configuration
   std::unordered_map<std::string, std::shared_ptr<spdlog::logger>>& GetLoggers() {
+    std::lock_guard<std::recursive_mutex> lock(mutex_);
```
jhchouuu, line 224: This lock doesn't protect anything: it's released when the function returns, so the caller ends up holding an unguarded reference to `loggers_`. And both call sites use it to iterate:

```cpp
// :438
for (auto& [name, moduleLogger] : logger.GetLoggers()) moduleLogger->set_pattern(logPattern);
// :450
for (auto& [name, moduleLogger] : logger.GetLoggers()) moduleLogger->sinks().push_back(file_sink);
```

If another thread hits a first-time `GetLogger()` during either loop, the insert can rehash `loggers_` and invalidate the iterator — the same failure mode this PR fixes. Probability is low (both sites sit on the env-var init path), but as written the lock is worse than none: it reads as protected when it isn't.

Suggest not handing out the reference at all:

```cpp
template <typename F>
void ForEachLogger(F&& fn) {
  std::lock_guard<std::recursive_mutex> lock(mutex_);
  for (auto& [name, lg] : loggers_) fn(name, lg);
}
```

then `logger.ForEachLogger([&](auto&, auto& lg) { lg->set_pattern(logPattern); });`. If you'd rather keep the accessor, at least drop the `lock_guard` and document that callers must synchronize themselves.

PR #531, opened 2026-08-07
test(ep): add cross-dtype and quantized coverage for dispatch/combine

amd-wsung102:
## Motivation
This PR is motivated by the task described [here](https://amd-hub.atlassian.net/jira/software/c/projects/AIMODELS/boards/2992?assignee=712020%3Acedccbdf-bb1e-4e04-8c3b-edcdd7744eda&selectedIssue=AIMODELS-996).

## Summary

Closes the correctness-coverage gaps for quantized dispatch/combine: FP8 direct-cast and FP4 dispatch, including cross-dtype pairings such as FP4 dispatch + BF16 combine, across IntraNode / IntraNodeLL / InterNodeV1 / InterNodeV1LL / AsyncLL and all three NIC device backends.

The root cause was narrower than "FP8/FP4 are untested". The shared test harness derived the combine element type from `config.data_type`, so no pytest could express a dispatch/combine pair with *different* dtypes — every test sat on the same-dtype diagonal. Two consequences:

- `test_dispatch_combine_internode_v1.py` contained **zero** `quant_type` references, so FP8 direct-cast was never exercised on InterNodeV1/V1LL.
- FP4 dispatch + BF16 combine — the configuration `tools/run_all_internode_tuning.sh` benchmarks for its `v1_fp4`, `v1_ll_fp4` and `async_ll_fp4` rows — had no correctness coverage anywhere. It was being tuned and benchmarked, but never verified.

## Changes

- **`dispatch_combine_test_utils.py`** — lifts the cross-dtype mechanism that already existed in `bench_dispatch_combine.py` into the shared harness (`combine_data_type` / `combine_hidden_dim` / `dispatch_hidden_dim`, plus `_get_combine_input` handling the FP4 unpack and the zero-copy staging copy). Adds `cross_dtype_hidden_dims` for the FP4 packed/unpacked hidden-dim math and `cross_dtype_skip_reason` so every kernel-type file agrees on which matrix cells are legitimately out of scope. Defaults preserve existing behaviour exactly.
- **`test_dispatch_combine_intranode.py` / `_internode_v1.py` / `_async_ll.py`** — new `test_dispatch_combine_cross_dtype` matrices sweeping dispatch dtype × BF16 combine × `{none, fp8_direct_cast}`.
- **`test_dispatch_combine_internode_v1.py`** — also adds `test_blockwise_combine_unsupported_internode`, pinning the internode blockwise gap (see below).
- **`tools/run_quant_dtype_coverage.sh`** (new) — sweeps the NIC backend across the quantized/cross-dtype selections and prints a pass/fail matrix.

84 new test cases (64 run, 20 skipped on gfx950 — the FNUZ FP8 rows, which are MI300-only).

## NIC coverage

`MORI_DEVICE_NIC` ∈ `{mlx5, bnxt, ionic}` maps to MLX / BRCM / AINIC, feeds `-DMORI_DEVICE_NIC_*` into the JIT, and is part of the JIT cache key — each value produces a distinct kernel build (confirmed: three separate cache trees with differing `shmem_kernels.hsaco` sizes). Sweeping it validates that the quantized kernels codegen and execute correctly against every NIC's device primitives even on a host with one vendor installed, or none. This is particularly useful for BRCM now that #516 removed bnxt CI.

## Known gaps deliberately left open

- **Internode blockwise combine.** `fp8_blockwise` works on IntraNode/IntraNodeLL/AsyncLL; `fp4_blockwise` is IntraNode/IntraNodeLL only; both are rejected on all internode kernel types. Closing this needs new kernels, not test wiring, so the new test asserts the rejection and fails with instructions the moment support lands.
- **C++/Python guard divergence.** `launch.cpp:524` permits blockwise only on `IntraNode`, while the Python path also permits `IntraNodeLL` and (for fp8) `AsyncLL`. AOT/C++ consumers therefore get a `runtime_error` on configs that work from Python. Not touched here; worth a separate issue.
- **Pre-existing session hang.** Running the three EP test files in a *single* pytest invocation stalls around 67% in the InterNodeV1 region. Confirmed to reproduce at HEAD with this branch reverted, so it is not introduced here — CI already sidesteps it by invoking each file separately, as does the new script.

## Test plan

All on 8× MI355X (gfx950, OCP FP8).

- [x] Per-file regression (matching CI's per-file invocation): intranode 359 passed, internode_v1 189 passed, async_ll 206 passed — **754 passed, 0 failed**.
- [x] `tools/run_quant_dtype_coverage.sh --all-nics`: 9/9 groups PASS (3 NICs × 3 kernel families, 60 cases per NIC).
- [x] All 6 shipped tuning-matrix rows verified through the real internode entry point with `--cmd test` (CI only runs `--cmd bench`, which skips verification). Flagship FP4 + BF16 + fp8_direct_cast on `v1`: **500/500 rounds, `error times: 0` on all 8 ranks, exit 0.**
- [x] `black --check` clean; `ruff` findings identical to the pre-existing baseline (6, all pre-existing).

## Submission Checklist


tools/run_quant_dtype_coverage.sh
jhchouuu, line 1: This script has not been used. Do you hope to have it added to the main body?
if not, please delete this file.

jhchouuu (approved):
LGTM
thanks for your work!

tools/run_quant_dtype_coverage.sh
amd-wsung102: Yes that's a good catch, the file was not used. I have deleted the file. Thanks!

PR #530, opened 2026-08-06
perf(EPv2): update gfx1250 tune config + optimize kernel impl

kawhil-amd:
### Summary

  Kernel-level optimizations for EPv2 intranode dispatch and combine, targeting memory-level parallelism, synchronization overhead, and architecture-adaptive tuning.

  ### Changes
*  Combine gather: per-block flag entry barrier, exit barrier removed. Replace `comb_bar` atomic sync with per-block monotonic flag counters — each block owns a private counter (single writer, no atomic). Exit barrier deleted entirely (was the dominant latency source).
*  Dispatch load unroll num update.

PR #536, opened 2026-08-10
fix(EPv2):  scale i64 buffer offsets in the flydsl 0.3.0 compat shim

kawhil-amd:
## Summary
  Fixes a hang in the EPv2 gather-combine entry barrier that reproduces on flydsl 0.3.0 for any block_num >= 2. Two commits:

  1. Root cause — i64 buffer offsets in the flydsl 0.3.0 compat shim. On flydsl 0.3.0 the flydsl.expr.buffer_ops submodule is gone, so flydsl_compat.py
  emulates buffer_load/buffer_store with an i32-typed pointer plus ptr_load/ptr_store(rsrc + offset). LLVM pointer arithmetic steps by the i32 pointee (4
  bytes), but the 0.2.x buffer_ops take the offset in elements and scale to bytes by element width. For i32 the two agree; for the i64 per-block xdb flag
  counters (added in #530) they diverge — block bid read/wrote byte bid*4 instead of bid*8, splicing two adjacent i64 slots. With the all-ones flag init,
  block 1 reads slot0.hi | slot1.lo = 0x1_0000_0000 and spins forever (peers only push 1). Fixed by scaling the offset by element_bytes/4 (i32 → 1, i64
  → 2), via a _dwords() helper that fails fast (NotImplementedError) on sub-4-byte element types.
  2. Guard rail — assert block_num <= xdb_flag_slots. The barrier shares one 256-slot cross_device_flag buffer across combine variants. Added an assert in
  make_combine so an oversized grid can't silently overflow the per-block xdb flag array.

### DIFF
--- a/python/mori/ops/dispatch_combine_v2/intranode_kernels.py
+++ b/python/mori/ops/dispatch_combine_v2/intranode_kernels.py
@@ -596,6 +596,10 @@ def make_combine(
     _s3_cache=2,
     _unroll=4,
 ):
+    assert block_num <= xdb_flag_slots, (
+        f"combine block_num {block_num} exceeds xdb_flag_slots {xdb_flag_slots}: "
+        f"the per-block xdb flag array would overflow"
+    )
     # Transport dtype = external dtype, except fp8_direct_cast wires fp8 while
     # the output (comb_out) stays bf16 (2 i32 per fp8 i32 unit). fp4: i32 = 8 fp4.
     _to_accum2, _from_accum2, _zero_accum = _accum_funcs(
--- a/python/mori/ops/dispatch_combine_v2/flydsl_compat.py
+++ b/python/mori/ops/dispatch_combine_v2/flydsl_compat.py
@@ -121,17 +121,37 @@ except ModuleNotFoundError:  # flydsl >= 0.3.0
             inttoptr(pty, addr_i64), num_records_bytes=num_records_bytes
         )
 
+    def _dwords(elem):
+        # The base ptr is i32, so pointer arithmetic steps 4 bytes; 0.2.x
+        # buffer_ops instead took `offset` in ELEMENTS. Convert element offsets
+        # to i32 units. An i64 access (the per-block xdb flag counters) is 2
+        # units — without this it would index at half its stride, read a spliced
+        # garbage flag, and hang the combine entry barrier.
+        n = elem.width // 32
+        if n < 1:
+            raise NotImplementedError(
+                f"0.3.0 shim indexes in 4-byte units; {elem} is unsupported"
+            )
+        return n
+
+    def _elem_type_of(data):
+        ty = _arith.unwrap(data).type
+        try:
+            return _ir.VectorType(ty).element_type
+        except (ValueError, TypeError):
+            return ty
+
     def buffer_load(rsrc, offset, vec_width=4, dtype=None, mask=None, cache_modifier=0):
         """Load `vec_width` x `dtype` at element `offset` of `rsrc`."""
         elem = dtype if dtype is not None else T.i32()
         ty = elem if vec_width == 1 else T.VectorType.get([vec_width], elem)
         # unwrap to an ArithValue, as 0.2.x returned: callers do arithmetic on
         # the result and also feed it to arith.* ops, which need an ir.Value.
-        return _arith.unwrap(ptr_load(rsrc + offset, result_type=ty))
+        return _arith.unwrap(ptr_load(rsrc + offset * _dwords(elem), result_type=ty))
 
     def buffer_store(data, rsrc, offset, mask=None, cache_modifier=0):
         """Store `data` at element `offset` of `rsrc`."""
-        return ptr_store(data, rsrc + offset)
+        return ptr_store(data, rsrc + offset * _dwords(_elem_type_of(data)))
 
 
 logger.debug(


jhchouuu:
Thanks for chasing this. I don't think this patch addresses the hang, and I believe I've found the actual cause. Details below.

## Why this patch is a no-op for the reported hang

The repro config (EP4, hidden=7168, topk=6) hits `_GFX1250_SCHED_BF16_T6`, whose last bucket is `(None, 256, 32, 256, 16)` — so `block_num == 256 == xdb_flag_slots`. In that case:

- the new `assert block_num <= xdb_flag_slots` passes (`256 <= 256`), so it doesn't catch anything;
- the new tail-fill guard is `slot < xdb_flag_slots` **and** `slot >= grid_num`, i.e. `slot < 256 and slot >= 256` — never true, so zero stores are emitted;
- the old code was `const_expr(xdb_flag_slots > block_num)` = `256 > 256` = False, which compiled the whole block away — also zero stores.

Both versions emit the same thing. Also, `fx.grid_dim.x` is always equal to `block_num` here: `make_combine` has exactly one caller, and the kernel it builds is launched with `.launch(grid=(block_num, 1, 1))` from the same closure (`intranode_kernels.py:904`). There's no path where the launch grid differs from the compile-time `block_num`.

## Actual root cause: the FlyDSL 0.3.0 compat shim indexes in 4-byte units, and the xdb flag is i64

`flydsl_compat.py:115-133` builds an **i32-typed** pointer and then does plain pointer arithmetic:

```python
pty = PointerType.get(T.i32(), AddressSpace.Global)   # <- i32 pointee
...
def buffer_load(rsrc, offset, vec_width=4, dtype=None, ...):
    return _arith.unwrap(ptr_load(rsrc + offset, result_type=ty))
def buffer_store(data, rsrc, offset, ...):
    return ptr_store(data, rsrc + offset)
```

In FlyDSL 0.3.x, `Pointer.__add__` maps to `add_offset`, which shifts by *offset elements of the pointer's element type* (`flydsl/expr/typing.py:995` → `primitive.py`, docstring: "Shift *ptr* by *offset* elements"). The pointee is hardcoded to i32, so **every offset advances 4 bytes regardless of the access dtype**. The shim's own comment states the assumption: *"i32-typed: every EPv2 access indexes in 4-byte units."*

That assumption held when the shim landed in #529. #530 then added the per-block xdb counter as **i64**, which is the one access in the whole file that breaks it. A census of every buffer access in `dispatch_combine_v2`:

| file | i32 | f32 | i64 |
|---|---|---|---|
| `intranode_kernels.py` | 26 | 6 | **1** |
| `flydsl_prims.py` | – | – | 0 |

All three xdb-flag accesses are affected:

- `intranode_kernels.py:648` — `buffer_load(rsrc_xdb_flag, bid, vec_width=1, dtype=T.i64())` wants byte `bid*8`, gets `bid*4`
- `intranode_kernels.py:656` — `buffer_store(phase + 1, rsrc_xdb_flag, bid)` writes 8 bytes on a 4-byte grid, so adjacent blocks' counters overlap
- `intranode_kernels.py:670-673` — the tail-fill store, same problem (so the tail fill has never been writing the slots it thinks it is)

**Failure trace.** `cross_device_flag` is initialised to all-ones i64:

```
bytes:  [01 00 00 00 00 00 00 00][01 00 00 00 00 00 00 00] ...
block 0 -> reads byte 0 -> 1              -> pushes 1, spins for >= 1 -> passes
block 1 -> reads byte 4 -> 0x1_00000000   -> spins for >= 4294967296, peer only ever pushes 1 -> hangs forever
```

So **any `block_num >= 2` deadlocks on the first combine**, independent of token count, schedule table, or whether `block_num` changes between calls. That matches the report that `SWEEP=128` alone hangs on the very first combine.

It also explains why only the local counter array is broken while the cross-device side is fine: the peer push and the poll go through `flydsl_prims.py`, which scales addresses by an explicit byte size (`_addr_plus(base, offset, elem_bytes)` at `:56-58`; `store_i64_system` passes `8` at `:89-92`). The codebase has two address-arithmetic paths — one explicit-byte (correct) and one pointee-typed (i32-only) — and the i64 counter went through the wrong one.

## This is a 0.3.0-only regression

FlyDSL 0.2.x scales the offset by the **access** dtype, not by a pointer type (`python/flydsl/expr/buffer_ops.py` @ v0.2.4):

```python
# buffer_load, :513-515
# IMPORTANT: Buffer load offset is in BYTES, not elements!
element_bytes = dtype.width // 8          # i64 -> 8

# buffer_store, :612-622
element_bytes = element_type.width // 8   # from data.type; i64 -> 8
```

So the same mori source is correct on 0.2.x and broken on >= 0.3.0. No backport needed — but anyone on 0.3.x is hitting this.

## Suggested fix

1. **Fix the scaling in the shim.** Since the pointee is i32, scale the offset by the dtype's dword count:

```python
def _dwords(elem):
    n = elem.width // 32
    if n < 1:
        raise NotImplementedError(f"0.3.0 shim indexes in 4-byte units; {elem} is unsupported")
    return n

def buffer_load(rsrc, offset, vec_width=4, dtype=None, mask=None, cache_modifier=0):
    elem = dtype if dtype is not None else T.i32()
    ty = elem if vec_width == 1 else T.VectorType.get([vec_width], elem)
    return _arith.unwrap(ptr_load(rsrc + offset * _dwords(elem), result_type=ty))

def buffer_store(data, rsrc, offset, mask=None, cache_modifier=0):
    return ptr_store(data, rsrc + offset * _dwords(_elem_type_of(data)))
```

`buffer_store` is the more dangerous of the two because it takes no `dtype` at all — the width comes silently from `data`.

2. **Make the invariant loud.** The `raise` above turns "GPU hangs" into a trace-time error, so the next i64/i16 access can't reintroduce this silently. Worth adding even if you take a different fix for the scaling.

3. Optionally, a cheaper alternative for this specific case: route the three xdb-flag accesses through the explicit-byte path in `flydsl_prims.py`, which is what the peer side already does.

## On the `assert` in this PR

Please keep it — it catches a *different* real bug. `_MI300X_DEFAULT = None` falls through to `_cu_scaled_default()`, which sets `combine_block_num = cu = 304` on MI300X, i.e. `block_num > xdb_flag_slots`. On the 0.2.x path (8-byte scaling) block 303 writes at byte 2424 of a 2048-byte tensor — a genuine OOB, and `make_buffer_ptr` is built with `num_records_bytes=None` → `0xFFFFFFFF`, so there is no hardware bounds check. It'd be good to land it together with a clamp, e.g. `combine_block_num=min(cu, xdb_flag_slots)` in `_cu_scaled_default()`.

## Quick confirmation

Pinning combine `block_num=1` should make the hang disappear (only block 0 runs, and it reads byte 0, which is the one correct slot); `block_num=2` should hang. If that holds, the diagnosis above is confirmed and independent of the schedule table.

python/mori/ops/dispatch_combine_v2/intranode_kernels.py
```diff
@@ -596,6 +596,10 @@ def make_combine(
     _s3_cache=2,
     _unroll=4,
 ):
+    assert block_num <= xdb_flag_slots, (
+        f"combine block_num {block_num} exceeds xdb_flag_slots {xdb_flag_slots}: "
+        f"the per-block xdb flag array would overflow"
+    )
```
jhchouuu, line 602: maybe num of xdb slots should greater than 304 to compatible with MI300x...

jhchouuu (approved):

python/mori/ops/dispatch_combine_v2/intranode_kernels.py
kawhil-amd: ...ok,  there have been no tune config where it has exceeded the limit. Will change it when best block num over 256.

PR #546, opened 2026-08-12
Fix silent cross-node combine corruption in InterNodeV1 (#475)

### LINKED ISSUE #475: InterNodeV1 combine returns silently-wrong results cross-node (Pollara / ionic), plus several latent bugs
## Summary

`EpDispatchCombineKernelType.InterNodeV1` produces **silently incorrect combine aggregates** for a small fraction of tokens (~0.5–2% per rank) when run across two nodes over an AMD Pensando ("Pollara", `ibv vendor_id 0x1dd8`, `ionic` provider) RoCE fabric, once tokens-per-rank exceeds ~128. There is no crash and no hang — the returned tensor is just wrong, which a real MoE workload would silently consume.

**Dispatch is provably correct**; only the cross-node combine return path is affected. Single-node (`gpu_per_node=8`, world 8) is clean. I tried four independent kernel-level fixes (memory fences, chunk retry, a watermark type fix + arrival barrier, and quiet-before-barrier) — **none eliminate the corruption** — which points the root cause below the kernel, at the provider/firmware ordering layer. Details, and a set of latent bugs found along the way, are below. This issue also doubles as a tracker for those secondary findings (each can be split out if you prefer).

## Environment

- 2 nodes × 8× AMD Instinct MI355X (gfx950), single-node XGMI scale-up, EP world size 16.
- Inter-node fabric: AMD Pensando NIC, `vendor_id 0x1dd8`, RoCE (Ethernet link layer); mori JIT arch/nic key `gfx950_ionic`; `ccqe` disabled (build default).
- Image: `rocm/sgl-dev:sglang-0.5.14-rocm720-mi35x-mori-0701` (torch 2.9.1+rocm7.2, mori source install). Line numbers below are from that image's `mori` tree.
- Kernel config: `InterNodeV1`, `block_num=96, rdma_block_num=64, warp_num_per_block=8`, `use_external_inp_buf=True`, `max_token_type_size=2` (bf16), `num_qp_per_pe=1`, `quant_type="none"`, hidden 7168, top-k 8, 256 experts (16/rank), uniform routing.

## Symptom

Measured with a 16-rank probe that mirrors a normal dispatch→combine round trip. Using a constant per-token payload lets the realized combine multiplier `c_t` be recovered exactly and compared to the expected `m_t` = the token's number of distinct destination ranks (the same expectation your `dispatch_combine_test_utils.check_combine_result` uses: `input × unique_pes`).

- `T ≤ 16`: exact (≈1 bf16 ulp).
- `T = 32..128`: occasional single-token errors on some ranks (run-dependent).
- `T = 256`: ~1–3 wrong tokens/rank, every run.
- `T = 512`: ~4–8 wrong tokens/rank, every run; worst-case relative error up to ~500×.

Error anatomy — two matched classes:

- **Dropped**: `c_t == l_t` exactly (`l_t` = distinct *local-node* destinations) — the token's entire remote-node partial is missing.
- **Polluted**: `c_t > m_t`, often non-integer — a *foreign* token's payload was summed in. The two classes are consistent with a read of stale/zeroed destination metadata aliasing to PE 0 (on node 0 that alias is local → over-accumulation; on node 1 it's remote → silent drop), matching the observed node polarity.

## What is ruled out (each with an on-metal experiment)

| Hypothesis | Result |
|---|---|
| Fabric / GPUDirect broken | Ruled out — RCCL world-16 allreduce, `shmem` init, op construction, and dispatch all succeed |
| Dispatch corruption | Ruled out — receive counts match an all-gathered routing oracle on all 16 ranks; constant-payload check shows 0 torn rows / 0 out-of-domain values across thousands of received rows per rank |
| Same-kernel intranode | Ruled out — world 8, `gpu_per_node=8` is clean through T=512 |
| Launch geometry | Ruled out — `rdma_block_num` 64→128 unchanged |
| QP count | Ruled out — `num_qp_per_pe` 1→4 unchanged |
| Oracle semantics | Ruled out — matches your own `check_combine_result` |

## Kernel-level fixes attempted (all compiled + presence-verified, all refuted)

| Patch | Change | Result |
|---|---|---|
| Fences | release fence before the chunk-completion `atomicAdd`, acquire fence before the RDMA return put | no change |
| Retry | make the chunk retry loop actually retry unarrived chunks (see bug #2) | no change |
| Watermark + arrival barrier | fix the watermark type (bug #1), then spin on the per-node watermark — the dispatcher's last RDMA op on the QP — before consuming any chunk, so RC ordering guarantees all payloads + flags have landed | no change |
| + Quiet | above + `ShmemQuietThread(peer)` to drain pending return puts before the cross-device barrier atomics (as the LL path does) | no change; occasional errors now visible even at T=32 |

## Hypothesis

Every mechanism the kernel controls was verified or forced — payload delivery (exact), chunk-flag arrival (waited on), put completion (drained), GPU-side ordering (fenced) — and the corruption survives all of it. The remaining candidate is a layer the kernel can't see: whether, on this NIC/provider, an observed completion / flag / atomic implies the corresponding DMA'd payload is actually readable in GPU memory. The signature (sparse, size-scaling, timing-dependent; exact-zeros consistent with reads *before* placement, non-integer mixtures consistent with reads *racing* placement) is the classic completion-before-placement / relaxed-ordering pathology of GPUDirect-style paths. One knob I did not exercise: `ionic` CCQE (default off in this build).

## Latent bugs found while investigating (independent of the above)

1. **Watermark type mismatch.** Combine reads the per-node token watermark (`nodeRecvTokenNum`) through an `index_t` (32-bit) pointer (`internode_v1.cpp:837`), but dispatch publishes it with `uint64_t` remote atomics (`:256`, `:327`). On LE, node ≥ 1 reads the high half of entry 0 → always 0, so the empty-chunk classification is dead code. Confirmed empirically: instrumented give-up events all report `nodeFlag=0` long after dispatch completes.
2. **Dead retry loop.** In the combine chunk loop, `processedMask |= (1u << relativeIdx)` is set unconditionally (~`internode_v1.cpp:899`), so a chunk whose flag has not arrived is marked done and skipped permanently — the enclosing `while (processedMask != full)` retry can never retry. (Interacts with #1: with the watermark fixed, this needs a bounded/correct retry design.)
3. **HT vs LL asymmetry.** `CombineInterNodeTyped` lacks the `__threadfence_system()` and `ShmemQuietThread` calls that `CombineInterNodeLLTyped` has at the analogous points. (Adding them did not fix the corruption above, so filing as a question: is this intentional?)
4. **JIT cache invalidation hole.** The JIT content hash covers `src/ops/kernels` + `include/mori` only, but those `.hip` files `#include` the kernel bodies from `src/ops/dispatch_combine/*.cpp`, which are not hashed. Editing a kernel body silently reuses the stale cached `.hsaco`. (Fix in #469.)
5. **Cold-start JIT race.** Concurrent first-use JIT from many ranks into the shared cache dir can wedge; pre-compiling once per node avoids it. (Observed once — lower confidence.)
6. **No fail-fast on `rdma_block_num > block_num`.** This configuration hard-hangs the first kernel launch rather than being rejected.
7. **Test coverage gap.** `test_dispatch_combine_internode_v1.py` parametrizes only `world_size=8, gpu_per_node=8` (the actual-RDMA `gpu_per_node=4` case is commented out) and caps `max_num_inp_token_per_rank` at 128 — i.e. the cross-node path and the >128-token regime where this corruption appears are never exercised in CI.

## Reproduction

Standalone 16-rank (2×8) probe, no framework needed: build a uniform top-8-of-256 routing trace, run `dispatch → combine`, and compare each token's realized multiplier against its distinct-destination-rank count. Constant per-token payloads make dropped vs polluted tokens directly visible. Launch via `torchrun --nnodes=2 --nproc_per_node=8` with `MORI_DISABLE_AUTO_XGMI=0 MORI_ENABLE_SDMA=1` and the RDMA device list. Happy to share the exact probe scripts.

When patching kernels to reproduce the fix attempts, note bug #4: `rm -rf` the JIT cache dir after editing `src/ops/dispatch_combine/*.cpp` or the stale object is reused.

QizhouZhang97:
Fixes ROCm/mori#475.

## Root cause
`EpCombineAll` reads `args.tokenIndices[tokenId * topk + lane]`, where `tokenId`
is this rank's own token id — the same key as `interNodeDispSendMap`. But
`combine()` bound the caller's argument there, and callers pass `dispatch()`'s
returned `out_idx`, which is recv-slot layout. The node predicate then answers
from an unrelated token's routing: spurious reads pollute (42/42 local-only
tokens), missed reads drop remote copies (0.387%, matching `(1/2)^8`).

## Fix
Kernels unchanged. `dispatch()` retains the indices it scattered with;
`combine()` reduces over those. No-op for correct callers.

## Validation
Two nodes, world=16: probe PASS at T=16–512, previously FAIL at every step.


report_475_reproduction.md
jhchouuu, line 1: delete log file

jhchouuu (approved):

tests/python/ops/bench_dispatch_combine.py
```diff
@@ -355,7 +355,7 @@ def run_once_test(
         combine_output, _ = op.combine(
             combine_input,
             None,
-            dispatch_indices,
+            all_rank_indices[self.config.rank],
```
jhchouuu, line 358: why need do this change, wrong indices usage?

jhchouuu:
The root-cause analysis is right, and I verified it independently against the kernel: `EpCombineAll` keys `args.tokenIndices` by the rank's own `tokenId` — the same key as `interNodeDispSendMap` — so feeding it `dispatch()`'s returned `out_idx` makes the node predicate answer from an unrelated token's routing. The bench fix and the new `combine()` docstring are both good; the docstring is the first time this contract appears on the API surface.

My concern is with the override itself.

## The override breaks the routing handle's layer-private guarantee

`self._dispatch_indices` is **op-global**, but it is fed to the kernel alongside the routing handle's **per-layer** fields. That directly contradicts what the routing handle exists for:

```python
# tests/python/ops/test_dispatch_combine_routing_handle.py:225
def _stale_symmetric_buffer_guard(rank, world_size):
    """Verify the disp_tok_id_to_src_tok_id_local snapshot is layer-private (IntraNode only)."""
    ...
    d0, R0 = _do_dispatch(op_shared, td0, return_routing=True)
    _      = _do_dispatch(op_shared, td1, return_routing=True)   # intervening dispatch
    out_shared, _ = _do_combine(op_shared, d0, td0[1][rank], routing=R0)
```

Here the caller passes the **correct** tensor (`td0[1][rank]`), but after this PR the single-slot stash has already been overwritten by `td1`, so the kernel gets `R0`'s maps paired with `td1`'s `tokenIndices`. If you always combined right after dispatching you would not need a handle at all — surviving an intervening dispatch *is* the handle's purpose.

This is latent rather than currently-red: IntraNode's combine never reads `tokenIndices`, and the v1 routing tests run `gpu_per_node=world_size` so `nNodes==1` and the predicate is degenerate. It is the same shape of latent bug as #475, only this time introduced by the fix.

## Three smaller things

1. The new `combine_indices.size(0) < cur_n` check can't fire on the path it documents. `cur_n` comes from the handle set by the last `dispatch()`, and `combine_indices` is that same dispatch's tensor, so the two are equal by construction. It also cannot catch the `out_idx` misuse, since `out_idx.size(0) == max_recv >= cur_n`.
2. The routing-path validation at `:1369-1376` checks `indices`, but the kernel now receives `combine_indices` — the check no longer guards the tensor that is actually used.
3. `combine_standard_moe()` gets the override without the accompanying check, and `_dispatch_indices` is never cleared (not in `reset()`).

## Suggestion: enforce the contract instead of guessing it

The routing path already has exactly the right check; only the default path is unguarded. `cur_n` is already computed a few lines above, so this is one addition and no new state:

```python
cur_n = (
    self._routing_source_token_count(routing)
    if routing is not None
    else self._get_cur_rank_num_token(self._handle)
)
+ if routing is None and int(indices.size(0)) != cur_n:
+     raise ValueError(
+         f"combine() indices has {int(indices.size(0))} rows but this rank dispatched {cur_n} "
+         f"tokens. Pass this rank's own [num_tokens, topk] routing -- the tensor given to "
+         f"dispatch() -- not dispatch()'s returned out_idx (ROCm/mori#475)."
+     )
```

and drop the four stash/override hunks (`dispatch()`, `combine()`, `dispatch_standard_moe()`, `combine_standard_moe()`).

Compared with the current approach this keeps the routing-handle semantics intact, keeps validation and use on the same tensor, covers `combine_standard_moe` for free, needs no lifetime management — and, importantly, makes the bug **testable**: a pure-Python unit test can assert that passing recv-layout indices raises, with no GPU and no RDMA. With a silent override there is no way to write a regression test for #475, since the wrong call can never fail again.

It would also have turned this into a one-line error message on the reporter's first run, instead of a month spent suspecting NIC firmware.

## Minor

The three repro/report files are added at the repository root. `repro_475_combine_probe.py` and `repro_475_launch_2node.sh` would fit better under `tools/` next to `run_internode_test.sh`, and `report_475_reproduction.md` reads more like a PR artifact than repository content.

One correction for the report's §4: the cross-node combine path **is** covered in CI. `ci.yml`'s `internode-test` job runs on two physical hosts, and `tools/run_internode_test.sh` launches `examples/ops/dispatch_combine/test_dispatch_combine_internode.py`, which re-interprets torchrun's `WORLD_SIZE` as the node count and spawns `GPU_PER_NODE=8` processes per node — so it runs `world=16, gpu_per_node=8, nNodes=2` on both `v1` and `v1_ll`, up to `--max-tokens 4096`, with verification on (`--skip-verify` appears nowhere in CI). It passes because that example passes `all_rank_indices[self.rank]`. The gap in `test_dispatch_combine_internode_v1.py` is real, but the conclusion "the cross-node combine path is never actually exercised" is not — which reinforces the point above: this was never a coverage problem, it was an unenforced contract.

tests/python/ops/bench_dispatch_combine.py
QizhouZhang97: Adopte. Dropped the four stash/override hunks; combine() now
validates indices.size(0) == cur_n on both paths, which covers
combine_standard_moe too. No cross-call state remains, so the routing handle
stays layer-private. Also fixed two callers still passing recv-layout indices

jhchouuu (approved):
ready to merge

PR #548, opened 2026-08-12
feat(EPv2): HIP/JIT backend for intranode dispatch/combine + gfx1250 TDM 

jhchouuu:
## What this is

A second kernel backend for the v2 EP intranode op: the kernels are generated, compiled and launched from C++ instead of FlyDSL, on cco-LSA. It sits behind the existing `EpDispatchCombineOp`, so callers pick it with `cfg.kernel_backend="hip"` or `MORI_V2_KERNEL_BACKEND=hip` and change nothing else. FlyDSL stays the default.

## The idea

One chain holds the whole design together:

```
Cfg value  =  specialisation identity  =  rendered source text  =  sha256 cache key
```

A configuration can only reach hipcc by becoming an `EpCfg` field and being rendered into that text, and that text is hashed whole. There is no second channel — no out-of-band `-D`, no hand-assembled key fragment — so "the config changed but the old binary was reused" is not a bug that has to be avoided, it is a state that cannot be constructed.

- `Cfg` is a single C++20 structural NTTP, so every in-kernel switch is `if constexpr (kCfg.x)` rather than a `#if` whose condition arrived out of band.
- Cache key is `name $$ compiler $$ nic $$ flags $$ include-hash $$ source-text`, published by directory rename (build into `tmp/<uuid>`, fsync, rename).
- The entry symbol is generated by `Spec::EntryName(cfg)` and describes the kernel — `mori_ep_dispatch_tdm_byte8_ws4_h3584_k8_128x16` — so a profile can tell dispatch from combine, one body from the other, and one tuned geometry from the next. The name is spelled once, in C++, and the same string is both interpolated into the source and passed to `hipModuleGetFunction`; Python never builds a kernel name.
- Host never needs hipcc: the Cfg header is HIP-free and the shared arithmetic is attribute-free `constexpr`. A CI guard enforces it.
- `Prepare`/`Launch` are separate so no launch can fork a compiler mid graph-capture.

Design doc: `docs/MORI_JIT_V2_DESIGN.md`.

## Two backends, one op

The base class owns the op state (arena, scratch, variant table, routing handle) and `dispatch`/`combine`; a backend supplies `_regions`, `_build_kernels` and `_unsupported`. Behavioural differences are *data* on `KernelSet`, not overrides. Selecting a backend imports only that one, so the HIP backend runs on a machine with no FlyDSL. The capability gate runs at construction, before the arena is allocated, so an unsupported config fails loudly instead of producing wrong numbers at launch.

## gfx1250

`origin/main`'s TDM intranode kernel is brought in as a second device body, selected by arch in `RenderSource` — host-side, not a device `#if`, so the render-time arch is the one hipcc gets and the choice lands in the cache key.

Dispatch also takes fp8/fp4 payloads now. No new kernel was needed: the dispatch leg only copies its payload, so a narrower element is plumbing (`EpDType::Byte8`, with the caller halving hiddenDim for fp4). Combine reduces, so `MakeEpCfg` rejects a byte dtype there. mori does no quantizing here — the payload arrives packed.

## Measured

4x gfx1250 (MI450) EP4, hidden 7168, topk 8, against v1's shipping default, µs:

| tok/rank | 64 | 512 | 2048 | 4096 | 16384 |
|---|---|---|---|---|---|
| dispatch v1 → v2 | 80.9 → 73.1 | 82.8 → 74.7 | 104.6 → 97.4 | 171.4 → 152.0 | 584.5 → 519.6 |
| combine v1 → v2 | 26.9 → 21.3 | 41.7 → 36.9 | 96.7 → 83.8 | 169.6 → 146.4 | 588.7 → 523.0 |

Faster at every size in both phases. At identical geometry the margin comes from the Cfg being a compile-time NTTP: the ISA carries 3 integer divisions where v1's carries 41 in dispatch and 10 where v1 has 57 in combine, and combine issues 120 loads against 225 because a peer address is window arithmetic rather than three dependent loads through a `SymmMemObj`. The rest is the tuning schedule.

A narrower dispatch is worth more again — 4096 tok/rank: bf16 156.6, fp8 98.5, fp4 80.0.

8x MI355 (gfx950) EP8, µs at 128/512/4096 tok/rank: dispatch 47.5/128.8/857.4 against FlyDSL's 76.0/127.8/857.5, combine 57.8/124.4/781.8 against 82.7/133.6/888.8.

## Tuning

`hip_tuning_configs.py` never borrows FlyDSL's schedule — different kernels, different optima. Dispatch and combine get independent tables, because combine does not depend on the dispatch dtype at all and a shared table would carry a fake axis and three copies of one answer. Keyed by `(world, hidden, topk, experts_per_rank)`.

From a 64..16384-token sweep over four geometries, three dtypes and two topk values on both architectures: topk moves the bucket edges, the expert count does not (64 vs 96 agree within ~2%, hence the wildcard), and the dtype only does for fp4 at topk 6. Where the choice was close the smaller geometry won on purpose — fewer blocks means fewer CUs held, which matters when this overlaps an expert GEMM.

## Behaviour change

`combine(input, weights=None, ...)` now returns `None` as its second value instead of a buffer. Passing weights requests the fold and gets it. This restores v1's contract — the fold exists for training backward and inference does not need it — and it is not free: it is topk peer reads of 32 B per token, which on gfx1250 costs more than the entire token payload. In-tree callers are updated; out-of-tree ones that unpack and use the second value will now see `None`.

## Bugs fixed on the way

- A wheel shipped neither the v2 kernel sources nor `libmori_jit.so` / `libmori_ops_v2.so`, so the hip backend could not compile a single kernel off an install. Invisible from a repo checkout, where both resolve out of the tree. The binding CI step now runs from `/tmp` so only the packaged copies can answer.
- The reverse map ("tis") was encoded differently by the two backends — FlyDSL publishes `rank*maxTokPerRank + srcTok`, this backend reused the forward encoder whose stride is `maxRecv` — so the same routing handle decoded differently depending on which backend produced it.
- Exceptions could escape four `extern "C"` entries into ctypes.
- From @kawhil-amd's review: `ResolveArch` passed the caller's arch to the resolver by writing the process environment and reading it back with `getenv`, which is a POSIX data race any multi-threaded server hits while creating plans concurrently — it goes through a mutex-guarded `SetArchOverride()` now. Four schema getters caught everything and returned the failure code without setting the error, so `mori_jit_last_error()` handed back the *previous* message. And the `EpArgs` wire schema, which used to be a string kept parallel to the struct, is generated from one field list that also static_asserts `offsetof` ascending in schema order — `sizeof` alone cannot see two same-type fields swapped, and 8 of the 22 are bare pointers.

## Known gaps

Rejected at construction by the capability gate: scatter combine, `quant_type` quant, StdMoE, per-token scales, routing replay, `local_expert_count`. No cross-node path.

## Verified

- 4x gfx1250: `test_op` across token counts, fp8/fp4 asymmetric dispatch byte-exact via the routing reverse map, HIP graph capture (20 replays), the full tuning sweep.
- 8x MI355: `test_op` under both backends, backend parity element-for-element, fp4 asymmetric under both, the MI355X tuning sweep, C++ 27/27 and binding 18/18 (the latter run from `/tmp`).
- Both device bodies `--genco` clean on both arches; the packaged source tree compiles on its own.
- Four CI steps are added here — jit ctests, the binding tests from `/tmp`, the hip backend under `test_op`, and FlyDSL-vs-HIP parity. All green.

### DIFF
--- a/.github/workflows/ci_cco.yml
+++ b/.github/workflows/ci_cco.yml
@@ -167,14 +167,21 @@ jobs:
             cd $GITHUB_WORKSPACE/build && \
             timeout 600 ctest --output-on-failure -R '^jit_'"
 
-      - name: Run JIT v2 Python binding tests (1 rank)
+      - name: Run JIT v2 Python binding tests (1 rank, outside the repo)
         run: |
           # 18 tests over the ctypes/C-ABI seam -- schema-vs-sizeof, request coercion,
           # unknown-field rejection. Single rank on purpose: EpDispatchPlan(world_size=1)
           # needs no communicator, which is what makes this cheap enough to always run.
+          #
+          # Run from /tmp, not the checkout, and that is the point: building a plan
+          # needs libmori_jit.so, libmori_ops_v2.so and the v2 kernel sources, and from
+          # a repo cwd every one of those resolves out of build/ and src/ even when the
+          # INSTALL ships none of them. Both halves of exactly that bug were live until
+          # this commit's siblings. Outside the repo only the packaged copies can answer.
           $CT exec $CONTAINER bash -c "\
-            cd $GITHUB_WORKSPACE && \
-            timeout 600 python -m pytest tests/python/ops/dispatch_combine_v2/test_jit_binding.py -v"
+            cd /tmp && \
+            timeout 600 python -m pytest \
+              $GITHUB_WORKSPACE/tests/python/ops/dispatch_combine_v2/test_jit_binding.py -v"
 
       - name: Run ops-v2 dispatch/combine test (EP8, 8 ranks)
         run: |
--- a/.github/workflows/ci_cco.yml
+++ b/.github/workflows/ci_cco.yml
@@ -159,25 +159,22 @@ jobs:
 
       - name: Run JIT v2 C++ tests + host/device split guard
         run: |
-          # Registered by tests/cpp/jit/CMakeLists.txt but selected by nothing until
-          # now, so the whole C++ JIT framework was shipping untested. Neither needs a
-          # GPU peer: jit_core exercises render/cache/schema, and the guard only checks
-          # that the Cfg headers still compile without hipcc.
+          # Registered by tests/cpp/jit/CMakeLists.txt but selected by no job before
+          # this one. Neither needs a GPU peer: jit_core exercises render/cache/schema,
+          # the guard only checks the Cfg headers still compile without hipcc.
           $CT exec $CONTAINER bash -c "\
             cd $GITHUB_WORKSPACE/build && \
             timeout 600 ctest --output-on-failure -R '^jit_'"
 
       - name: Run JIT v2 Python binding tests (1 rank, outside the repo)
         run: |
-          # 18 tests over the ctypes/C-ABI seam -- schema-vs-sizeof, request coercion,
-          # unknown-field rejection. Single rank on purpose: EpDispatchPlan(world_size=1)
-          # needs no communicator, which is what makes this cheap enough to always run.
+          # The ctypes/C-ABI seam: schema-vs-sizeof, request coercion, unknown-field
+          # rejection. Single rank -- EpDispatchPlan(world_size=1) needs no communicator.
           #
-          # Run from /tmp, not the checkout, and that is the point: building a plan
-          # needs libmori_jit.so, libmori_ops_v2.so and the v2 kernel sources, and from
-          # a repo cwd every one of those resolves out of build/ and src/ even when the
-          # INSTALL ships none of them. Both halves of exactly that bug were live until
-          # this commit's siblings. Outside the repo only the packaged copies can answer.
+          # Run from /tmp rather than the checkout, and that is the point: a plan needs
+          # libmori_jit.so, libmori_ops_v2.so and the v2 kernel sources, all of which
+          # resolve out of build/ and src/ from a repo cwd even when the install ships
+          # none of them. Outside the repo only the packaged copies can answer.
           $CT exec $CONTAINER bash -c "\
             cd /tmp && \
             timeout 600 python -m pytest \
@@ -191,14 +188,10 @@ jobs:
 
       - name: Run ops-v2 HIP backend (EP8, 8 ranks)
         run: |
-          # The step above drives the FlyDSL backend; nothing covered the HIP one. Both
-          # are worth running because they are separate kernels behind one op: test_op
-          # checks the op layer against the HIP kernels, and the parity test then
-          # compares the two backends element for element on the same input.
-          # test_op.py reports per-case PASS/FAIL and does not set a non-zero exit
-          # code for a failed CHECK, so grep the output as well as the status -- and
-          # treat "no PASS at all" as a failure, or a run that died before the first
-          # case would read as green.
+          # The step above drives the FlyDSL backend; this one drives the HIP kernels
+          # behind the same op. test_op.py reports per-case PASS/FAIL and does not exit
+          # non-zero on a failed CHECK, so grep the output too -- and treat "no PASS at
+          # all" as failure, or a run that died before the first case reads as green.
           $CT exec -e MORI_V2_KERNEL_BACKEND=hip $CONTAINER bash -c "\
             cd $GITHUB_WORKSPACE/tests/python/ops/dispatch_combine_v2 && \
             out=\$(timeout 900 torchrun --standalone --nproc_per_node=8 test_op.py 2>&1) || \
--- a/.github/workflows/ci_cco.yml
+++ b/.github/workflows/ci_cco.yml
@@ -157,11 +157,54 @@ jobs:
             cd $GITHUB_WORKSPACE && \
             timeout 300 pytest tests/python/cco/test_sdma_api.py -v"
 
+      - name: Run JIT v2 C++ tests + host/device split guard
+        run: |
+          # Registered by tests/cpp/jit/CMakeLists.txt but selected by nothing until
+          # now, so the whole C++ JIT framework was shipping untested. Neither needs a
+          # GPU peer: jit_core exercises render/cache/schema, and the guard only checks
+          # that the Cfg headers still compile without hipcc.
+          $CT exec $CONTAINER bash -c "\
+            cd $GITHUB_WORKSPACE/build && \
+            timeout 600 ctest --output-on-failure -R '^jit_'"
+
+      - name: Run JIT v2 Python binding tests (1 rank)
+        run: |
+          # 18 tests over the ctypes/C-ABI seam -- schema-vs-sizeof, request coercion,
+          # unknown-field rejection. Single rank on purpose: EpDispatchPlan(world_size=1)
+          # needs no communicator, which is what makes this cheap enough to always run.
+          $CT exec $CONTAINER bash -c "\
+            cd $GITHUB_WORKSPACE && \
+            timeout 600 python -m pytest tests/python/ops/dispatch_combine_v2/test_jit_binding.py -v"
+
       - name: Run ops-v2 dispatch/combine test (EP8, 8 ranks)
         run: |
           $CT exec $CONTAINER bash -c "\
             cd $GITHUB_WORKSPACE && \
             python -m pytest tests/python/ops/dispatch_combine_v2/test_dispatch_combine_v2_intranode.py -v"
+
+      - name: Run ops-v2 HIP backend (EP8, 8 ranks)
+        run: |
+          # The step above drives the FlyDSL backend; nothing covered the HIP one. Both
+          # are worth running because they are separate kernels behind one op: test_op
+          # checks the op layer against the HIP kernels, and the parity test then
+          # compares the two backends element for element on the same input.
+          # test_op.py reports per-case PASS/FAIL and does not set a non-zero exit
+          # code for a failed CHECK, so grep the output as well as the status -- and
+          # treat "no PASS at all" as a failure, or a run that died before the first
+          # case would read as green.
+          $CT exec -e MORI_V2_KERNEL_BACKEND=hip $CONTAINER bash -c "\
+            cd $GITHUB_WORKSPACE/tests/python/ops/dispatch_combine_v2 && \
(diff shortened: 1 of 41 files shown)


kawhil-amd (changes requested):
## `setenv()` in `ResolveArch` is not thread-safe — UB in multi-threaded Python processes

### The problem

`src/jit/v2/plan_api.cpp:120`:

```cpp
std::string ResolveArch(const char* arch) {
  if (arch && *arch) setenv("MORI_JIT_ARCH", arch, /*overwrite=*/1);
  const std::string resolved = GetToolchain().arch;
  ...
}
```

`setenv()` is not thread-safe per POSIX — it modifies the global environment block. `GetToolchain()` internally calls `DetectArch()` → `getenv("MORI_JIT_ARCH")`. A concurrent `setenv` + `getenv` from two threads is a data race with undefined behavior (crash, heap corruption, or reading a partially-written string).

This is reachable in production: a multi-threaded inference server (e.g., vLLM TP workers) can call `mori_jit_plan_create` from multiple threads during initialization. Even though `GetToolchain()` uses `std::call_once` internally, the `setenv` happens *before* that call, so the race window is open.

### Suggested fix

Replace the env var side-channel with a mutex-protected internal variable:

**toolchain.hpp** — add:
```cpp
void SetArchOverride(const std::string& arch);
```

**toolchain.cpp** — add:
```cpp
namespace {
std::mutex& ArchOverrideMutex() {
  static std::mutex m;
  return m;
}
std::string& ArchOverrideStorage() {
  static std::string s;
  return s;
}
}  // namespace

void SetArchOverride(const std::string& arch) {
  std::lock_guard<std::mutex> lock(ArchOverrideMutex());
  ArchOverrideStorage() = arch;
}
```

And change `DetectArch()`:
```cpp
std::string DetectArch() {
  {
    std::lock_guard<std::mutex> lock(ArchOverrideMutex());
    const std::string& ov = ArchOverrideStorage();
    if (!ov.empty()) return ov;
  }
  if (const char* v = std::getenv("MORI_JIT_ARCH"); v && *v) return v;
  ...
}
```

**plan_api.cpp** — replace:
```cpp
std::string ResolveArch(const char* arch) {
  if (arch && *arch) SetArchOverride(arch);
  const std::string resolved = GetToolchain().arch;
  ...
}
```

This keeps the "first call wins" semantics while eliminating the POSIX data race. The env var `MORI_JIT_ARCH` still works as a static override (set before any JIT call), but the runtime override path no longer touches the process environment.

---

Other notes (lower severity, no action required for merge):

1. **`EpArgs` has no field-order compile-time guard** — the `MORI_EP_ARGS_SCHEMA` string must match the struct field-by-field, but only `sizeof` is checked. Two same-type fields swapped would pass sizeof but silently corrupt kernel arguments. Consider adding `MORI_JIT_ASSERT_FIELD_COUNT(EpArgs, N, ...)` and/or `offsetof` assertions on the first/last field.

2. **`catch(...)` blocks in `mori_jit_plan_args_schema` / `mori_jit_plan_request_schema`** swallow exceptions without calling `SetPlanError()`, so `mori_jit_last_error()` returns the *previous* error. Other entry points (`_create`, `_launch`) handle this correctly.

3. **`flydsl_backend.py:local_expert_count()`** has unreachable code after `return` (`self.cross_device_flag.fill_(1)`) — either a lost reset or dead code to remove.

jhchouuu:
Thanks — all four addressed, one commit each on top of `95fe76bb`.

**`setenv()` in `ResolveArch` (b0eeeaf)** — you're right, and the environment was never the point, only the channel. `SetArchOverride()` now keeps the value in a mutex-guarded string that `DetectArch()` consults ahead of `MORI_JIT_ARCH`, exactly as you laid out. That also stops a JIT call from editing the environment of the process hosting it, which is its own small rudeness in a server. Semantics are unchanged: explicit arch beats the env var, the env var still works as a static override, and asking for an arch after the toolchain has resolved still throws.

**`EpArgs` field-order guard (9c4a551)** — agreed that `sizeof` is not a guard, and `offsetof` on the first and last field would not have caught the case you named either: two swapped same-type fields have the same offsets. So the schema is no longer a parallel string. One X-macro list generates both the schema and an array of `offsetof` *taken in schema order*; if the declaration order disagrees with it in any way the sequence stops increasing and the build fails by name. Verified by swapping `dispDestTokIdMap`/`destPeTokenCounter` — static assertion, not a silent pass. Eight of the twenty-two fields are bare pointers, so this was worth doing properly.

**`catch(...)` not setting the error (555ac71)** — fixed in all four getters (`_args_schema`, `_request_schema`, `_args_size`, `_registered_plans`). Worth spelling out why it mattered more than "no message": `mori_jit_last_error()` returned the *previous* error and the binding raises with that string, so the caller got a message naming a different kernel. `plan_destroy` still swallows silently; that one is deliberate and the comment says so.

**`local_expert_count()` dead code (ed85a34)** — a lost reset, not dead code, and thank you for not letting me hand-wave it. Splitting the monolithic op into a base plus two backends moved `reset()` to the base and dropped its last two statements; the tail of one landed after the `return`. Both `total_recv.zero_()` and `cross_device_flag.fill_(1)` are back, in the base, since both backends allocate both. Half-resetting is the bad state here specifically: `cross_device_flag` is the local half of the cross-device barrier and the `arena.zero()` one line above has just wiped the peer half.

Retested on 8× MI355 (gfx950): `ctest -R '^jit_'` 2/2, `test_jit_binding.py` 18/18 run from `/tmp` (which is what exercises the new schema string end to end), `test_op.py` PASS on both backends, `test_ep_backend_parity.py` 0 failures, `test_dispatch_combine_v2_intranode.py` 18/18.

kawhil-amd:
One question, how can i change the epv2 kernel backend from flydsl to hip, or inverse?

jhchouuu:
Three ways, in precedence order. Nothing else about the call site changes — same class, same `dispatch`/`combine` signatures, same routing handle.

**1. On the config** (recommended):

```python
cfg = EpDispatchCombineConfig(..., kernel_backend="hip")   # or "flydsl"
op = EpDispatchCombineOp(cfg, comm)
assert op.backend_name == "hip"
```

**2. Environment**, when you want to A/B an existing script without touching it:

```bash
MORI_V2_KERNEL_BACKEND=hip torchrun --standalone --nproc_per_node=8 your_script.py
```

**3. Name the subclass**, if you would rather be explicit than configurable:

```python
from mori.ops.dispatch_combine_v2.hip_backend import EpDispatchCombineOpHip
op = EpDispatchCombineOpHip(cfg, comm)     # isinstance(op, EpDispatchCombineOp) still holds
```

(From the submodule, not the package: `mori.ops.dispatch_combine_v2/__init__.py` exports only the
base, the config and the routing handle, precisely so that importing the package pulls in neither
backend.)

`cfg.kernel_backend` wins over the env var, which wins over the default (`flydsl`). Naming the subclass bypasses both.

**One caveat with (3).** The tuned geometry is resolved in `EpDispatchCombineConfig.__post_init__`, i.e. before the op exists, and it reads the same `cfg.kernel_backend or MORI_V2_KERNEL_BACKEND or "flydsl"` chain. So if you construct `EpDispatchCombineOpHip` directly while leaving `cfg.kernel_backend=None` and no env var, you get the HIP kernels running on FlyDSL's tuned schedule — correct results, wrong geometry. Set `kernel_backend="hip"` on the config as well and it lines up. (Wart on my side; I can make the op reconcile the two if you would rather it just did the right thing.)

**Two things worth knowing when you switch:**

- `EpDispatchCombineOp.available_backends()` tells you which ones actually import here. Only the selected backend is imported, so `hip` works on a machine with no FlyDSL installed — and vice versa.
- The HIP backend serves gather combine in bf16/fp32, with fp8/fp4 available on the dispatch leg. Anything else — scatter combine, `quant_type`, StdMoE, per-token scales, routing replay, `local_expert_count` — is rejected **at construction**, before any symmetric memory is taken, with the list of what is missing:

  ```
  ValueError: hip backend cannot serve this config:
    combine_mode=scatter (gather only)
    ...
  ```

  So flipping the switch on a config the HIP backend cannot serve fails immediately and says why, rather than producing wrong numbers at launch.

`tests/python/ops/dispatch_combine_v2/test_ep_backend_parity.py` builds one of each in the same process on the same input if you want a worked example.

jhchouuu:
Two corrections to the comment above, one of which I got wrong.

**The import.** I first wrote `from mori.ops.dispatch_combine_v2 import EpDispatchCombineOpHip`; that raises `ImportError`. The package's `__init__` deliberately exports only `EpDispatchCombineConfig`, `EpDispatchCombineOp` and `EpDispatchRoutingHandle` — eagerly exporting the subclasses would import both backends and defeat the "works without FlyDSL installed" property that is half the point of the seam. The working line, now edited into that comment:

```python
from mori.ops.dispatch_combine_v2.hip_backend import EpDispatchCombineOpHip
```

**The geometry caveat: deferring it to a follow-up PR.** Reasoning, so it is on the record rather than just dropped:

- It costs performance, never correctness. The kernels are chosen by the backend; only the `(block, warp)` schedule comes from the wrong table, and every geometry either table produces is a valid one that `MakeEpCfg` accepts.
- Nothing in-tree can hit it. Every caller here sets `kernel_backend` on the config or uses the env var; direct subclass instantiation appears in no test, bench or example. It needs an out-of-tree caller who picks option 3 *and* leaves `cfg.kernel_backend` unset.
- It is symmetric and the reverse is the likelier accident: instantiating `EpDispatchCombineOpFlyDSL` directly with a stale `MORI_V2_KERNEL_BACKEND=hip` in the environment gets FlyDSL kernels on HIP's schedule.
- The fix is not a one-liner worth smuggling in. The tuned geometry is resolved in `EpDispatchCombineConfig.__post_init__`, before the op exists, so reconciling it means either having the op write back to the config it was handed (making the config mutable after construction) or moving geometry resolution out of `__post_init__` into the op (cleaner, but `cfg.dispatch_block_num` and friends stop being readable before an op is built — an externally visible change). That deserves its own review, not a rider on this one.

Recommendation for anyone switching backends today: set `kernel_backend` on the config. That is option 1, it is the recommended form anyway, and it has none of this.

kawhil-amd (approved):

PR #556, opened 2026-08-13
Logger and timer macros improvements

pemeliya:
## Summary

This PR hardens the logging subsystem (`include/mori/utils/mori_log.hpp`) on two fronts:

1. **Fixes a data race** in `ModuleLogger`'s internal maps that, under the
   multithreaded RDMA workload, could hand back an empty `shared_ptr` that
   callers then cached forever.
2. **Removes `ScopedTimer` overhead on the hot path** so timers compile down to
   a true no-op unless DEBUG logging is actually enabled for the module.

It also adds a dedicated, TSAN-friendly concurrency test and silences a
deprecated-literal warning coming from the bundled `fmt`/`spdlog`.

## Motivation

`ModuleLogger` guarded its `loggers_` / `envOverrides_` maps inconsistently and
relied on a `recursive_mutex`, while several methods called each other and
re-locked. Concurrent first-touch of a module (or resolution of an unknown
module) raced the on-demand insert, occasionally returning a null logger that a
call-site cached permanently. Separately, `ScopedTimer` always copied strings,
read the clock twice, and did a per-call `GetLogger()` hashtable lookup even when
DEBUG was off — measurable on hot paths such as a per-batch `MORI_IO_TRACE`.

### `ModuleLogger` thread-safety
- Eagerly create every known module logger in the singleton constructor, while
  still single-threaded, so `loggers_` is fully populated before any
  `GetLogger()` can touch the map.
- Split the public API from `*Locked` helpers (`GetLoggerLocked`,
  `InitModuleLocked`, `HasEnvOverrideLocked`) so methods no longer re-lock via
  each other.
- Replace the `recursive_mutex` with a plain `std::mutex`.
- `GetLogger()` never returns null: unknown modules are created on demand, and a
  hard failure logs a fatal message and `abort()`s instead of silently dropping
  logs.
- `SetModuleLevel()` now emits the env-override warning via the pre-resolved
  logger while holding the lock (no re-entrant `GetLogger`).

### `ScopedTimer` hot-path optimization
- Resolve and cache the `(module -> logger)` lookup once per call-site in a
  function-local static (`MORI_TIMER` / `MORI_FUNCTION_TIMER`).
- Constructors early-out when the level is disabled: no string copy, no clock
  read, no allocation. The enabled path is marked `MORI_UNLIKELY`.
- Unified name handling under `std::string_view`, copying into an owned string
  only on the enabled path (safe for temporaries).
- Dropped the redundant `enabled_`/`module_` members; `logger_ != nullptr` now
  serves as the "enabled" sentinel.

### Supporting changes
- `include/mori/core/utils/utils.hpp`: add `MORI_LIKELY` / `MORI_UNLIKELY`
  branch-hint macros and scope the `warpSize` macro to device compilation.
- `CMakeLists.txt`: silence `-Wdeprecated-literal-operator` from bundled `fmt`
  for both the `spdlog` build and `mori_logging` consumers (guarded by a
  compiler-flag check).

## Tests
- New `tests/cpp/utils/test_logging.cpp` with multi-threaded stress tests:
  - `ResolveNeverReturnsNull` — concurrent resolution of known + unknown modules.
  - `ConcurrentFirstTouchSameModule` — 64 threads first-touch the same new
    module; all must observe one stable, non-null instance.
  - `LogWhileReconfiguring` — logging while levels are reconfigured concurrently.
- Wired into `tests/cpp/CMakeLists.txt` as `test_logging` (builds independently
  of `BUILD_IO` since core logging is header-only). Intended to be run under
  ThreadSanitizer to prove map access is synchronized.

### DIFF
--- a/include/mori/utils/mori_log.hpp
+++ b/include/mori/utils/mori_log.hpp
@@ -302,19 +302,20 @@ constexpr const char* METRICS = "metrics";
 
 // General MORI logging macros.
 //
-// Hot-path note: MORI_GET_LOGGER() resolves the logger by a string-hashed
+// Hot-path note: GetLogger() resolves the logger by a string-hashed
 // unordered_map lookup. Doing that on every call — even when the level is
-// disabled and spdlog drops the message — shows up on hot paths (e.g. a
-// per-batch MORI_IO_TRACE). We cache the resolved logger pointer in a
-// function-local static so the map lookup happens once per call-site, ever.
-// spdlog's own logger->level(...) already checks the level and skips
-// formatting when disabled, and it reads the level on each call, so runtime
-// level changes are still honored.
-#define MORI_LOG(module, level, ...)                                 \
-  do {                                                               \
-    static ::spdlog::logger* _mori_log_logger =                      \
-        ::mori::ModuleLogger::GetInstance().GetLogger(module).get(); \
-    _mori_log_logger->level(__VA_ARGS__);                            \
+// disabled and spdlog drops the message — shows up on hot paths. 
+// We cache the resolved logger shared_ptr in a  function-local static so the 
+// map lookup happens once per call-site, ever. spdlog's own logger->level(...) 
+// already checks the level and skips formatting when disabled, and it reads the 
+// level on each call, so runtime level changes are still honored. 
+// Matches MORI_TIMER and avoids a dangling pointer if a logger is 
+// spdlog::drop()-ed or torn down during static destruction.
+#define MORI_LOG(module, level, ...)                            \
+  do {                                                          \
+    static const auto _mori_log_logger =                        \
+        ::mori::ModuleLogger::GetInstance().GetLogger(module);  \
+    _mori_log_logger->level(__VA_ARGS__);                       \
   } while (0)
 
 #define MORI_TRACE(module, ...) MORI_LOG(module, trace, __VA_ARGS__)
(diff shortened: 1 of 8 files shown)


pemeliya:
Hi @maning00, you mentioned that includng utils.hpp from mori_log.hpp breaks some device module compilations?

Can you point me to the relevant module? I do not see any compile errors locally. The reason I did it is to have MORI_LIKELY/UNLIKELY macros to be defined in one central place..

include/mori/utils/mori_log.hpp
```diff
@@ -267,8 +300,22 @@ constexpr const char* METRICS = "metrics";
...
+// function-local static so the map lookup happens once per call-site, ever.
+// spdlog's own logger->level(...) already checks the level and skips
+// formatting when disabled, and it reads the level on each call, so runtime
+// level changes are still honored.
+#define MORI_LOG(module, level, ...)                                 \
+  do {                                                               \
+    static ::spdlog::logger* _mori_log_logger =                      \
+        ::mori::ModuleLogger::GetInstance().GetLogger(module).get(); \
+    _mori_log_logger->level(__VA_ARGS__);                            \
+  } while (0)
```
jhchouuu: could you remove the .get()
Like MORI_TIMER(Line 438-442), only get when using.

jhchouuu (approved):
LGTM Thanks, minor suggestion

include/mori/utils/mori_log.hpp
pemeliya: Yes, good catch. I have rewrote that part to cache shared_ptr instead of a raw pointer, same as MORI_TIMER does. I have also cleaned up MORI_TIMER to get a more compact implementation

jhchouuu: please fix pre-commit？

PR #558, opened 2026-08-13
feat(shmem): add a CPU host-proxy RDMA transport alongside IBGDA

itej89:
## Motivation

Enable Expert Parallelism (EP) on clusters where GPU-initiated IBGDA doorbells don't work or the vendor DV library is unavailable.

## Technical Details

Adds TransportType::PROXY = 3 to the existing dispatch framework. EP kernels route through DISPATCH_TRANSPORT_TYPE → <PROXY> template specializations in shmem_proxy_kernels.hpp, leaving shmem_ibgda_kernels.hpp untouched. Proxy code compiles only when -DMORI_PROXY_ENABLED is set through JIT pipeline.

Per-NIC CPU threads poll host-pinned ring buffers (64K slots × 128B, hipHostRegister mapped). GPU warps write ProxyCmd descriptors; CPU threads convert to ibv_post_send with per-NIC lkey/rkey override. Signal delivery uses RDMA_WRITE for data signals and SEND_WITH_IMM + CPU-side __atomic_fetch_add for barrier atomics (ionic/bnxt). CX7 uses native IBV_WR_ATOMIC_FETCH_AND_ADD for barrier atomics.

Supported NIC backends: **Pensando AINIC (ionic)**, **Mellanox CX7 (mlx5)**, **Broadcom Thor2 (bnxt)**.

When `MORI_EP_OVER_RDMA=1`, the DV API availability check is skipped for all providers — proxy uses plain ibverbs and does not require vendor DV libraries.

## How to Run

```bash
export MORI_EP_OVER_RDMA=1
export MORI_RDMA_DEVICES=<nic_list>
export MORI_IB_GID_INDEX=<gid>
export MORI_SHMEM_HEAP_SIZE=4G
export MORI_SOCKET_IFNAME=<ctrl_nic>

torchrun --nnodes=2 --node_rank=<0|1> --nproc_per_node=1 \
  --master_addr=<master_ip> --master_port=1234 \
  examples/ops/dispatch_combine/test_dispatch_combine_internode.py \
  --cmd bench --num-qp 1
```

## Test Results

Commit: `9ae6ebdf` — 2-node EP16 benchmark

### Proxy mode (MORI_EP_OVER_RDMA=1)

| Metric | AINIC (ionic) | CX7 (mlx5) | Broadcom (bnxt) |
|---|---|---|---|
| Dispatch Best (GB/s) | 39.78 | 53.47 | 63.52 |
| Dispatch Worst (GB/s) | 28.57 | 36.13 | 43.46 |
| Combine Best (GB/s) | 69.93 | 71.09 | 82.72 |
| Combine Worst (GB/s) | 40.10 | 44.68 | 52.42 |

### Native IBGDA (no proxy, CX7 only)

| Metric | CX7 (mlx5) |
|---|---|
| Dispatch Best (GB/s) | 70.97 |
| Combine Best (GB/s) | 72.42 |

AINIC native: IBGDA hangs (GPU-written WQEs not visible to NIC on VFIO platform).
Broadcom native: DV library not available on test cluster (RDMA uABI 1).

### Clusters tested

| Cluster | GPU | NIC |
|---|---|---|
| DigitalOcean AINIC | MI350X (gfx950) | Pensando AINIC (ionic) |
| AMD MI300X | MI300X (gfx942) | Mellanox CX7 (mlx5) |
| Ruby (rckg) | MI350X (gfx950) | Broadcom Thor2 (bnxt) |

## Submission Checklist


itej89:
## Verification — commit `32f8129c`

  All 4 clusters pass. 2-node EP16 benchmark, 10 iterations each, zero crashes.

  ### Proxy (`MORI_EP_OVER_RDMA=1`)

  | Cluster | NIC | Dispatch Best (GB/s) | Dispatch Avg (GB/s) | Combine Best (GB/s) | Combine Avg (GB/s)
  |---|---|---|---|---|---|
  | AINIC DigitalOcean | ionic | 35.85 | 30.66 | 58.54 | 54.02 |
  | CX7 MI300X | mlx5 | 45.23 | 39.85 | 55.84 | 54.36 |
  | Broadcom Ruby | bnxt | 54.33 | 53.55 | 69.84 | 68.49 |
  | AINIC Crusoe | ionic | 35.43 | 31.73 | 57.50 | 56.61 |

  ### Native IBGDA (CX7 only)

  | Cluster | Dispatch Best (GB/s) | Dispatch Avg (GB/s) | Combine Best (GB/s) | Combine Avg (GB/s) |
  |---|---|---|---|---|
  | CX7 MI300X | 58.14 | 51.17 | 60.07 | 57.31 |

  Native path verified unaffected by proxy changes.

  AINIC native: IBGDA hangs (GPU-written WQEs not visible to NIC).
  Broadcom native: DV library not available (RDMA uABI 1).
  Crusoe native: IBGDA hangs (same root cause as AINIC DO).

itej89:
# CPU Proxy vs Native IBGDA — EP Benchmark Comparison

## Setup

- **Branch:** `feat/ep-rdma-sharing` @ `32f8129c`
- **Cluster:** CX7 MI300X (gfx942), Mellanox ConnectX-7 (mlx5)
- **Proxy:** `MORI_EP_OVER_RDMA=1` — CPU pthread calls `ibv_post_send` on behalf of GPU
- **Native:** Standard IBGDA — GPU writes WQEs and rings doorbells directly
- **Test:** `test_dispatch_combine_internode.py --cmd bench --max-tokens <N>` at each token count

## EP16 Internode (2 nodes, 8 GPUs each, RDMA)

| Tokens | Proxy Disp (GB/s) | Native Disp (GB/s) | Proxy Comb (GB/s) | Native Comb (GB/s) | Proxy D Lat (µs) | Native D Lat (µs) | Proxy C Lat (µs) | Native C Lat (µs) |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 8 | 2.23 | 2.46 | 1.74 | 1.91 | 102.7 | 93.4 | 131.8 | 119.8 |
| 16 | 4.48 | 4.94 | 3.42 | 3.90 | 102.5 | 92.8 | 134.0 | 117.6 |
| 32 | 8.65 | 8.58 | 6.63 | 7.43 | 106.1 | 107.0 | 136.3 | 121.6 |
| 64 | 15.18 | 16.10 | 13.10 | 14.86 | 120.9 | 113.1 | 139.0 | 122.5 |
| 128 | 26.08 | 25.72 | 25.55 | 28.62 | 140.7 | 142.1 | 143.1 | 128.0 |
| 256 | 47.23 | 47.29 | 41.97 | 44.45 | 155.1 | 154.9 | 174.2 | 164.7 |
| 512 | 51.86 | 56.37 | 49.31 | 46.97 | 282.3 | 259.9 | 297.5 | 311.0 |
| 1024 | 60.14 | 61.61 | 62.40 | 58.63 | 486.8 | 475.2 | 468.2 | 499.3 |
| 2048 | 56.79 | 66.50 | 70.84 | 67.73 | 1029.9 | 880.0 | 825.3 | 864.9 |
| 4096 | 52.01 | 71.86 | 70.58 | 72.31 | 2251.0 | 1629.0 | 1657.0 | 1616.9 |
| 16384 | 61.87 | 73.10 | 71.58 | 72.38 | 7566.6 | 6402.1 | 6539.2 | 6467.7 |
| 65536 | 65.83 | 73.42 | 72.77 | 72.78 | 28442.5 | 25501.1 | 25725.8 | 25724.6 |

itej89:
## DSV3 2P2D EP16, ISL=8192/OSL=1024, MI300X CX7, vLLM

| Concurrency | IBGDA tok/s | CPU Workaround tok/s | tok/s Δ% | IBGDA TTFT (ms) | CPU Workaround TTFT (ms) | TTFT Δ% | IBGDA TPOT (ms) | CPU Workaround TPOT (ms) | TPOT Δ% |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 8 | 41.63 | 39.90 | -4.2% | 4579.65 | 4552.15 | -0.6% | 186.58 | 195.15 | +4.6% |
| 16 | 81.74 | 79.12 | -3.2% | 6358.87 | 6659.37 | +4.7% | 187.02 | 193.28 | +3.3% |
| 64 | 295.40 | 287.64 | -2.6% | 12920.98 | 13495.50 | +4.4% | 192.37 | 199.84 | +3.9% |
| 128 | 550.06 | 543.60 | -1.2% | 20728.86 | 20442.44 | -1.4% | 193.39 | 200.14 | +3.5% |
| 256 | 1010.44 | 971.59 | -3.8% | 33193.89 | 34858.86 | +5.0% | 193.15 | 199.17 | +3.1% |

src/application/transport/rdma/proxy/proxy_thread.cpp
```diff
@@ -0,0 +1,275 @@
...
+  while ((n = ibv_poll_cq(qph.cq, 64, wc)) > 0) {
+    for (int i = 0; i < n; i++) {
+      if (wc[i].opcode == IBV_WC_RECV || wc[i].opcode == IBV_WC_RECV_RDMA_WITH_IMM) {
+        if (wc[i].status == IBV_WC_SUCCESS && wc[i].byte_len >= 16) {
+          uint32_t recv_idx = static_cast<uint32_t>(wc[i].wr_id);
+          if (recv_idx < qph.recv_count && qph.recv_buf) {
+            struct { uint64_t addr; uint64_t val; } payload;
+            memcpy(&payload, reinterpret_cast<char*>(qph.recv_buf) + recv_idx * 64, 16);
+            volatile uint64_t* target = reinterpret_cast<volatile uint64_t*>(payload.addr);
+            __atomic_fetch_add(target, payload.val, __ATOMIC_SEQ_CST);
```
jhchouuu: the emulated atomic dereferences a pointer taken straight off the wire, with no bounds check. Is a bounds check necessary?

src/application/transport/rdma/proxy/proxy_thread.cpp
```diff
@@ -0,0 +1,275 @@
...
+      wr.opcode = IBV_WR_RDMA_WRITE;
+      wr.wr.rdma.remote_addr = cmd->dst_addr;
+      wr.wr.rdma.rkey = (qph.rkey_override != 0) ? qph.rkey_override : cmd->rkey;
+      break;
+    case PROXY_RDMA_WRITE_INLINE:
+      wr.opcode = IBV_WR_RDMA_WRITE;
+      wr.send_flags |= IBV_SEND_INLINE;
+      wr.wr.rdma.remote_addr = cmd->dst_addr;
+      wr.wr.rdma.rkey = (qph.rkey_override != 0) ? qph.rkey_override : cmd->rkey;
+      break;
```
jhchouuu, line 158: Does IBV_SEND_INLINE require specifying the inline data?

src/shmem/init.cpp
```diff
@@ -592,8 +597,66 @@ void GpuStateInit(ShmemStates* states) {
   states->gpuStates.worldSize = states->bootStates->worldSize;
   states->gpuStates.numQpPerPe = states->rdmaStates->commContext->GetNumQpPerPe();
 
-  // Copy communication metadata to GPU
-  CopyTransportTypesToGpu(states);
+  // Check if IBGDA proxy mode is requested
+  if (states->rdmaStates->commContext->IsProxyEnabled()) {
+    fprintf(stderr, "[MORI] Rank %d: CPU proxy RDMA mode enabled (MORI_EP_OVER_RDMA=1)\n", states->bootStates->rank);
+    // Determine number of NICs for per-NIC ring allocation
```
jhchouuu: Use MORI LOG module?

python/mori/jit/core.py
jhchouuu, line 125: pls help add _tunable_defines() to _hipcc_device_bc. 
_hipcc_device_bc omits _tunable_defines() while get_cache_dir keys on it, so the bitcode is built without -DMORI_PROXY_ENABLED. will affects Triton/FlyDSL

src/application/memory/symmetric_memory.cpp
```diff
@@ -195,18 +195,51 @@ SymmMemObjPtr SymmMemManager::RegisterSymmMemObj(void* localPtr, size_t size, bo
...
+            auto mr2 = allCtxs[n]->RegisterRdmaMemoryRegionAuto(localPtr, size);
+            perNicLkeys[n] = mr2.lkey;
+            perNicPeerRkeys[n][rank] = mr2.rkey;
+          }
+          bootNet.Allgather(&perNicPeerRkeys[n][rank], perNicPeerRkeys[n].data(), sizeof(uint32_t));
+        }
+      }
+    } else {
+      cpuMemObj->lkey = heapLkey_;
+      memcpy(cpuMemObj->peerRkeys, heapRkeys_.data(), worldSize * sizeof(uint32_t));
```
jhchouuu: non-heap_begin registrations inherit the heap's keys; in ShmemMode::Isolation heapRkeys_ is empty, so this memcpys from nullptr. 
We default to using the static heap. This issue might not be exposed, so it's recommended to add a protection mechanism.

src/application/memory/symmetric_memory.cpp
```diff
@@ -195,18 +195,51 @@ SymmMemObjPtr SymmMemManager::RegisterSymmMemObj(void* localPtr, size_t size, bo
...
-  // buffer is never an RDMA src/dst). This dodges the ionic single-MR limit
-  // (ibv_reg_mr fails at >=~2 GiB) for the hierarchical AllGather's intra
-  // node-block. The rkey stays 0 and the Allgather below still runs, so the
-  // collective register stays in lockstep.
-  if (rdmaDeviceContext && anyRdmaPeer && rdmaRegister) {
-    application::RdmaMemoryRegion mr =
-        rdmaDeviceContext->RegisterRdmaMemoryRegionAuto(localPtr, size);
-    cpuMemObj->lkey = mr.lkey;
-    cpuMemObj->peerRkeys[rank] = mr.rkey;
+  if (context.IsProxyEnabled() && rdmaDeviceContext && anyRdmaPeer) {
```
jhchouuu: the proxy branch drops the rdmaRegister guard

src/application/transport/rdma/proxy/proxy_thread.cpp
```diff
@@ -0,0 +1,275 @@
...
+}
+
+void ProxyThread::DrainCq(ProxyQpHandle& qph) {
+  if (!qph.cq) return;
+  ibv_wc wc[64];
+  int n;
+  while ((n = ibv_poll_cq(qph.cq, 64, wc)) > 0) {
+    for (int i = 0; i < n; i++) {
+      if (wc[i].opcode == IBV_WC_RECV || wc[i].opcode == IBV_WC_RECV_RDMA_WITH_IMM) {
+        if (wc[i].status == IBV_WC_SUCCESS && wc[i].byte_len >= 16) {
```
jhchouuu: Should CQE check wc.status first?

include/mori/core/transport/rdma/proxy/proxy_device_primitives.hpp
```diff
@@ -0,0 +1,186 @@
...
+  ring->cmds[slot].rkey = rkey;
+  ring->cmds[slot].atomic_arg = add_value;
+  ring->cmds[slot].flags = 1;
+  ring->cmds[slot].result = 0;
+
+  __threadfence_system();
+  ring->cmds[slot].status = PROXY_PENDING;
+
+  ProxyWaitSlotCompleted(ring, slot);
+  return ring->cmds[slot].result;
```
jhchouuu, line 168: Do atomic fetch operations always return 0? Is this the expected?

src/application/transport/rdma/proxy/proxy_thread.cpp
```diff
@@ -0,0 +1,275 @@
...
+        sge.length = 8;
+        sge.lkey = cmd->lkey;
+      } else {
+        ibuf.data[0] = cmd->dst_addr;
+        ibuf.data[1] = cmd->atomic_arg;
+        sge.addr = reinterpret_cast<uintptr_t>(&ibuf.data[0]);
+        sge.length = 16;
+        wr.opcode = IBV_WR_SEND_WITH_IMM;
+        wr.imm_data = htonl(0xA70C);
+        wr.send_flags |= IBV_SEND_INLINE;
```
jhchouuu, line 187: Perhaps it is the current workaround for the DO and crusoe clusters. Once we have a clear understanding of the  DO and crusoe clusters’ issues, we should restore the nature atomic function....

src/application/transport/rdma/providers/bnxt/bnxt.cpp
```diff
@@ -601,6 +648,62 @@ RdmaEndpoint BnxtDeviceContext::CreateRdmaEndpoint(const RdmaEndpointConfig& con
...
+      a.ah_attr.sl = ReadRdmaServiceLevelEnv().value_or(0);
+      std::optional<uint8_t> tc = ReadRdmaTrafficClassEnv();
+      if (tc.has_value()) a.ah_attr.grh.traffic_class = tc.value();
+      ibv_modify_qp(plainQp, &a, IBV_QP_STATE | IBV_QP_PATH_MTU | IBV_QP_DEST_QPN |
+          IBV_QP_RQ_PSN | IBV_QP_AV | IBV_QP_MAX_DEST_RD_ATOMIC | IBV_QP_MIN_RNR_TIMER); }
+
+    { ibv_qp_attr a{}; a.qp_state = IBV_QPS_RTS; a.sq_psn = local.psn;
+      a.timeout = 14; a.retry_cnt = 7; a.rnr_retry = 7; a.max_rd_atomic = 15;
+      ibv_modify_qp(plainQp, &a, IBV_QP_STATE | IBV_QP_SQ_PSN | IBV_QP_TIMEOUT |
+          IBV_QP_RETRY_CNT | IBV_QP_RNR_RETRY | IBV_QP_MAX_QP_RD_ATOMIC); }
```
jhchouuu: can we check the return value of all ibv_modify_qp?

src/application/transport/rdma/providers/ionic/ionic.cpp
jhchouuu, line 1: Architecture: the proxy path should not touch the DV providers at all.

The proxy transport is pure host-side ibverbs, and mori already has an IBVerbs backend for exactly that — RdmaBackendType::IBVerbs, IBVerbsDevice, IBVerbsDeviceContext, all vendor-agnostic and already used by topology/net.cpp and the io/rdma stack. The DV providers exist to build the IBGDA path (parent domains, CMB, UDMA masks, CCQE); none of that is used in proxy mode.

The ~333 lines added across ionic.cpp/mlx5.cpp/bnxt.cpp reimplement IBVerbsDeviceContext::ConnectEndpoint, and the existing one is strictly better: it checks every ibv_modify_qp via ModifyOrThrow, queries max_qp_rd_atom instead of hardcoding 15, uses hop_limit = 16, honours MORI_IB_PATH_MTU, handles the InfiniBand link layer (the new code assumes RoCE unconditionally), and has a working DestroyRdmaEndpointNoThrow.

I think this collapses to one line at context.cpp:195:


rdmaContext.reset(new RdmaContext(
    proxyEnabled ? RdmaBackendType::IBVerbs : RdmaBackendType::DirectVerbs));
That also removes the three && !env::IsEnvVarEnabled("MORI_EP_OVER_RDMA") edits in rdma.cpp (unnecessary once the factory returns before the vendor switch), the dynamic_cast chain in context.cpp:443-462, and the new members in the three provider headers — a net reduction in the feature's footprint on existing code, which is the stated goal.

Two pieces would need a home: atomicIbuf setup (~15 lines in IBVerbsDeviceContext::CreateRdmaEndpoint, gated on config.atomicIbufSlots), and the 512 receive WRs — which arguably belong in ProxyThread::Init rather than any provider, since they exist only because the proxy chose SEND_WITH_IMM for atomics. Putting them there also fixes the RNR exposure, since replenishment and consumption would finally live in the same object.

src/application/context/context.cpp
```diff
@@ -51,6 +55,7 @@ Context::Context(BootstrapNetwork& bootNet) : bootNet(bootNet) {
   // uncached SDMA buffers, leading to cache/IPC inconsistency hangs.
   sdmaEnabled = env::IsEnvVarEnabled("MORI_ENABLE_SDMA");
   p2pDisabled = env::IsEnvVarEnabled("MORI_DISABLE_P2P");
+  proxyEnabled = env::IsEnvVarEnabled("MORI_EP_OVER_RDMA");
```
jhchouuu: two suggestions:
1. MORI_EP_OVER_RDMA rename to MORI_ENABLE_HOST_PROXY
2. read it once in Context via IsEnvVarEnabled and route everything else through proxyEnabled?

jhchouuu (commented):
Thanks for the PR, this looks like a complete cpu proxy design, so I went through it carefully.

I also validated it on our Crusoe cluster. With the proxy enabled it works — v1 and v1_ll both pass. async_ll does not work under this scheme, though; it faults reproducibly.

I've left a number of comments below — could you take a look? Happy to discuss, Thanks!

src/application/context/context.cpp
itej89: Resolved by
1. Changing the toggle name to "MORI_ENABLE_HOST_PROXY".  
2. Reusing the proxyEnabled from Context everywhere

src/application/transport/rdma/providers/ionic/ionic.cpp
itej89: Resolved by
1. Reused the Mori's IBVerbs backend 
2. Updated with the factory design pattern and removed the dynamic caste chain

src/application/transport/rdma/providers/bnxt/bnxt.cpp
itej89: Handled by the Mori's IBVerbs backend

src/application/transport/rdma/proxy/proxy_thread.cpp
itej89: I agree. We can remove this once the native atomics are confirmed working.

src/application/transport/rdma/proxy/proxy_thread.cpp
itej89: Data is already set outside the switch statement. There are no special fields for inline data.  I think, the flag alone decides the transport type. 

`sge.addr = cmd->src_addr;
  sge.length = cmd->length;`

src/shmem/init.cpp
itej89: Updated to "_MORI_SHMEM_INFO_" api

python/mori/jit/core.py
itej89: added "__tunable_defines()_"

src/application/transport/rdma/proxy/proxy_thread.cpp
itej89: Added the check and error log.

src/application/memory/symmetric_memory.cpp
itej89: Added the null pointer check.

include/mori/core/transport/rdma/proxy/proxy_device_primitives.hpp
itej89: resolved it by writing the value -
https://github.com/itej89/mori/blob/de983fb2067291715660a1b9f48e372abcec924f/src/application/transport/rdma/proxy/proxy_thread.cpp#L111

**Problem:**
The "_IBV_WR_SEND_WITH_IMM_" atomics workaround doesn't read the count from the remote CPU - this would require full network round trip per atomic. Enabling this can have performance impact.   

As of now "_ProxyPostAtomicNonFetch_" is the only caller.  For "_ProxyPostAtomicFetchAdd_" to work, we would need native-atomics to be working.

If this is for future proofing, I can look into this in the next PR.

src/application/memory/symmetric_memory.cpp
itej89: Included the guard.

jhchouuu (commented):
Hi @itej89, Thanks for the refactor!
Could you fix the conflict before merge?
And could you have a test on crusoe cluster again？

And three of my earlier comments look only partly addressed — following up on each rather than re-opening the threads:

---

**1. `proxy_thread.cpp:58` — bounds check on the wire pointer**

The `payload.addr == 0` guard helps, but a null check isn't the property I was asking about. A native `IBV_WR_ATOMIC_FETCH_AND_ADD` has the responder NIC validate the target address against the MR named by the `rkey`. A `SEND` carries no such constraint, so the emulation drops that validation entirely and the CPU does an 8-byte add wherever the payload points. Any wire corruption or a heap-layout mismatch between ranks becomes an arbitrary write into the local process.

`GpuStates` already tracks `heapBaseAddr` / `heapEndAddr` (`internal.hpp:128-129`); passing those into `ProxyThread::Init` and rejecting out-of-range payloads would restore the check cheaply. Was there a reason not to?

**2. `proxy_thread.cpp:140-145` — `IBV_SEND_INLINE` inline data**

This one looks unchanged. `PROXY_RDMA_WRITE_INLINE` sets `IBV_SEND_INLINE` but leaves `sge.addr` as the raw `cmd->src_addr` assigned at line 122, which is a GPU address — so libibverbs memcpys the payload from a GPU pointer in the proxy thread's host context.

The answer is in the next case down: `PROXY_SIGNAL_WRITE` (`:146-155`) correctly redirects `sge.addr` into the host-side `InlineBuf` before setting the flag. `PROXY_RDMA_WRITE_INLINE` needs the same treatment. Its producer is `shmem_proxy_kernels.hpp:151`.

**3. `proxy_device_primitives.hpp:158` — fetch atomics returning 0**

The fix in `DrainCq` is gated on `qph.use_native_atomics`:

```c
if (ring_->cmds[slot].op == PROXY_ATOMIC_FETCH_ADD && qph.use_native_atomics) {
  ring_->cmds[slot].result = *reinterpret_cast<volatile uint64_t*>(ring_->cmds[slot].src_addr);
}
```

and `use_native_atomics` is `vendorId == Mellanox` (`init.cpp:784`). So ionic and bnxt — the two NICs this PR is aimed at — still take the `SEND_WITH_IMM` path and still get 0 back. The emulated path would need the responder to send the pre-add value back, or `ProxyPostAtomicFetch` should fail loudly rather than return a wrong value silently.

---

**4. New, and I think it should block: `proxy_thread.cpp:96` breaks non-x86 builds even with the feature off**

```c
asm volatile("clflush (%0)" :: "r"(target) : "memory");
asm volatile("sfence" ::: "memory");
```

There's no `#if defined(__x86_64__)`, and `CMakeLists.txt:66` adds `proxy_thread.cpp` to the source list unconditionally — so an aarch64 build fails whether or not `MORI_ENABLE_HOST_PROXY` is set. Everything else in this PR is cleanly opt-in (`git diff -w` shows zero removed lines in `init.cpp`, and `symmetric_memory.cpp`'s native branch is byte-identical to main), so this is the one place the feature reaches past its own switch.

Guarding the asm is enough; `__builtin_ia32_clflush` plus `std::atomic_thread_fence` would also work. Worth noting the `clflush` is a no-op for its stated purpose anyway if the target is uncached device memory.

itej89:
> Hi @itej89, Thanks for the refactor! Could you fix the conflict before merge? And could you have a test on crusoe cluster again？
> 
> And three of my earlier comments look only partly addressed — following up on each rather than re-opening the threads:
> 
> **1. `proxy_thread.cpp:58` — bounds check on the wire pointer**
> 
> The `payload.addr == 0` guard helps, but a null check isn't the property I was asking about. A native `IBV_WR_ATOMIC_FETCH_AND_ADD` has the responder NIC validate the target address against the MR named by the `rkey`. A `SEND` carries no such constraint, so the emulation drops that validation entirely and the CPU does an 8-byte add wherever the payload points. Any wire corruption or a heap-layout mismatch between ranks becomes an arbitrary write into the local process.
> 
> `GpuStates` already tracks `heapBaseAddr` / `heapEndAddr` (`internal.hpp:128-129`); passing those into `ProxyThread::Init` and rejecting out-of-range payloads would restore the check cheaply. Was there a reason not to?
> 
> **2. `proxy_thread.cpp:140-145` — `IBV_SEND_INLINE` inline data**
> 
> This one looks unchanged. `PROXY_RDMA_WRITE_INLINE` sets `IBV_SEND_INLINE` but leaves `sge.addr` as the raw `cmd->src_addr` assigned at line 122, which is a GPU address — so libibverbs memcpys the payload from a GPU pointer in the proxy thread's host context.
> 
> The answer is in the next case down: `PROXY_SIGNAL_WRITE` (`:146-155`) correctly redirects `sge.addr` into the host-side `InlineBuf` before setting the flag. `PROXY_RDMA_WRITE_INLINE` needs the same treatment. Its producer is `shmem_proxy_kernels.hpp:151`.
> 
> **3. `proxy_device_primitives.hpp:158` — fetch atomics returning 0**
> 
> The fix in `DrainCq` is gated on `qph.use_native_atomics`:
> 
> ```c
> if (ring_->cmds[slot].op == PROXY_ATOMIC_FETCH_ADD && qph.use_native_atomics) {
>   ring_->cmds[slot].result = *reinterpret_cast<volatile uint64_t*>(ring_->cmds[slot].src_addr);
> }
> ```
> 
> and `use_native_atomics` is `vendorId == Mellanox` (`init.cpp:784`). So ionic and bnxt — the two NICs this PR is aimed at — still take the `SEND_WITH_IMM` path and still get 0 back. The emulated path would need the responder to send the pre-add value back, or `ProxyPostAtomicFetch` should fail loudly rather than return a wrong value silently.
> 
> **4. New, and I think it should block: `proxy_thread.cpp:96` breaks non-x86 builds even with the feature off**
> 
> ```c
> asm volatile("clflush (%0)" :: "r"(target) : "memory");
> asm volatile("sfence" ::: "memory");
> ```
> 
> There's no `#if defined(__x86_64__)`, and `CMakeLists.txt:66` adds `proxy_thread.cpp` to the source list unconditionally — so an aarch64 build fails whether or not `MORI_ENABLE_HOST_PROXY` is set. Everything else in this PR is cleanly opt-in (`git diff -w` shows zero removed lines in `init.cpp`, and `symmetric_memory.cpp`'s native branch is byte-identical to main), so this is the one place the feature reaches past its own switch.
> 
> Guarding the asm is enough; `__builtin_ia32_clflush` plus `std::atomic_thread_fence` would also work. Worth noting the `clflush` is a no-op for its stated purpose anyway if the target is uncached device memory.

Hi @jhchouuu, thank you for the feedback. I've made the requested changes as follows:
1. Fixed - heap range passed to ProxyThread::Init, payload.addr validated against [heapBaseAddr, heapEndAddr) before dereference.

2. Fixed - inline_data[48] field added to ProxyCmd. GPU stages data there, CPU reads from it. Tagged with ProxyInlineTag

3. Enabled - Implemented the  full-round trip.  ender sends PROXY_IMM_ATOMIC_FETCH, receiver does atomic and replies with PROXY_IMM_ATOMIC_REPLY carrying the old value. ~18% combine overhead vs fire-and-forget when forced on all atomics; zero overhead on normal path (ProxyPostAtomicNonFetch stays fire-and-forget)

4. Fixed - replaced with std::atomic_thread_fence(std::memory_order_seq_cst)

itej89:
cc: @jhchouuu 

# EP-Over-HOST-PROXY Verification Report - PR(1781fa1a) vs main

**Branch:** `feat/ep-rdma-sharing`  
**Date:** 2026-08-29  
**Test:** `test_dispatch_combine_internode.py --cmd bench` (2-node, 8 GPU/node, 16 ranks total)

## Results

### Dispatch Performance (bfloat16)

| Cluster | Commit | Mode | Best (GB/s) | Worst (GB/s) | Avg (GB/s) | Best Latency (us) | Worst Latency (us) | Avg Latency (us) |
|---|---|---|---|---|---|---|---|---|
| CX7 MI300X | `7c51d18f` (main) | Native (IBGDA) | 67.87 | 9.44 | 48.69 | 1724 | 12393 | 3358 |
| CX7 MI300X | `1781fa1a` (PR #558) | Native (IBGDA) | 71.08 | 1.36 | 37.57 | 1647 | 86046 | 17413 |
| CX7 MI300X | `7c51d18f` (main) | Rail-only | 70.77 | 1.47 | 45.26 | 1654 | 79739 | 10210 |
| CX7 MI300X | `1781fa1a` (PR #558) | Rail-only | 70.98 | 1.26 | 45.92 | 1649 | 93179 | 9555 |
| CX7 MI300X | `1781fa1a` (PR #558) | Proxy | 56.23 | 36.13 | 44.46 | 2082 | 3238 | 2659 |
| DO AINIC | `1781fa1a` (PR #558) | Proxy | 50.95 | 32.27 | 40.63 | 2296 | 3623 | 2938 |
| Crusoe AINIC | `1781fa1a` (PR #558) | Proxy | 40.52 | 26.24 | 31.81 | 2887 | 4455 | 3753 |

### Combine Performance (bfloat16)

| Cluster | Commit | Mode | Best (GB/s) | Worst (GB/s) | Avg (GB/s) | Best Latency (us) | Worst Latency (us) | Avg Latency (us) |
|---|---|---|---|---|---|---|---|---|
| CX7 MI300X | `7c51d18f` (main) | Native (IBGDA) | 71.86 | 9.60 | 55.60 | 1627 | 12182 | 2890 |
| CX7 MI300X | `1781fa1a` (PR #558) | Native (IBGDA) | 72.45 | 1.37 | 50.48 | 1615 | 85609 | 12816 |
| CX7 MI300X | `7c51d18f` (main) | Rail-only | 72.38 | 1.47 | 52.38 | 1616 | 79530 | 8249 |
| CX7 MI300X | `1781fa1a` (PR #558) | Rail-only | 72.24 | 1.58 | 51.69 | 1619 | 74113 | 7128 |
| CX7 MI300X | `1781fa1a` (PR #558) | Proxy | 71.53 | 38.41 | 55.88 | 1636 | 3046 | 2129 |
| DO AINIC | `1781fa1a` (PR #558) | Proxy | 75.71 | 39.66 | 54.35 | 1544 | 2949 | 2236 |
| Crusoe AINIC | `1781fa1a` (PR #558) | Proxy | 73.80 | 37.29 | 54.43 | 1586 | 3137 | 2263 |

## Analysis

- **Best-case throughput is identical** across main and our branch (67-72 GB/s dispatch, 71-76 GB/s combine) — no regression in steady-state performance.
- **Average differences** are driven by JIT cold-start outliers (worst-case 12-93 ms) that vary per run. Rail-only averages are within noise between main and ours.
- **Proxy mode has the most consistent latency** — worst-case is only 3-4.5 ms vs 12-93 ms for Native/Rail-only, because the proxy path doesn't trigger GPU-side JIT recompilation on the IBGDA control path.

## Cluster Details

| Cluster | GPU | NIC | Firmware | Nodes |
|---|---|---|---|---|
| DO AINIC | MI350X (gfx950) | AMD Pensando AINIC |  1.117.5-a-56 | fabric-2, fabric-5 |
| CX7 MI300X | MI300X (gfx942) | Mellanox CX7 | mlx5 | useocpm2m-097-026, useocpm2m-097-032 |
| Crusoe AINIC | MI355X (gfx950) | AMD Pensando AINIC | 1.117.5-a-77 | crsuse2-m2m-v2-027, crsuse2-m2m-v2-030 |

src/application/transport/rdma/proxy/proxy_thread.cpp
```diff
@@ -0,0 +1,405 @@
...
+              reply_buf.data[0] = old_val;
+              reply_buf.data[1] = req.reply_slot;
+              ibv_sge rsge{};
+              rsge.addr = reinterpret_cast<uintptr_t>(&reply_buf.data[0]);
+              rsge.length = 16;
+              ibv_send_wr rwr{}, *rbad = nullptr;
+              rwr.opcode = IBV_WR_SEND_WITH_IMM;
+              rwr.imm_data = htonl(PROXY_IMM_ATOMIC_REPLY);
+              rwr.send_flags = IBV_SEND_SIGNALED | IBV_SEND_INLINE;
+              rwr.sg_list = &rsge;
```
jhchouuu, line 172: @itej89 Would you please take a look at this items?

**Blocking: the atomic-fetch reply WR has no `wr_id`, so its completion corrupts ring slot 0**

```c
160:  ibv_send_wr rwr{}, *rbad = nullptr;                    // wr_id zero-initialised
163:  rwr.send_flags = IBV_SEND_SIGNALED | IBV_SEND_INLINE;  // always produces a send CQE
166:  ibv_post_send(qph.qp, &rwr, &rbad);                    // wr_id never assigned
```

`send_cq == recv_cq` (`ibverbs.cpp:213-214`), so this completion lands in the CQ `DrainCq` polls. Opcode is `IBV_WC_SEND`, so it skips the RECV branch and hits the tail:

```c
185:  uint32_t slot = static_cast<uint32_t>(wc[i].wr_id) & PROXY_RING_MASK;  // = 0
195:  ring_->cmds[slot].status = PROXY_COMPLETED;
```

So every atomic-fetch reply marks **slot 0 of the responder's own ring** as completed. Worst case isn't a lost completion — if slot 0 is `PENDING` but not yet consumed, MainLoop's `if (cmd->status != PROXY_PENDING) break;` fires without advancing `next_slot_`, so that ring **stops posting permanently** while the GPU's `ProxyQuiet` sees `COMPLETED` and moves on. Same route at `:106` would mark slot 0 `PROXY_ERROR`.

**Fix** — high-bit sentinel, tested *before* the mask (`~0ull & PROXY_RING_MASK` is 65535, a valid slot):

```c
static constexpr uint64_t PROXY_WRID_INTERNAL = 1ull << 63;   // proxy_types.hpp
rwr.wr_id = PROXY_WRID_INTERNAL;                              // at :160
if (wc[i].wr_id & PROXY_WRID_INTERNAL) continue;              // before :106 and :185
```
**Testing:** no EP operator uses fetch atomics (dispatch/combine use `ShmemAtomicTypeNonFetchThread` / `ShmemPutMemNbiSignalThread`), so the benchmarks can't have covered this path.

jhchouuu (approved):
LGTM overall — thanks for working through all of these. Just the one last item above to sort out.

One other thing: `async_ll` still doesn't seem to work under the proxy — it faults reproducibly right after the first dispatch. `v1` and `v1_ll` are both fine. Is that expected / out of scope for this PR, or worth a look before merging?

src/application/transport/rdma/proxy/proxy_thread.cpp
itej89: Hi @jhchouuu , thanks for noticing the bug, I've updated the PR.

1. Fixed atomic-fetch reply 

2. Regarding async_ll,  it passess for me all three clusters but the performs numbers appears to be wrong. Please confirm if this is the right way to test it.
 

## async_ll vs v1 Kernel Comparison — PR #558 (`1781fa1a`)

**Branch:** `feat/ep-rdma-sharing`  
**Commit:** `1781fa1a`  
**Date:** 2026-08-31  
**Test:** `test_dispatch_combine_internode.py --cmd bench` (2-node, 8 GPU/node, 16 ranks, default 4096 max tokens)

**Test commands:**

v1:
```bash
torchrun --nnodes=2 --node_rank=<0|1> --nproc_per_node=1 \
    --master_addr=<MASTER_IP> --master_port=1234 \
    examples/ops/dispatch_combine/test_dispatch_combine_internode.py \
    --cmd bench --kernel-type v1
```

async_ll:
```bash
torchrun --nnodes=2 --node_rank=<0|1> --nproc_per_node=1 \
    --master_addr=<MASTER_IP> --master_port=1235 \
    examples/ops/dispatch_combine/test_dispatch_combine_internode.py \
    --cmd bench --kernel-type async_ll
```

Proxy enabled via `MORI_ENABLE_HOST_PROXY=1` (build-time and runtime). Native (IBGDA): no `MORI_ENABLE_HOST_PROXY` set.

## Results

### v1 Kernel (InterNodeV1)

| Cluster | Mode | Dispatch Avg (GB/s) | Dispatch Avg Lat (us) | Combine Avg (GB/s) | Combine Avg Lat (us) |
|---|---|---|---|---|---|
| CX7 MI300X | Native (IBGDA) | 41.92 | 13547 | 52.18 | 7725 |
| CX7 MI300X | Proxy | 44.49 | 2649 | 55.86 | 2116 |
| DO AINIC | Proxy | 32.19 | 3706 | 52.20 | 2355 |
| Crusoe AINIC | Proxy | 31.81 | 3753 | 54.43 | 2263 |

### async_ll Kernel (AsyncLL)

| Cluster | Mode | Dispatch Avg (GB/s) | Dispatch Avg Lat (us) | Combine Avg (GB/s) | Combine Avg Lat (us) |
|---|---|---|---|---|---|
| CX7 MI300X | Native (IBGDA) | 0.28 | 420029 | 0.27 | 427291 |
| CX7 MI300X | Proxy | 0.28 | 420181 | 0.27 | 428093 |
| DO AINIC | Proxy | 0.24 | 487724 | 0.26 | 457308 |
| Crusoe AINIC | Proxy | — (cluster full) | — | — | — |

PR #577, opened 2026-08-18
perf(EPv2): rework the gfx1250 combine entry barrier and widen its grid

kawhil-amd:
## Results
ct=512, EP4, hidden 7168, topk 6, bf16, eager, in-place, `bench_ep.py`:
| combine blocks | baseline | this PR |
|---:|---:|---:|
| 64 | **37.2 µs** | 34.1 µs |
| 128 | 38.7 µs | 32.8 µs |
| 256 | 43.7 µs | **30.6 µs** |

Tuned pick to tuned pick: **37.2 → 30.6 µs (−17.7%), 663 → 807 GB/s**.

### DIFF
--- a/include/mori/ops/dispatch_combine_v2/ep_cfg.hpp
+++ b/include/mori/ops/dispatch_combine_v2/ep_cfg.hpp
@@ -309,6 +309,10 @@ constexpr int EpTokenBytes(const EpCfg& c) { return c.hiddenDim * EpElemSize(c.d
 // engine; combine reserves the whole budget and sizes its tiles at runtime.
 // EpCombine1250xLdsBudget must match MORI_COMB_LDS_BUDGET in ep_intranode_1250x.hpp.
 constexpr int EpCombine1250xLdsBudget = 327680;
+// xdbFlag slots for the per-block combine entry barrier: one uint64 per block, so a
+// block is the only writer of its own epoch. 256 == the CU count, which caps the
+// combine block_num; the host allocates this many and every call keeps them in step.
+constexpr int EpXdbFlagSlots = 256;
 constexpr int EpDispatch1250xLdsBytes(const EpCfg& c) {
   return c.warpPerBlock * c.hiddenDim * EpElemSize(c.dtype);
 }
--- a/python/mori/ops/dispatch_combine_v2/hip_backend.py
+++ b/python/mori/ops/dispatch_combine_v2/hip_backend.py
@@ -67,6 +67,8 @@ _DISPATCH_DTYPES = {
     torch.float4_e2m1fn_x2: 1,  # nominal: cfg.token_nbytes is what sizes buffers
 }
 _COMBINE_DTYPES = {torch.bfloat16: 2, torch.float32: 4}
+# Must match EpXdbFlagSlots in include/mori/ops/dispatch_combine_v2/ep_cfg.hpp.
+_XDB_FLAG_SLOTS = 256
 
 
 class EpDispatchCombineOpHip(EpDispatchCombineOp, backend="hip"):
@@ -110,8 +112,12 @@ class EpDispatchCombineOpHip(EpDispatchCombineOp, backend="hip"):
         self.dispatch_barrier = torch.zeros(1, dtype=torch.uint32, device=dev)
         self.combine_barrier = torch.zeros(1, dtype=torch.uint32, device=dev)
         # Monotone epoch. Starts at 1 so the zeroed barrier slots cannot alias
-        # the first launch's flag value.
-        self.cross_device_flag = torch.ones(1, dtype=torch.int64, device=dev)
+        # the first launch's flag value. The gfx1250 combine barrier gives every
+        # block a private slot (EpXdbFlagSlots in ep_cfg.hpp), so it is the only
+        # writer of its own epoch; the portable path only ever uses slot 0.
+        self.cross_device_flag = torch.ones(
+            _XDB_FLAG_SLOTS if self._is1250 else 1, dtype=torch.int64, device=dev
+        )
         self.combine_out = torch.zeros(
             max_tok * cfg.hidden_dim, dtype=cfg.combine_dtype, device=dev
         )
@@ -124,6 +130,11 @@ class EpDispatchCombineOpHip(EpDispatchCombineOp, backend="hip"):
         self.combine_barrier_fan = None
         if self._is1250:
             max_comb_blocks = max(b for b, _ in self._combine_specs)
+            if max_comb_blocks > _XDB_FLAG_SLOTS:
+                raise ValueError(
+                    f"combine block_num {max_comb_blocks} exceeds the {_XDB_FLAG_SLOTS} "
+                    "per-block xdb epoch slots the entry barrier owns"
+                )
             self.combine_barrier_fan = torch.zeros(max_comb_blocks * 16, **i32)
 
     # -- backend hooks -----------------------------------------------------
--- a/python/mori/ops/dispatch_combine_v2/hip_tuning_configs.py
+++ b/python/mori/ops/dispatch_combine_v2/hip_tuning_configs.py
@@ -158,7 +158,12 @@ _COMBINE_TABLE: dict = {
         #   4096  155.5   132.6  133.5
         # The tail is a tie, so it keeps the smaller grid. topk 8 is left alone: topk
         # sets the tokens consumed per round, so its edges need their own sweep.
-        (4, 7168, 6, None): ((128, 64, 8), (256, 128, 8), (512, 256, 8), (None, 128, 8)),
+        (4, 7168, 6, None): (
+            (128, 64, 8),
+            (256, 128, 8),
+            (512, 256, 8),
+            (None, 128, 8),
+        ),
     },
 }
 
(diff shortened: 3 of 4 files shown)


jhchouuu:
Read through the barrier rework carefully. The direction and the implementation both look right to me, and the reasoning in the commit message is unusually good. One blocking item, two things I'd want before merge, and some nits.

## Blocking

**`ep_intranode_1250x.hpp:743-762` has two comment blocks that contradict each other, with no code between them.**

The first (743-754) argues the acquire is required on the staged path — *"Dropping this fails the staged half of the identity check at every token count"* — and goes on to justify device scope and per-block issue. The second (755-762) says it was removed, verified, and that the earlier failures were really the missing staging release.

Both land in the merged file, and the fence they describe no longer exists, so the first block is a leftover from an earlier iteration that now asserts the opposite of what the code does. In this particular function that is worse than no comment at all. Please drop one.

## Before merge

**Graph capture isn't in the verification list.** The barrier is more stateful across launches than it was: a private per-block epoch, plus block 0 fixing up `[gridDim.x, EpXdbFlagSlots)` so a later, wider grid still reads an in-step value. `tests/python/ops/dispatch_combine_v2/test_graph_capture.py` is exactly the test for that, and a serving stack replays this captured. The geometry-switching run you describe (ct=128/256/512/2048) is the right idea and does target the fix-up — worth doing the captured version too.

**The two result tables disagree, and the protocol is thin for the size of the claim.** PR body says 64 blocks `37.2 → 34.1` and 128 blocks `38.7 → 32.8`; the commit message says `37.2 → 35.0` and `38.7 → 33.8` (256 agrees at 30.6). Worth saying which run is which.

More importantly: `min of 2-3 runs`, eager only. `bench_ep.py` already reports means over `ITERS` and supports `MODES=graph`, so means over a few hundred iterations in both modes would cost little and carry more weight for a ±1-2 us difference.

One trap specific to this harness, which I hit this week measuring the same op: **the number you get at a fixed `ct` depends on the `SWEEP` list you asked for.** I have a reproducible case where `SWEEP=1024,2048,4096` reads 12% slower at ct=2048 than `SWEEP=2048` does — same binary, same routing, ITERS=400, repeated. I ran a position x M factorial and only that one cell is anomalous; I haven't found the mechanism yet. Practical consequence: please state the `SWEEP` used, and make sure numbers compared against each other came from the same one.

## Nits

**The fence removals now carry an implicit dependency on the window being uncached.** Both the dropped acquire and the in-place release rest on `CcoWindowAllocType()` returning uncached. That is the default and it holds in every configuration we actually ship — the two compile-time fallbacks sit below `HIP_VERSION 70051831`, which cannot build gfx1250 anyway — so this is not a correctness objection to the PR. But `CCO_UNCACHED_WINDOW=0` is a runtime switch that nothing in the repo sets, and flipping it on gfx1250 would now give silently wrong MoE output rather than an error. A line in the comment at :751 saying that env var is incompatible with this kernel, or a construction-time check, would close it for free.

**`_XDB_FLAG_SLOTS = 256` in `hip_backend.py` is the third hand-synced constant pair in this family** (`EpCombine1250xLdsBudget` / `MORI_COMB_LDS_BUDGET` is the existing one). The `raise` guarding it is good; deriving it through the JIT config would be better.

**The PR body's Motivation / Technical Details / Test Plan are empty template sections** while the commit message covers all of them well — please move it across, since the commit message is where all the reasoning currently lives.

`_qStore`'s comment change (output-width vs tile-width under fp8) is unrelated to this PR; harmless, just noting it.

## Checked and happy with

- **Epoch alignment holds** across a grid that grows or shrinks: block 0 writes `[gridDim.x, 256)`, every other block writes only its own slot, indices never overlap, and the peer lap is bounded at one.
- **The `>=` / `==` split is right** and for the reasons given — 64-bit arena slots a peer can lap, 32-bit truncated fan lines only our own block 0 writes.
- **Single-thread `*totalRecvTokenNum = 0` is safe**: it is only consumed by the staging loop at :797, the rendezvous covers the staged path, and in place nothing uses it. That invariant is fragile though — anything added after the barrier that reads it would break silently on the in-place path, so it may be worth pinning down in a comment.
- **The table matches the data** bucket for bucket, and the tie at the tail keeping the smaller grid follows the file's stated policy. `_merge` takes the union of both edge sets so the unaligned dispatch/combine buckets compose correctly (one extra combine variant to JIT).
- **The new 256-block cap doesn't hit anything shipping**: no hip-side pin exceeds it, and the 80/152/192 entries in `tuning_configs.py` are FlyDSL's table and don't reach this path.

kawhil-amd:
> Read through the barrier rework carefully. The direction and the implementation both look right to me, and the reasoning in the commit message is unusually good. One blocking item, two things I'd want before merge, and some nits.
> 
> ## Blocking
> **`ep_intranode_1250x.hpp:743-762` has two comment blocks that contradict each other, with no code between them.**
> 
> The first (743-754) argues the acquire is required on the staged path — _"Dropping this fails the staged half of the identity check at every token count"_ — and goes on to justify device scope and per-block issue. The second (755-762) says it was removed, verified, and that the earlier failures were really the missing staging release.
> 
> Both land in the merged file, and the fence they describe no longer exists, so the first block is a leftover from an earlier iteration that now asserts the opposite of what the code does. In this particular function that is worse than no comment at all. Please drop one.
> 
> ## Before merge
> **Graph capture isn't in the verification list.** The barrier is more stateful across launches than it was: a private per-block epoch, plus block 0 fixing up `[gridDim.x, EpXdbFlagSlots)` so a later, wider grid still reads an in-step value. `tests/python/ops/dispatch_combine_v2/test_graph_capture.py` is exactly the test for that, and a serving stack replays this captured. The geometry-switching run you describe (ct=128/256/512/2048) is the right idea and does target the fix-up — worth doing the captured version too.
> 
> **The two result tables disagree, and the protocol is thin for the size of the claim.** PR body says 64 blocks `37.2 → 34.1` and 128 blocks `38.7 → 32.8`; the commit message says `37.2 → 35.0` and `38.7 → 33.8` (256 agrees at 30.6). Worth saying which run is which.
> 
> More importantly: `min of 2-3 runs`, eager only. `bench_ep.py` already reports means over `ITERS` and supports `MODES=graph`, so means over a few hundred iterations in both modes would cost little and carry more weight for a ±1-2 us difference.
> 
> One trap specific to this harness, which I hit this week measuring the same op: **the number you get at a fixed `ct` depends on the `SWEEP` list you asked for.** I have a reproducible case where `SWEEP=1024,2048,4096` reads 12% slower at ct=2048 than `SWEEP=2048` does — same binary, same routing, ITERS=400, repeated. I ran a position x M factorial and only that one cell is anomalous; I haven't found the mechanism yet. Practical consequence: please state the `SWEEP` used, and make sure numbers compared against each other came from the same one.
> 
> ## Nits
> **The fence removals now carry an implicit dependency on the window being uncached.** Both the dropped acquire and the in-place release rest on `CcoWindowAllocType()` returning uncached. That is the default and it holds in every configuration we actually ship — the two compile-time fallbacks sit below `HIP_VERSION 70051831`, which cannot build gfx1250 anyway — so this is not a correctness objection to the PR. But `CCO_UNCACHED_WINDOW=0` is a runtime switch that nothing in the repo sets, and flipping it on gfx1250 would now give silently wrong MoE output rather than an error. A line in the comment at :751 saying that env var is incompatible with this kernel, or a construction-time check, would close it for free.
> 
> **`_XDB_FLAG_SLOTS = 256` in `hip_backend.py` is the third hand-synced constant pair in this family** (`EpCombine1250xLdsBudget` / `MORI_COMB_LDS_BUDGET` is the existing one). The `raise` guarding it is good; deriving it through the JIT config would be better.
> 
> **The PR body's Motivation / Technical Details / Test Plan are empty template sections** while the commit message covers all of them well — please move it across, since the commit message is where all the reasoning currently lives.
> 
> `_qStore`'s comment change (output-width vs tile-width under fp8) is unrelated to this PR; harmless, just noting it.
> 
> ## Checked and happy with
> * **Epoch alignment holds** across a grid that grows or shrinks: block 0 writes `[gridDim.x, 256)`, every other block writes only its own slot, indices never overlap, and the peer lap is bounded at one.
> * **The `>=` / `==` split is right** and for the reasons given — 64-bit arena slots a peer can lap, 32-bit truncated fan lines only our own block 0 writes.
> * **Single-thread `*totalRecvTokenNum = 0` is safe**: it is only consumed by the staging loop at :797, the rendezvous covers the staged path, and in place nothing uses it. That invariant is fragile though — anything added after the barrier that reads it would break silently on the in-place path, so it may be worth pinning down in a comment.
> * **The table matches the data** bucket for bucket, and the tie at the tail keeping the smaller grid follows the file's stated policy. `_merge` takes the union of both edge sets so the unaligned dispatch/combine buckets compose correctly (one extra combine variant to JIT).
> * **The new 256-block cap doesn't hit anything shipping**: no hip-side pin exceeds it, and the 80/152/192 entries in `tuning_configs.py` are FlyDSL's table and don't reach this path.

  The barrier's fences and atomic scopes are chosen to be correct for any allocation type — arena slots use SCOPE_SYSTEM load/store, the staging release is __threadfence_system(), and fan-out uses SCOPE_AGENT. The release→signal(SYS)→poll(SYS) chain establishes ordering through scope-qualified operations, not through an assumption about caching. The removed acquire was redundant given this chain. The comments mention uncached as a secondary observation (peer rows additionally skip L2), but correctness doesn't rest on it.

jhchouuu (approved):

PR #582, opened 2026-08-19
chore(EPv2): initialize the routing dest_map once within the op construct

kawhil-amd:
##  Summary

  - Pre-allocate routing_dest_map in __init__ (both flydsl and hip backends) instead of calling torch.full_like(self.token_dest_map, -1) on every dispatch(return_routing=True) call.
  - The dispatch kernel writes every (tok, k) slot below num_tokens, and no consumer reads past that bound, so the -1 sentinel only matters for the initial state — a single fill at init time is sufficient.
  - routing_dest_map is kept separate from token_dest_map because return_routing=True passes the buffer into EpDispatchRoutingHandle for combine replay; sharing one buffer would let the next dispatch overwrite data that combine still needs.

 ##  Why the original code was wasteful

  - torch.full_like both allocates a new tensor and fills it on every call. In the bench loop this added a per-iteration alloc+fill whose cost was silently folded into the dispatch CUDA-event timing. In e2e serving the allocation only ran once per forward (dispatch → experts → combine), but the fill was still unnecessary.

##  Safety
  - The handle from call N is aliased by call N+1 — acceptable because the serving path never holds a handle across steps (dispatch → experts → combine within one forward).

### DIFF
--- a/python/mori/ops/dispatch_combine_v2/dispatch_combine_op.py
+++ b/python/mori/ops/dispatch_combine_v2/dispatch_combine_op.py
@@ -795,6 +795,7 @@ class EpDispatchCombineOp:
         half, so one rank resetting alone would leave the two disagreeing."""
         self.arena.zero()
         self.token_dest_map.fill_(-1)
+        self.routing_dest_map.fill_(self._null_flat)
         self.dest_pe_counter.zero_()
         self.dispatch_barrier.zero_()
         self.combine_barrier.zero_()
--- a/python/mori/ops/dispatch_combine_v2/dispatch_combine_op.py
+++ b/python/mori/ops/dispatch_combine_v2/dispatch_combine_op.py
@@ -695,13 +695,10 @@ class EpDispatchCombineOp:
             kern, dest_map = table[disp_spec], routing.disp_dest_tok_id_map
         else:
             kern = self._kernels.dispatch[disp_spec]
-            # return_routing hands the handle its own dest_map (filled -1 so unset
-            # (tok,k) slots read as null); a plain dispatch reuses the scratch one.
-            dest_map = (
-                torch.full_like(self.token_dest_map, -1)
-                if return_routing
-                else self.token_dest_map
-            )
+            if return_routing:
+                dest_map = self.routing_dest_map
+            else:
+                dest_map = self.token_dest_map
 
         kern(
             input=input,
--- a/python/mori/ops/dispatch_combine_v2/flydsl_backend.py
+++ b/python/mori/ops/dispatch_combine_v2/flydsl_backend.py
@@ -118,7 +118,8 @@ class EpDispatchCombineOpFlyDSL(EpDispatchCombineOp, backend="flydsl"):
         self.token_dest_map = torch.full(
             (max_tok_per_rank * topk,), -1, dtype=torch.int32, device=device
         )
-        self.routing_dest_map = torch.full_like(self.token_dest_map, -1)
+        self._null_flat = cfg.world_size * cfg.effective_max_recv
+        self.routing_dest_map = torch.full_like(self.token_dest_map, self._null_flat)
         self.dest_pe_counter = torch.zeros(
             cfg.world_size, dtype=torch.int32, device=device
         )
--- a/python/mori/ops/dispatch_combine_v2/flydsl_backend.py
+++ b/python/mori/ops/dispatch_combine_v2/flydsl_backend.py
@@ -118,6 +118,7 @@ class EpDispatchCombineOpFlyDSL(EpDispatchCombineOp, backend="flydsl"):
         self.token_dest_map = torch.full(
             (max_tok_per_rank * topk,), -1, dtype=torch.int32, device=device
         )
+        self.routing_dest_map = torch.full_like(self.token_dest_map, -1)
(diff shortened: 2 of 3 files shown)


jhchouuu:
Checked the coverage claim against all three dispatch bodies and it holds: portable (`ep_intranode_kernel.hpp:166-206`, every early-out writes `kNullFlat`), flydsl (`intranode_kernels.py:219-222`, one unconditional select + store over the full grid-strided range), and the 1250x COUNT/FINALIZE pair (matching `keep` predicates, lane partition covers `[0,numTokens) x [0,topk)` in both the `_etpi==1` and `_etpi>1` cases).

There's a stronger argument than the one in the body, worth using: `token_dest_map` is already reused across calls and never re-filled (only `reset()`, :800), and in `hip_backend.py` it starts as `torch.zeros` — where 0 decodes to a *valid* (pe 0, slot 0), not a null. So if coverage ever failed, the plain-dispatch path would already be silently wrong today. Proof by existing behaviour beats "the kernel writes every slot".

Two things I'd fix:

**1. `-1` is not the null sentinel in either backend.** hip's is `worldSize * EpFlatStride`, flydsl's is `npes * max_recv` (`intranode_kernels.py:138`). `-1 / stride` truncates to 0, so `-1` decodes to pe 0 / slot −1, passes `destPe < npes`, and reads one token *before* peer 0's buffer. The comment this PR removes ("filled -1 so unset (tok,k) slots read as null") was already wrong, and the body inherits it — "the -1 sentinel only matters for the initial state" describes something that was never a sentinel. If the init fill is meant to mean anything, fill with `cfg.world_size * cfg.max_recv`.

**2. The handle already has the antidote for this aliasing, one field over.** `disp_tok_id_to_src_tok_id_local` clones on first access precisely "so it survives the next dispatch overwriting the region" (:446-449). `disp_dest_tok_id_map` didn't need that before, because each call owned its buffer; now it does. Making it the same lazy-clone property costs nothing on the dispatch → experts → combine path and removes the one-live-handle constraint entirely. Worth doing because no current test holds two handles, so a regression here wouldn't be caught by anything.

Nits: `reset()` fills `token_dest_map` but not the new buffer. And the body would be more convincing with the number — the fill is `max_tok * topk` int32, ~393 KB at M=16384 / topk 6, so roughly 0.25 µs: 0.6% of a 39 µs dispatch at ct=64 and 0.05% at ct=16384. Fixed cost independent of `ct`, so it matters most in the decode regime. (It does sit inside `bench_ep.py`'s cuda-event window, so it is in every dispatch number that harness has produced — at that magnitude it changes no conclusion, which is worth saying explicitly rather than leaving implied.)

jhchouuu (approved):

kawhil-amd:
> Checked the coverage claim against all three dispatch bodies and it holds: portable (`ep_intranode_kernel.hpp:166-206`, every early-out writes `kNullFlat`), flydsl (`intranode_kernels.py:219-222`, one unconditional select + store over the full grid-strided range), and the 1250x COUNT/FINALIZE pair (matching `keep` predicates, lane partition covers `[0,numTokens) x [0,topk)` in both the `_etpi==1` and `_etpi>1` cases).
> 
> There's a stronger argument than the one in the body, worth using: `token_dest_map` is already reused across calls and never re-filled (only `reset()`, :800), and in `hip_backend.py` it starts as `torch.zeros` — where 0 decodes to a _valid_ (pe 0, slot 0), not a null. So if coverage ever failed, the plain-dispatch path would already be silently wrong today. Proof by existing behaviour beats "the kernel writes every slot".
> 
> Two things I'd fix:
> 
> **1. `-1` is not the null sentinel in either backend.** hip's is `worldSize * EpFlatStride`, flydsl's is `npes * max_recv` (`intranode_kernels.py:138`). `-1 / stride` truncates to 0, so `-1` decodes to pe 0 / slot −1, passes `destPe < npes`, and reads one token _before_ peer 0's buffer. The comment this PR removes ("filled -1 so unset (tok,k) slots read as null") was already wrong, and the body inherits it — "the -1 sentinel only matters for the initial state" describes something that was never a sentinel. If the init fill is meant to mean anything, fill with `cfg.world_size * cfg.max_recv`.
> 
> **2. The handle already has the antidote for this aliasing, one field over.** `disp_tok_id_to_src_tok_id_local` clones on first access precisely "so it survives the next dispatch overwriting the region" (:446-449). `disp_dest_tok_id_map` didn't need that before, because each call owned its buffer; now it does. Making it the same lazy-clone property costs nothing on the dispatch → experts → combine path and removes the one-live-handle constraint entirely. Worth doing because no current test holds two handles, so a regression here wouldn't be caught by anything.
> 
> Nits: `reset()` fills `token_dest_map` but not the new buffer. And the body would be more convincing with the number — the fill is `max_tok * topk` int32, ~393 KB at M=16384 / topk 6, so roughly 0.25 µs: 0.6% of a 39 µs dispatch at ct=64 and 0.05% at ct=16384. Fixed cost independent of `ct`, so it matters most in the decode regime. (It does sit inside `bench_ep.py`'s cuda-event window, so it is in every dispatch number that harness has produced — at that magnitude it changes no conclusion, which is worth saying explicitly rather than leaving implied.)

Good comments, all fixed now.

PR #584, opened 2026-08-19
Drop local xla_ffi headers in favor of jax-backed header path

pemeliya:
There is no need to keep a local copy of FFI headers inside MORI repo - we can fetch them from JAX installed in the system (or from a user-provided directory)


jhchouuu (approved):
LGTM, thanks！

PR #587, opened 2026-08-20
feat(umbp): peer-local multi-backend placement and logical tier management

wuyl1:
## What this adds

Lets one peer own several named backends and route between them, instead of
serving exactly one medium.

- **Multi-backend peers.** A peer registers up to 16 backends
  (`kMaxBackendsPerPeer`); the registry hands out dense ids from 0 and an id
  travels as one half of a buffer address, `(backend_id, buffer_index)`. The
  bound is sized for what a policy expands into — an 8-GPU node declaring HBM
  once expands to eight instances, plus hot/warm DRAM and a pair of SSDs lands
  at twelve. `medium` remains the legacy path, read only when no policy names
  the backends.
- **Placement policies.** `SingleBackendPolicy`, `WeightedPlacementPolicy`
  (deterministic hashing over per-backend weights) and `TieredPlacementPolicy`
  behind one `PoolPolicy` interface, which separates `PutOrder` from
  `ReadOrder`.
- **Logical tiers.** A JSON policy compiles into a `LogicalTierGraph` answering
  `PutOrder`, `ReadOrder`, `TransitionTargetOrder`, `PromoteTargetOrder` and
  `MemberUtilization`. One shared validator enforces the invariants: each
  backend belongs to exactly one tier, exactly one tier is the entry, and an
  offload edge must point at a later-declared tier — stricter than acyclic, and
  what makes a backward edge a parse error.
- **Tier transitions.** Plan under the pool lock, release it for the byte copy,
  then finish under the lock again, so reorganising capacity in the background
  never stalls the data plane. A failed job is retried at most three times with
  a 1/2/4 ms backoff, and only while the source tier is still above its low
  watermark — the point of the retry is to relieve pressure, not to complete
  one particular move. After a successful step that leaves the tier still
  pressured, the worker tops the candidate batch back up instead of waiting for
  the next commit to notice. Reclaim frees bytes and lets an `on_evict` tier
  demote rather than drop the key, so the data stays reachable; discard removes
  every copy on the peer for when the value itself must stop existing.
- **Ranged I/O routed through the pool.** `BatchPutRanges` and `BatchGetRanges`
  reached past the pool at the backend registry, one picking a backend with
  `registry_.Get(route.tier)` and the other scanning `registry_.All()`. A
  configured policy still loaded and the master still reported both tiers, but
  nothing was placed by it and the placement index stayed empty, which left
  watermark offload and promote-on-read inert for the tree connector, which
  issues only ranged calls. Ranged writes now allocate, commit and abort
  through `PeerPool` with `route.logical_tier` threaded into the placement
  request; slots are still held across one `Transfer` for the whole batch, with
  the commits and aborts gathered into batch calls rather than issued per key.
  Ranged reads resolve in one batch, which repairs the placement index, records
  the read against the tier that served it and applies promote-on-read.
- **Benchmark tooling.** `umbp_tier_bench` drives real `PoolClient` instances
  from synthetic workloads or replayed traces. The generator, runner and
  payload code live in a separate `umbp_workload_bench` library; the trace
  codec and recorder stay in `umbp_common` on purpose, because `PoolClient`
  records live production traffic through them. A trace carries a schema
  version and an envelope version, and stores no payload bytes — they are
  regenerated from (key, operation_id, seed).

`examples/umbp/multi_backend_policy.json` is the one authoritative example
topology; `umbp_backend_policy_config` loads that exact file through
`UMBP_EXAMPLE_POLICY_PATH`, so the shipped example and the parser cannot drift
apart silently.

## Testing

- `test_peer_pool` adds 23 cases over tier placement and migration: watermark
  offload, promote-on-read, move vs copy, eviction racing a migration, reclaim
  vs discard. Registered as `umbp_peer_pool` so the CI gate discovers it.
- Policy parsing is covered positively against the shipped example, and
  negatively for inverted watermarks, a backend in two tiers and a backward
  offload edge, plus a case pinning that a failed lowering leaves the config
  untouched.
- The ranged routing case makes the entry tier the second-registered DRAM
  instance, so selecting by `TierType` and asking the policy disagree. Without
  the fix it puts the key on the wrong backend and leaves `TierReadHits` empty.
- The non-CCO ctest suite passes here (52 of 53 cases, 34 of the 35 the gate
  regex matches). The exception is `umbp_local_client`, flaky at this merge
  base in a local GPU gather path this branch does not touch.

## Known follow-ups

- `PoolWorkloadClient::Get` sleeps before testing the deadline, so every miss
  costs one `retry_interval` (5 ms default) of benchmark wall clock even when
  the key was never going to arrive.
- The publication-deadline map is pruned only on a completed GET, so a
  write-only run grows it with the number of keys written.
- `LoadPolicy` re-checks the per-peer backend limit that `ApplyBackendPolicy`
  already rejects.

Formatting will be handled with the `/fix-precommit` label — the repo's C++ is
written at ~90 columns while `.clang-format` sets `ColumnLimit: 100`, so
clang-format v20 wants to reflow lines this branch did not touch.


TianDi101:
Some suggestions:
1. Need to support ranged APIs
2. Optimize lock granularity for better performance, we can do this after merge this PR
3. Currently we only have a boolean promote_on_read option, it is better to make promote policy key-value based just like offload policy so we can add more policies, such as: promote after several hits, promote on read and etc.

PR #588, opened 2026-08-20
fix(ionic): enable MoRI-EP inter-node on gfx942 + AINIC/ionic

raviguptaamd:
## Summary

Inter-node **MoRI-EP** (dispatch/combine over RDMA/IBGDA) hung on **MI308X (gfx942) + AINIC (ionic/PSD)**. Intra-node EP and MoRI-IO already worked; only cross-physical-node EP hung, entering the first dispatch and later spinning in `EpCombineInterNodeV1Kernel`. This PR fixes four ionic/PSD-specific issues so 2-node EP16 dispatch+combine pass.

This `gfx942 + ionic` combination is not currently covered by CI (CI runs gfx950+ionic and gfx942+bnxt), which is why these slipped through.

## Root causes & fixes

1. **QP never put in GDA mode** — `ionic_dv_qp_set_gda()` was never called, so the GPU-issued doorbell had no effect and dispatch never transmitted. We now enable GDA (GPU-Direct Async) at QP creation. Optional symbol (older `libionic` lacks it); `MORI_DISABLE_IONIC_GDA` opts out. (`ionic.cpp`, `dv_loader.hpp`)

2. **GRH `hop_limit = 1`** — dropped every packet crossing a routed leaf-spine fabric (OK for same-leaf/loopback, fatal cross-node). Set to `255` (bnxt uses 255, mlx5 uses 64). (`ionic.cpp`)

3. **Atomic ibuf registered with plain `ibv_reg_mr`** — needs `amdgpu_peermem` for GPU memory (absent on this stack). Register via **dmabuf** first (matching the working symmetric-heap dmabuf path), fall back to `ibv_reg_mr` for host memory. (`ionic.cpp`)

4. **GPU-issued RDMA atomic-add does not transmit on ionic** — with `PostAtomic` the NIC tx counter stays 0 (host-posted atomics via `ib_atomic_bw` do work; GPU-posted do not). MoRI-EP uses non-fetch atomics only for **single-writer signal counters**, so we emulate them with a plain remote **write** of the accumulated value (`AMO_ADD`-from-zero == write of value). The cross-device combine barrier accordingly writes the absolute expected value `barrierFlag * numQpPerPe` once instead of `numQpPerPe` add-1s. (`shmem_ibgda_kernels.hpp`, `internode_v1.cpp`)

All PSD-specific changes are gated to the ionic path (`ProviderType::PSD` / `MORI_DEVICE_NIC_IONIC`); mlx5 and bnxt paths are untouched.

## Validation

- Hardware: 2 physical nodes × 8× MI308X (gfx942) + AINIC/ionic, EP16 inter-node.
- `run_internode_test.sh --cmd test --kernel-type v1`: **"Node 0 Dispatch Pass" + "Node 0 Combine Pass" across many rounds, 0 failures** (previously hung on round 1).
- Prereqicites in the environment: ionic driver with working GPU doorbell, and lossless RoCE (PFC) on the fabric.

## Known limitation / discussion

- Currently requires **`MORI_NUM_QP_PER_PE=1`**. With multiple QPs the emulated atomic **writes** can be reordered/lost across queues (combine receiver observes a stale `nodeRecvTokenNum` and spins). Single QP serializes the signals. Making the emulated-atomic signal path safe under multi-QP is left as follow-up — feedback welcome on the preferred approach (e.g. per-generation counters vs. ordering fences), or whether a proper GPU-atomic path is planned for ionic.

## Test plan

- [ ] Maintainer review of the PSD-gated changes
- [ ] CI coverage for gfx942 + ionic if feasible
- [ ] Multi-QP follow-up for the emulated-atomic signal path

### DIFF
--- a/include/mori/application/transport/rdma/providers/dv_loader.hpp
+++ b/include/mori/application/transport/rdma/providers/dv_loader.hpp
@@ -183,6 +183,7 @@ struct IonicDvApi {
   using pd_set_udma_mask_t = int (*)(struct ibv_pd*, uint32_t);
   using create_cq_ex_t = struct ibv_cq_ex* (*)(struct ibv_context*, struct ibv_cq_init_attr_ex*,
                                                struct ionic_cq_init_attr_ex*);
+  using qp_set_gda_t = int (*)(struct ibv_qp*, bool, bool);
 
   get_ctx_t get_ctx = nullptr;
   qp_get_udma_idx_t qp_get_udma_idx = nullptr;
@@ -192,6 +193,7 @@ struct IonicDvApi {
   pd_set_rqcmb_t pd_set_rqcmb = nullptr;
   pd_set_udma_mask_t pd_set_udma_mask = nullptr;
   create_cq_ex_t create_cq_ex = nullptr;
+  qp_set_gda_t qp_set_gda = nullptr;
 
   void* handle = nullptr;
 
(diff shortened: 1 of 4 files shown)


raviguptaamd:
Closing this PR — after a proper A/B rebuild test on current hardware, these code changes turn out to be **unnecessary**.

I re-verified against latest `main` (403b02a4) with a fresh in-container `pip install .` rebuild on 2 physical nodes (2×8 MI308X/gfx942 + AINIC/ionic), running `run_internode_test.sh --cmd test --kernel-type v1`:

- **Clean latest main, no changes:** 57/57 dispatch + 57/57 combine PASS (both default `NUM_QP_PER_PE=4` and `=1`).
- **Even the older base (35e2effb), completely unpatched:** 57/57 PASS.

The inter-node MoRI-EP hang I originally saw was an **environment** problem, not a MoRI code bug. It is resolved by two infrastructure fixes:
1. AINIC ionic driver upgrade to **26.03.27.002** (fixes the GPU-issued IBGDA doorbell so device-initiated RDMA actually transmits), and
2. enabling **PFC / lossless RoCE** on the leaf switches (fixes multi-QP incast tail-drop).

With both in place, upstream MoRI works out-of-the-box on gfx942+ionic — no `hop_limit`, GDA, dmabuf, or atomic-emulation changes needed. Apologies for the noise; closing.

PR #594, opened 2026-08-24
feat(cco): add Triton device API bindings

yangyuhuiling:
## Motivation

`mori.ir.triton` previously supported only MORI-SHMEM. CCO already exposed LSA, SDMA, and GDA through device bitcode for FlyDSL, but Triton kernels could not use these APIs.

This PR adds CCO Triton support, shares one ABI definition across FlyDSL and Triton, and provides correctness and performance comparisons against the existing C++ implementation.

The process-wide ROCm initialization policy was split into [#609](https://github.com/ROCm/mori/pull/609) and is now part of `main`; it is no longer in this PR's effective diff.

### Device API

- Added `mori.ir.triton.cco` with 68 operations covering DevComm queries, LSA pointers, SDMA, and GDA.
- Added framework-independent `CCO_DEVICE_FUNCTIONS` metadata shared by FlyDSL and Triton.
- Converted side-effect wrappers to return `int32` status values required by Triton `extern_elementwise`.
- Reused `libmori_cco_device.bc` with code object version 5.
- Passed `dc.ptr` and `win.handle` explicitly; no SHMEM module initialization hook is required.
- Added class-style `DevComm`, `Window`, `Gda`, and `Sdma` facades aligned with FlyDSL naming.
- Compile-time checks reject invalid cooperative scope, signal operation, and thread-mode combinations. `ThreadMode.AGGREGATE` requires `CoopScope.THREAD` and participation by every warp lane.
- ABI tests disassemble the generated bitcode and verify the C++ wrapper return types against the Python metadata.

### CCO correctness fix

- Fixed the tolerated `hipIpcOpenMemHandle` path so it consumes only its own sticky `hipErrorPeerAccessAlreadyEnabled` status instead of leaking it to the next Torch allocation.

### Tests and benchmarks

- Added an end-to-end CCO Triton LSA/SDMA/GDA example and pytest coverage.
- Added matched Triton and C++ LSA, SDMA, and IBGDA put/get latency and bandwidth benchmarks.
- Added an automated comparison runner producing JSON and Markdown reports.

## Test Plan

```bash
BUILD_BENCHMARK=ON BUILD_CCO_SDMA=ON BUILD_UMBP=OFF \
  pip install -e . --no-build-isolation

pytest -q \
  tests/python/cco/test_triton_bindings.py \
  tests/python/cco/test_triton_lsa.py \
  tests/python/cco/test_triton_sdma.py \
  tests/python/cco/test_triton_benchmark.py

pytest -q tests/python/cco/test_sdma_api.py

# Runtime GDA requires a machine with a working device-initiated RDMA path.
MORI_ENABLE_SDMA=1 \
MORI_DISABLE_TOPO=1 \
MORI_RDMA_DEVICES=rocep9s0 \
MORI_DEVICE_NIC=ionic \
torchrun --standalone --nproc_per_node=2 \
  examples/cco/ir/test_triton_cco.py --transport all
```

## Coverage and Results

- CI runtime coverage: Triton bindings, LSA, SDMA, and benchmark helpers.
- CI GDA coverage: compile/link of every monomorphized Triton GDA variant only. The current MI355X-AINIC runner does not provide a working intranode GDA-FULL runtime path, even when both ranks are pinned to one HCA.
- Manual hardware coverage: the all-transport example and matched IBGDA benchmarks passed on two GPUs with a working single AINIC rail.
- Existing FlyDSL SDMA suite: 12 tests passed.
- Existing Triton SHMEM regression passed.
- Full Triton/C++ comparison completed across 132 data points.

## Performance Notes

On MI355X with one working AINIC rail:

- IBGDA 8 MiB put: both implementations approximately `38.70 GB/s`.
- IBGDA 8 MiB get: Triton `48.05 GB/s`, C++ `48.16 GB/s`.
- SDMA 8 MiB: Triton reaches approximately 98.6% of C++ throughput.
- LSA 8 MiB put is effectively identical.
- Small-message SDMA latency is 33-37% higher in Triton. The size-independent shape indicates fixed per-call overhead, but it has not yet been attributed conclusively; likely contributors include the extern call and argument-marshalling path.
- IBGDA small-message latency remains within approximately ±5% of C++.

## Known Gap

GDA runtime correctness is not exercised by CI until a runner with a supported intranode GDA path is available. CI currently prevents API/compiler regressions through compile-only coverage; runtime and performance numbers above are manual hardware results.

## Submission Checklist


jhchouuu:
Read through this fairly closely. No correctness bug found in the binding layer itself, and the C++ fix bundled in here is a real one. The points below are mostly about scope, one lost invariant, and a gap in the ABI checks.

First, the good parts, because they are load-bearing:

**`src/cco/cco_init.cpp:1897-1900` is a genuine bug fix, done the right way.** cco leaving a sticky HIP error for the next caller has a history: a blanket `hipGetLastError()` consume added at the *Python* boundary in `Communicator.init` was reverted because it SIGSEGV'd all 12 cco SDMA Python tests. The correct form is a narrow `(void)hipGetLastError()` immediately after the one call whose failure is tolerated, capturing the real status first — which is exactly what this does, and it matches the four sites `main` already has in this file. Worth splitting into its own commit for bisectability, but the change is right.

**The FlyDSL-regression risk from `void` → `int` is directly tested.** `tests/python/cco/test_triton_bindings.py:111` asserts FlyDSL's externs and the shared table agree on `_args`/`_ret` per symbol. That is the guard I would have asked for.

**The GDA compile-only gate does not silently pass.** `assert "GDA compile-only PASS" in output` means a skipped GDA step fails the job rather than turning it green.

---

## 1. A correctness precondition was dropped in the move

`python/mori/cco/device/ops.py:53`. The old `ThreadMode` in `flydsl/handles.py` documented:

> `AGGREGATE` — the warp's lanes are coalesced into one transfer by cco (**requires `CoopScope.THREAD`; all lanes must enter**).

The shared version keeps only *"How participating lanes contribute to a GDA data-path operation."* That parenthetical is not prose — it is why `DATA_PATH_TAG` has `(AGGREGATE, THREAD)` and no AGGREGATE+WARP/BLOCK, and violating it (partial lane participation) hangs or corrupts. It is now recorded nowhere in the repo, and Triton users are precisely the audience that has never seen it. Please carry it across.

## 2. The ABI check stops one step short of the ground truth

Three tests exist, but none compares the Python table against the C++ *return type*:

- `test_triton_bindings.py:39` — `ret != "void"` checks the table.
- `test_triton_bindings.py:59` — checks the bitcode *defines* each symbol; names only, not types.
- `test_triton_bindings.py:111` — FlyDSL vs Triton, but both read the same table, so agreement is structural.

So: add a wrapper in `cco_device_wrapper.cpp`, forget to change `void`/add `return 0`, declare it `int32` in the table — all three pass, and it breaks at Triton call time. `llvm-dis` on the `.bc` plus a `define\s+(\S+)\s+@cco_` match in the existing bitcode test would close it cheaply. (I did check the converted wrappers by hand — `return 0` is placed after the body in each, and the 68 count matches the families.)

## 3. `coop` dispatch has an open `else`

`python/mori/ir/triton/cco/handles.py`, 9 sites (`:338`, `:383`, `:431`, `:523`, `:607`, `:621`, …), all shaped:

```python
if coop == 0:      # THREAD
elif coop == 1:    # WARP
else:              # BLOCK -- and also 3, -1, or any typo
```

`coop` is a `tl.constexpr`, so `tl.static_assert(coop in (0, 1, 2))` catches this at compile time instead of silently running BLOCK. The literals also bypass `CoopScope`, so if those values ever move, nothing here fails.

## 4. `cco_system_fence` needs its ordering guarantee written down

`src/cco/device/cco_device_wrapper.cpp:107`:

```c
CCO_DEV int cco_system_fence(int leaderOnly) {
  if (!leaderOnly || threadIdx.x == 0) __threadfence_system();
  return 0;
}
```

A divergent fence — under `leaderOnly` only thread 0 of the block issues it. That is usually enough for publication (a system fence writes back the device-shared L2), but it is a subtle primitive to expose in the shared wrapper for benchmark use, and people will copy it. Either document what it does and does not order, or keep it in the benchmark if that is its only consumer.

## 5. The test directory gained a package for one import style, and now has two

`tests/python/cco/__init__.py` is 21 lines of licence header and no code; its only effect is making the directory a package so the new tests can say `from tests.python.cco.helper import run_triton_example`. Meanwhile the pre-existing `test_sdma_api.py:81` still does the unqualified `from helper import sync, fill, zero, read`, relying on `sys.path.insert(0, this_dir)` at `:627`.

That still works — but only because the unqualified import sits inside the spawned-child function whose `sys.path` was patched, not at module scope. Move that line up, or change pytest's import mode, and it breaks. `conftest.py` is what pytest provides for exactly this and needs no `__init__.py` and no change to existing files. At minimum, pick one style for the directory.

## 6. Scope: this is five changes in one PR

Triton bindings + a global import-time ROCm bootstrap + a cco C++ bug fix + a benchmark suite and comparison runner + an edit to `.claude/skills/deploy-mori/SKILL.md`.

The bootstrap is the one I would split out. `python/mori/cco/__init__.py:28` calls `_ensure_rocm_runtime()` at module scope, so **every** `import mori.cco` now reads `/proc/self/maps` and may preload a runtime — not just Triton users. It also makes HIP runtime selection depend on import order ("If HIP is already mapped, for example because Torch was imported first, it is reused"), so `import torch; import mori.cco` and the reverse can now resolve differently. That is 125 lines plus 227 lines of tests (`test_rocm_bootstrap.py`, `test_rocm_import_order.py`), and the PR description does not mention it at all. It deserves its own PR and its own description. The skills file looks unrelated — drop it.

## 7. Coverage is narrower than the description implies

CI runs bindings/LSA/SDMA, the benchmark tests, and GDA **compile-only**. So GDA has no runtime coverage — every IBGDA number in the description is hand-measured. The runner limitation is real and the comment explains it, but it is worth stating as a known gap with an issue to track. Separately, `examples/cco/ir/test_triton_cco.py` is 816 lines — the largest single file here — and is a test in all but name, yet only appears in the manual Test Plan.

## 8. Smaller

- `setup.py:395` `_ROCM_VERSION_RE` runs over `hipconfig --version` captured with `stderr=subprocess.STDOUT` (`:420-422`), so any warning line containing `X.Y` can match first. The `.info/version` candidates are tried first so the risk is low, but a regex over merged stderr is fragile.
- The 33–37% small-message SDMA latency gap is reported without attribution. Large messages are at parity and IBGDA small messages are within ±5%, so the shape says fixed per-call overhead (`extern_elementwise` calling convention / argument marshalling) rather than a path problem. That is the number people will quote — worth a sentence on where it goes and whether it is expected to stay.

`pure` is correctly applied, for the record: only `lsa_ptr` and the four `devcomm_*` queries. Nothing that polls (`gda_wait_signal`, `gda_read_signal`) is marked pure, which is the mistake that would have let the compiler hoist a wait out of a loop.

kawhil-amd:
Maybe we can add a triton gluon example？

jhchouuu (approved):

PR #600, opened 2026-08-24
[AMD][DSV4] fix: recognize fp4_blockwise in the AUTO tuning config

karverma-amd:
dispatch_combine.py supports fp4_blockwise combine throughout -- _QUANT_TYPE_MAP, _BLOCKWISE_COMBINE_QUANT_TYPES, _FP4_COMBINE_KERNELS -- but quant_type_to_config_str was never updated alongside it. Its enum branch stops at Fp8BlockwiseQuant and falls through to `raise ValueError`, so a config using fp4 blockwise combine cannot be normalized into the string the tuned tables are keyed by.

The cost is larger than a missing combine lookup. The exception escapes before either table is consulted, so such a config also loses its tuned *dispatch* rules and falls back to hard-coded launch geometry. The only signal is one  WARNING line from the enclosing `except` ("AUTO tuning: failed to load config"); results are unaffected, so the cost is throughput nobody is looking for.
Note on scope: no shipped config under `tuning_configs/` contains fp4_blockwise` (only `none` and `fp8_direct_cast`), so the combine lookup for fp4 still misses and falls back. The throughput recovered here is on dispatch.

Add "fp4_blockwise" to _QUANT_TYPE_CONFIG_STRS and the matching Fp4BlockwiseQuant enum branch, so both the string and enum forms normalize consistently.

Genuinely unknown quant types still raise: this closes a gap between two modules that had drifted apart, it does not make the function permissive.

Adds tests/python/test_tuning_config_quant_types.py covering both directions of every mapped type, the invariant that anything nameable by enum is also in the string set, and that unknown types still raise.

## Submission Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,6 +189,15 @@ jobs:
             cd $GITHUB_WORKSPACE && MORI_DISABLE_P2P=1 MORI_ENABLE_SDMA=0 timeout 300 pytest tests/python/ops/test_dispatch_combine_async_ll.py -v
           "
 
+      - name: MORI-EP (AUTO tuning config quant types, CPU)
+        run: |
+          # Pure mapping test: no GPU, no RDMA. Named explicitly because the
+          # only directory-level collection in this workflow is tests/python/io,
+          # so a file elsewhere under tests/python runs nowhere by default.
+          $CT exec -e PYTHONPATH=$GITHUB_WORKSPACE $CONTAINER bash -c "
+            cd $GITHUB_WORKSPACE && timeout 60 pytest tests/python/test_tuning_config_quant_types.py -v
+          "
+
       - name: MORI-EP bench
         run: |
           $CT exec -e PYTHONPATH=$GITHUB_WORKSPACE $CONTAINER bash -c "
--- a/python/mori/ops/tuning_config.py
+++ b/python/mori/ops/tuning_config.py
@@ -88,7 +88,12 @@ _KERNEL_TYPE_NAMES = frozenset(
     {"IntraNode", "InterNode", "InterNodeV1", "InterNodeV1LL", "AsyncLL", "IntraNodeLL"}
 )
 
-_QUANT_TYPE_CONFIG_STRS = {"none", "fp8_direct_cast", "fp8_blockwise"}
+_QUANT_TYPE_CONFIG_STRS = {
+    "none",
+    "fp8_direct_cast",
+    "fp8_blockwise",
+    "fp4_blockwise",
+}
 
 
 def kernel_type_to_config_str(kernel_type) -> str:
@@ -118,6 +123,8 @@ def quant_type_to_config_str(quant_type) -> str:
             return "fp8_direct_cast"
         if name == "Fp8BlockwiseQuant":
             return "fp8_blockwise"
+        if name == "Fp4BlockwiseQuant":
+            return "fp4_blockwise"
     raise ValueError(f"Unknown quant_type: {quant_type!r}")
 
 
--- a/python/mori/ops/tuning_config.py
+++ b/python/mori/ops/tuning_config.py
@@ -88,12 +88,18 @@ _KERNEL_TYPE_NAMES = frozenset(
     {"IntraNode", "InterNode", "InterNodeV1", "InterNodeV1LL", "AsyncLL", "IntraNodeLL"}
 )
 
-_QUANT_TYPE_CONFIG_STRS = {
-    "none",
-    "fp8_direct_cast",
-    "fp8_blockwise",
-    "fp4_blockwise",
+# Single source for the quant-type mapping. Keyed by EpDispatchCombineQuantType
+# member name; the accepted config strings are derived rather than restated,
+# because holding the two in separate literals is what let them drift apart
+# (fp4_blockwise reached dispatch_combine.py but never the enum branch here).
+_QUANT_ENUM_TO_CONFIG_STR = {
+    "None_": "none",
+    "Fp8DirectCast": "fp8_direct_cast",
+    "Fp8BlockwiseQuant": "fp8_blockwise",
+    "Fp4BlockwiseQuant": "fp4_blockwise",
 }
+
+_QUANT_TYPE_CONFIG_STRS = frozenset(_QUANT_ENUM_TO_CONFIG_STR.values())
 
 
 def kernel_type_to_config_str(kernel_type) -> str:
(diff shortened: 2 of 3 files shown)


jhchouuu:
Verified the fix locally rather than just reading it — `Fp4BlockwiseQuant` is a real enum member (`dispatch_combine.hpp:64`, exported at `pybind_ops.cpp:372`) that the enum branch never mapped, and I ran all 14 mapping cases standalone: all pass.

The "it also loses the tuned *dispatch* rules" claim checks out. `_qt_str` is assigned at `dispatch_combine.py:552`, before `TuningConfigManager` is built at :553, inside the same `try` — and `_qt_str` only ever filters *combine* lookups (`:975`, `qt = self._qt_str if not is_dispatch else None`). So dispatch never needed the string at all; losing its rules was pure collateral. Good catch, and the fix is right.

Three things before merge:

**1. `test_every_mapped_enum_is_in_the_config_str_set` can never fail.** The `except ValueError: continue` swallows exactly the drift the test exists to detect — add `Fp6Quant` tomorrow and this stays green. After this fix all four members map, so there is nothing to skip: drop the try/except and assert directly.

**2. No CI job collects the new file.** `ci.yml` lists test files individually; the only directory-level run is `pytest tests/python/io/`. A file at the `tests/python/` root is picked up by nobody. This is one of the few mori tests needing neither GPU nor RDMA, so wiring it in is cheap — please add a line to an existing job.

**3. "Nothing fails loudly" is not quite right.** `dispatch_combine.py:589` does emit `logger.warning("AUTO tuning: failed to load config (%s); using hard-coded fallback.")`. Suggest rewording to "only a WARNING line, with no effect on results" — otherwise the next reader hunts for a silent path that does not exist.

One note for the body: no shipped JSON under `tuning_configs/` contains `fp4_blockwise` (only `none` and `fp8_direct_cast`), so the combine lookup for fp4 will still miss and fall back. That is correct behaviour, but worth stating that the win recovered here is on **dispatch**, so nobody expects a combine speedup from this PR.

Optional, to stop this recurring: the mapping is written twice — the set and the if-chain — which is what drifted. One dict removes that by construction:

```python
_QUANT_ENUM_TO_STR = {
    "None_": "none",
    "Fp8DirectCast": "fp8_direct_cast",
    "Fp8BlockwiseQuant": "fp8_blockwise",
    "Fp4BlockwiseQuant": "fp4_blockwise",
}
_QUANT_TYPE_CONFIG_STRS = set(_QUANT_ENUM_TO_STR.values())
```

and it turns the invariant test into an actual invariant.

karverma-amd:
> Verified the fix locally rather than just reading it — `Fp4BlockwiseQuant` is a real enum member (`dispatch_combine.hpp:64`, exported at `pybind_ops.cpp:372`) that the enum branch never mapped, and I ran all 14 mapping cases standalone: all pass.
> 
> The "it also loses the tuned _dispatch_ rules" claim checks out. `_qt_str` is assigned at `dispatch_combine.py:552`, before `TuningConfigManager` is built at :553, inside the same `try` — and `_qt_str` only ever filters _combine_ lookups (`:975`, `qt = self._qt_str if not is_dispatch else None`). So dispatch never needed the string at all; losing its rules was pure collateral. Good catch, and the fix is right.
> 
> Three things before merge:
> 
> **1. `test_every_mapped_enum_is_in_the_config_str_set` can never fail.** The `except ValueError: continue` swallows exactly the drift the test exists to detect — add `Fp6Quant` tomorrow and this stays green. After this fix all four members map, so there is nothing to skip: drop the try/except and assert directly.
> 
> **2. No CI job collects the new file.** `ci.yml` lists test files individually; the only directory-level run is `pytest tests/python/io/`. A file at the `tests/python/` root is picked up by nobody. This is one of the few mori tests needing neither GPU nor RDMA, so wiring it in is cheap — please add a line to an existing job.
> 
> **3. "Nothing fails loudly" is not quite right.** `dispatch_combine.py:589` does emit `logger.warning("AUTO tuning: failed to load config (%s); using hard-coded fallback.")`. Suggest rewording to "only a WARNING line, with no effect on results" — otherwise the next reader hunts for a silent path that does not exist.
> 
> One note for the body: no shipped JSON under `tuning_configs/` contains `fp4_blockwise` (only `none` and `fp8_direct_cast`), so the combine lookup for fp4 will still miss and fall back. That is correct behaviour, but worth stating that the win recovered here is on **dispatch**, so nobody expects a combine speedup from this PR.
> 
> Optional, to stop this recurring: the mapping is written twice — the set and the if-chain — which is what drifted. One dict removes that by construction:
> 
> ```python
> _QUANT_ENUM_TO_STR = {
>     "None_": "none",
>     "Fp8DirectCast": "fp8_direct_cast",
>     "Fp8BlockwiseQuant": "fp8_blockwise",
>     "Fp4BlockwiseQuant": "fp4_blockwise",
> }
> _QUANT_TYPE_CONFIG_STRS = set(_QUANT_ENUM_TO_STR.values())
> ```
> 
> and it turns the invariant test into an actual invariant.

@jhchouuu  -

Pushed as `7194db5`. All three points addressed, plus the optional dedup, which
turned out to be the root cause rather than a nicety.

**1. The invariant test could not fail — fixed.** You were right that
`except ValueError: continue` swallowed exactly the drift the test exists to
detect. Removed; the loop now calls `quant_type_to_config_str` directly, so an
enum member added without a mapping fails. Verified by mutation rather than by
inspection: deleting the `fp4_blockwise` entry turns `15 passed` into
`4 failed` (the invariant test plus the three fp4 cases), where the previous
version stayed green.

**2. Nothing collected the file — fixed.** Added a step to the existing job:

```yaml
- name: MORI-EP (AUTO tuning config quant types, CPU)
```

placed with the EP tests, with a comment noting that `tests/python/io` is the
only directory-level collection so a file elsewhere runs nowhere by default.
Needs neither GPU nor RDMA, as you said.

**3. "Nothing fails loudly" was wrong — corrected.** Confirmed
`dispatch_combine.py:591` emits
`logger.warning("AUTO tuning: failed to load config (%s); using hard-coded fallback.", exc)`.
The docstring now says results stay correct and the cost is throughput, instead
of implying a silent path that does not exist.

**Optional dedup — taken, and it was the actual root cause.** The mapping was
written twice, as a set of accepted strings and as an if-chain, and those two
literals drifting apart is precisely how `fp4_blockwise` came to be supported in
`dispatch_combine.py` but unmappable here. There is now one dict and the set is
derived from it, so they cannot disagree by construction:

```python
_QUANT_ENUM_TO_CONFIG_STR = {
    "None_": "none",
    "Fp8DirectCast": "fp8_direct_cast",
    "Fp8BlockwiseQuant": "fp8_blockwise",
    "Fp4BlockwiseQuant": "fp4_blockwise",
}
_QUANT_TYPE_CONFIG_STRS = frozenset(_QUANT_ENUM_TO_CONFIG_STR.values())
```

That also makes the invariant test a real invariant, as you noted.

**On your note about shipped configs — confirmed, and worth stating in the
body.** Nothing under `python/mori/ops/tuning_configs/` contains
`fp4_blockwise`; counting the quant strings across all shipped JSON gives 297
`"none"` and 257 `"fp8_direct_cast"` and nothing else. So the combine lookup for
fp4 still misses and falls back, correctly. The throughput this PR recovers is
on **dispatch** only. Nobody should expect a combine speedup from it.

Also confirming your reading of the control flow, since it is the reason the
blast radius is wider than a combine miss: `self._qt_str` is assigned before
`TuningConfigManager.get_instance(...)` inside the same `try`, and `_qt_str`
only ever filters combine lookups. Dispatch never needed the string, so losing
its tuned rules was pure collateral damage from the exception escaping early.

jhchouuu (approved):

PR #606, opened 2026-08-26
feat(EPv2): support wide EP (larger than 32) in intranode kernels

kawhil-amd:
## Summary  
- Support EP configurations where `worldSize` exceeds the hardware wave size
  (e.g. EP36 on gfx1250 with WAVE=32), across both the C++ TDM kernels and the
  FlyDSL combine kernel.

### C++ TDM kernels (`ep_intranode_kernel.hpp`, `ep_intranode_1250x.hpp`)

  - **Phase 2 barrier hoist**: move the grid barrier wait+reset before the
    per-peer loop so that wide EP multi-iterates safely (barrier consumed and
    reset exactly once regardless of iteration count).
  - **Cross-device barrier**: add a multi-warp-safe wait+reset path via
    `if constexpr (EpIsWideEp(kCfg))` — single-thread wait+reset followed by
    `__syncthreads()` prevents the race where warp 0 resets the barrier before
    warp 1 reads `gridDim.x`. The narrow path is unchanged.
  - **Validation**: relax `EpCfgIsValid` to accept
    `worldSize <= warpPerBlock * waveSize` (one block) via `EpIsWideEp`, and
    cover the new geometry in `test_jit_core`.

### FlyDSL combine kernel (`intranode_kernels.py`)

  - **XDB barrier deadlock fix**: the combine cross-device barrier used a naive
    `grid_thread_id < npes` / `tid < npes` mapping for the signal push and poll.
    At `npes > WAVE` this spills active lanes onto a partially-active sibling
    wave; GPUs give no forward-progress guarantee between sibling waves, so the
    spinning wave0 can starve wave1 before it issues its stores — deadlocking
    peers whose flags come from the sibling wave. Fix: keep both push and poll
    within wave0 using a strided `for p in range(lane, npes, WAVE)` loop.

### DIFF
--- a/include/mori/ops/dispatch_combine_v2/ep_cfg.hpp
+++ b/include/mori/ops/dispatch_combine_v2/ep_cfg.hpp
@@ -301,6 +301,11 @@ inline void EpApplyFields(T& dst, const std::string& prefix, const Has& has, con
 // ---------------------------------------------------------------------------
 constexpr int EpBlockThreads(const EpCfg& c) { return c.warpPerBlock * c.waveSize; }
 
+// True when worldSize exceeds a single wavefront. The per-peer loops in
+// dispatch Phase 2 and the XDB barrier then iterate more than once per lane
+// and need the multi-iteration-safe code path. Compile-time via the Cfg NTTP.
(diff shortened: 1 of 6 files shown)


zhangfei829 (approved):
nice work

PR #615, opened 2026-08-28
docs(skills): [Pre-flight] Add cluster-network-topology skill and scripts

lcskrishna:
Complements `mori check`: where env_check answers "is this host configured correctly?" and drives its peer over SSH, this maps the fabric and localizes a failure to a tier from inside a scheduler allocation, where SSH to compute nodes is often unavailable.

Bundles the probe/matrix/report scripts referenced by the document.

## Submission Checklist

### DIFF
--- a/.claude/skills/cluster-network-topology/make_report.py
+++ b/.claude/skills/cluster-network-topology/make_report.py
@@ -1,0 +1,290 @@
+#!/usr/bin/env python3
+"""Generate a self-contained HTML report from a cluster-topology output folder.
+
+Per-folder report:
+    python3 make_report.py <output_dir> [--title TITLE] [--out report.html]
+
+Combined comparison index across several folders:
+    python3 make_report.py --index <index.html> <dir1> <dir2> ...
+
+The report embeds the diagrams (base64, so the HTML is a single shareable file),
+renders the summary .md, and includes the raw cross-rail result + probe report in
+collapsible sections. No third-party dependencies.
+
+File discovery (by glob within the folder):
+  *node*.png (not crossrail)  -> single-node diagram
+  *crossrail*.png             -> 2-node cross-rail diagram
+  topology_*.md / *.md        -> summary
+  *result*.txt                -> cross-rail result
+  topo_report*.txt            -> probe report
+"""
+import sys, os, re, glob, base64, html, argparse
+
+# ---------- helpers ----------
+def find(folder, pat, exclude=None):
+    hits = sorted(glob.glob(os.path.join(folder, pat)))
+    if exclude:
+        hits = [h for h in hits if exclude not in os.path.basename(h)]
+    return hits[0] if hits else None
+
+def read(path):
+    if not path or not os.path.isfile(path):
+        return ""
+    with open(path, encoding="utf-8", errors="replace") as f:
+        return f.read()
+
+def img_data_uri(path):
+    if not path or not os.path.isfile(path):
+        return ""
+    b = base64.b64encode(open(path, "rb").read()).decode()
+    return "data:image/png;base64," + b
+
+def detect_gpu(text):
+    m = re.search(r"MI\d{3}X(?:\s*VF)?", text)
+    return m.group(0) if m else "GPU"
+
+def parse_result(text):
+    """From a cross-rail result.txt, summarize both experiments as same/cross-rail status.
+    Returns dict: rdma_same, rdma_cross ('OK'/'FAIL'/'-'), ip_same, ip_cross (ok,total)."""
+    rdma_same = rdma_cross = "-"
+    ip_same = [0, 0]; ip_cross = [0, 0]
+    for ln in text.splitlines():
+        s = ln.strip()
+        low = s.lower()
+        if low.startswith("same-rail") and ":" in s:
+            rdma_same = "OK" if ("REACHABLE" in s and "UNREACHABLE" not in s) else ("FAIL" if "UNREACHABLE" in s else "-")
+        elif low.startswith("cross-rail") and ":" in s:
+            if "SKIPPED" in s: rdma_cross = "-"
+            else: rdma_cross = "OK" if ("REACHABLE" in s and "UNREACHABLE" not in s) else ("FAIL" if "UNREACHABLE" in s else "-")
+        else:
+            m = re.match(r"^rail\d+\S*\s+rail\d+\s+(\S+)\s+(same-rail|cross-rail)", s)
+            if m:
+                ok = 1 if m.group(1) == "OK" else 0
+                (ip_same if m.group(2) == "same-rail" else ip_cross)[0] += ok
+                (ip_same if m.group(2) == "same-rail" else ip_cross)[1] += 1
+    return {"rdma_same": rdma_same, "rdma_cross": rdma_cross, "ip_same": ip_same, "ip_cross": ip_cross}
+
+def ip_cell(pair):
+    o, t = pair
+    if t == 0: return ("n/a", "muted")
+    if o == t: return (f"OK ({o}/{t})", "ok")
+    if o == 0: return (f"FAIL (0/{t})", "bad")
+    return (f"mixed ({o}/{t})", "warn")
+
+def rdma_cell(v):
+    return {"OK": ("OK", "ok"), "FAIL": ("FAIL", "bad")}.get(v, ("n/a", "muted"))
+
+def detect_fabric(text):
+    # Prefer the explicit "Fabric classification:" heading so we don't get fooled by
+    # a comparison table that mentions every fabric type.
+    m = re.search(r"Fabric\s+class(?:ification)?:\s*(.+)", text, re.I)
+    scope = m.group(1).lower() if m else text.lower()
+    if "full-mesh" in scope:
+        return ("FULL-MESH", "cross-rail RDMA works (all-to-all / EP supported)", "ok")
+    if "both" in scope or "100% loss" in scope:
+        return ("RAIL-ONLY (IP + RDMA)", "rails fully isolated; no cross-rail at any layer", "bad")
+    if "rail-only" in scope or "rdma" in scope:
+        return ("RDMA RAIL-ONLY", "IP routable cross-rail, but RDMA cross-rail fails", "warn")
+    return ("UNDETERMINED", "see details", "warn")
+
+# ---------- minimal markdown -> HTML ----------
+def md_inline(s):
+    s = html.escape(s)
+    s = re.sub(r"`([^`]+)`", r"<code>\1</code>", s)
+    s = re.sub(r"\*\*([^*]+)\*\*", r"<strong>\1</strong>", s)
+    return s
+
+def md_to_html(md):
+    out, i, lines = [], 0, md.splitlines()
+    while i < len(lines):
+        ln = lines[i]
+        if ln.startswith("```"):
+            buf = []
+            i += 1
+            while i < len(lines) and not lines[i].startswith("```"):
+                buf.append(html.escape(lines[i])); i += 1
+            i += 1
+            out.append("<pre class='code'>" + "\n".join(buf) + "</pre>")
+            continue
+        m = re.match(r"^(#{1,4})\s+(.*)$", ln)
+        if m:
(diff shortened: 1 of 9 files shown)


.claude/skills/cluster-network-topology/xrail_matrix.sh
```diff
@@ -0,0 +1,218 @@
...
+MGMT_NDEV=$(ip route show default 2>/dev/null | awk '/default/{print $5; exit}')
+DEVS=(); declare -A NDEV ADDR
+for d in $(ls "$IB_ROOT" 2>/dev/null | sort -V); do
+  P="$IB_ROOT/$d/ports/1"
+  [ "$(awk '{print $2}' "$P/state" 2>/dev/null)" = ACTIVE ] || continue
+  [ "$(cat "$P/gid_attrs/types/$GID_INDEX" 2>/dev/null)" = "RoCE v2" ] || continue
+  nd=$(cat "$P/gid_attrs/ndevs/$GID_INDEX" 2>/dev/null)
+  [ -n "$nd" ] || continue
+  [ -n "$MGMT_NDEV" ] && [ "$nd" = "$MGMT_NDEV" ] && continue
+  a=$(ip -o -6 addr show "$nd" scope global 2>/dev/null | awk '{print $4}' | cut -d/ -f1 | head -1)
```
jhchouuu: **(1) This makes the script find zero rails on an IPv4 RoCE fabric.**

Only an IPv6 global address is accepted, so every device without one is dropped. On the ionic box above, whose rails are IPv4 (192.168.N.10), I replayed this detection block verbatim:

```
cand ionic_0 ndev=benic1p1 v6global=[]
cand ionic_1 ndev=benic2p1 v6global=[]
... (all eight)
RESULT rails: none  NR=0
```

`probe_topology.sh` in this same PR already handles both IPv4-mapped and IPv6-ULA GIDs, and the skill description advertises exactly that. Could the two scripts share one detection path? Right now they disagree about the same machine.

Same issue one line up at `GID_INDEX=${GID_INDEX:-1}` — probe auto-detects the index, this hardcodes it.

.claude/skills/cluster-network-topology/xrail_matrix.sh
```diff
@@ -0,0 +1,218 @@
...
+  [ -n "$a" ] || continue
+  DEVS+=("$d"); NDEV[$d]=$nd; ADDR[$d]=$a
+done
+NR=${#DEVS[@]}
```
jhchouuu, line 33: **(2) `NR` is computed but never checked**, and the script runs under `set -uo pipefail`. `${DEVS[0]}` / `${DEVS[1]}` are then dereferenced unguarded at lines 72, 103, 110, 118, 179, 181, 188, 191, 199, 203:

```
$ bash -c 'set -uo pipefail; DEVS=(); echo "${DEVS[1]}"'
bash: DEVS[1]: unbound variable    rc=127
```

So a machine this does not understand gets a bare 127 instead of a diagnosis — the opposite of what a pre-flight tool should do. Suggest right here:

```bash
NR=${#DEVS[@]}
[ "$NR" -ge 2 ] || { log "need >=2 active rails, found $NR — nothing to sweep"; exit 0; }
```

.claude/skills/cluster-network-topology/xrail_matrix.sh
```diff
@@ -0,0 +1,218 @@
...
+  for t in "${!DEVS[@]}"; do
+    d=${DEVS[$t]}
+    ( for _ in $(seq 1 200); do
+        timeout 10 ibv_rc_pingpong -d "$d" -g "$GID_INDEX" -p "$((PORT_BASE+t))" -n 20 >/dev/null 2>&1
+      done ) &
+  done
+  # servers for the SL sweep on rail 1 (dest), one per SL
+  for sl in 0 1 2 3 4 5 6 7; do
+    ( for _ in $(seq 1 12); do
+        timeout 10 ibv_rc_pingpong -d "${DEVS[1]}" -g "$GID_INDEX" -p "$((PORT_BASE+40+sl))" -n 20 -l "$sl" >/dev/null 2>&1
```
jhchouuu: First of the unguarded `${DEVS[1]}` dereferences (see the note on line 39). With `set -u` and an empty `DEVS` this aborts the script here, before any phase runs.

.claude/skills/cluster-network-topology/xrail_matrix.sh
```diff
@@ -0,0 +1,218 @@
...
+  touch "$RUNDIR/servers_up.$me"
+fi
+
+for _ in $(seq 1 40); do [ -s "$RUNDIR/addrs.$A" ] && [ -s "$RUNDIR/addrs.$B" ] && break; sleep 1; done
+
+# ---------- Node A: run the sweeps ----------
+if [ "$me" = "$A" ] && mkdir "$RUNDIR/lock" 2>/dev/null; then
+  for _ in $(seq 1 30); do [ -f "$RUNDIR/servers_up.$B" ] && break; sleep 1; done
+  sleep 5
+  R="$RUNDIR/matrix.txt"; : > "$R"
```
jhchouuu, line 171: **(3a) This writes `matrix.txt`, but `make_report.py` globs for `*result*.txt`** and SKILL.md:288 states the output folder contains `result.txt`. Neither matches.

Renaming alone would not fix it — see the note on `make_report.py:60` for the format mismatch.

.claude/skills/cluster-network-topology/make_report.py
```diff
@@ -0,0 +1,290 @@
...
+<meta name="viewport" content="width=device-width,initial-scale=1">
+<title>{html.escape(title)}</title><style>{CSS}</style></head>
+<body><div class="wrap">{body}</div></body></html>"""
+
+# ---------- per-folder report ----------
+def build_report(folder, title=None):
+    md_path = find(folder, "topology_*.md") or find(folder, "*.md")
+    node_png = find(folder, "*node*.png", exclude="crossrail") or find(folder, "*node*.png")
+    xr_png = find(folder, "*crossrail*.png")
+    result = find(folder, "*result*.txt")
```
jhchouuu, line 399: **(3b) Most of what this globs for is produced by nothing in the PR.**

Run against real `probe_topology.sh` output, the generator succeeds but emits a 5.4 KB page with fabric class `UNDETERMINED`, no diagrams, no summary, no cross-rail data:

| looked for | produced by |
|---|---|
| `topo_report*.txt` (line 187) | `probe_topology.sh` |
| `*result*.txt` (line 186) | nothing — `xrail_matrix.sh` writes `matrix.txt` |
| `*node*.png`, `*crossrail*.png` (184-185) | nothing — probe only *prints* the `dot -Tpng` command, never runs it |
| `topology_*.md` (line 183) | nothing |

So the three scripts do not chain; the report is reachable only by hand-authoring its inputs.

.claude/skills/cluster-network-topology/make_report.py
```diff
@@ -0,0 +1,290 @@
...
+    for ln in text.splitlines():
+        s = ln.strip()
+        low = s.lower()
+        if low.startswith("same-rail") and ":" in s:
+            rdma_same = "OK" if ("REACHABLE" in s and "UNREACHABLE" not in s) else ("FAIL" if "UNREACHABLE" in s else "-")
+        elif low.startswith("cross-rail") and ":" in s:
+            if "SKIPPED" in s: rdma_cross = "-"
+            else: rdma_cross = "OK" if ("REACHABLE" in s and "UNREACHABLE" not in s) else ("FAIL" if "UNREACHABLE" in s else "-")
+        else:
+            m = re.match(r"^rail\d+\S*\s+rail\d+\s+(\S+)\s+(same-rail|cross-rail)", s)
```
jhchouuu, line 109: **(3c) Renaming `matrix.txt` to `result.txt` would still parse to nothing.**

This expects `rail0 rail1 OK same-rail` rows, and line 54 expects `same-rail: ... REACHABLE`. What `xrail_matrix.sh` actually writes is a grid:

```
src\dst      r0    r1    r2 ...
r0            OK    --    --
```

Unrelated formats. And the grammar this function requires is not documented anywhere in SKILL.md, so someone following the skill by hand cannot produce a parseable file either.

Either align the filenames *and* the format so `probe → xrail → make_report` chains, or specify the exact `result.txt` grammar in SKILL.md and have `xrail_matrix.sh` emit it.

jhchouuu (commented):
Thanks for putting this together — the methodology in SKILL.md is useful, especially Step 3's tier localization and the discipline of drawing an inferred tier as inferred rather than measured.

I ran the scripts on one of our AMD ionic boxes (8 RoCE rails, MI355X) before commenting. `probe_topology.sh` works: it correctly paired GPU2→ionic_2 and GPU3→ionic_3 even though the netdev names are transposed on that machine (`ionic_2`→`benic4p1`, `ionic_3`→`benic3p1`), because it sorts by PCI address rather than by name. That is the right call and it is not obvious. The mgmt NIC was excluded correctly too.

The other two scripts have three problems; details inline.

lcskrishna:
@jhchouuu Thanks for reviewing the PR. I noticed I haven't fully added finished updating this PR. Let me move it to draft mode and we can re-review again.

.claude/skills/cluster-network-topology/xrail_matrix.sh
lcskrishna: Fixed. Rail discovery now lives in a single shared library, rail_detect.sh, which both xrail_matrix.sh and xrail_worker.sh source (xrail_matrix.sh:26-30) and which probe_topology.sh uses as well — so the three tools can no longer disagree about the same machine.
The library walks every GID entry per device and picks the first global RoCEv2 one, accepting both address families: IPv4-mapped (0000:...:ffff:AABBCCDD, tail decoded to the rail IPv4) and global IPv6 (ULA/GUA), while skipping fe80:: link-local. The GID index is whatever that scan lands on — GID_INDEX is now only an override, not a default of 1.

.claude/skills/cluster-network-topology/xrail_matrix.sh
lcskrishna: Fixed this - NR is now gate

.claude/skills/cluster-network-topology/xrail_matrix.sh
lcskrishna: Resolved by splitting the roles rather than renaming: the new xrail_worker.sh runs one same-rail + one cross-rail RDMA probe plus the full IP ping matrix and writes the result.txt that make_report.py consumes, while xrail_matrix.sh stays the deeper NxN + SL/MTU/TC sweep and keeps matrix.txt as a human-read artifact. SKILL.md:331 documents which worker to reach for when, and the result.txt grammar is now specified there too (see the reply on make_report.py:109), so probe → xrail_worker → make_report chains end-to-end.

.claude/skills/cluster-network-topology/make_report.py
lcskrishna: Fixed — every glob now has a producer, and the chain runs unattended:
 
  | looked for                  | now produced by                                                                                              |
  |-----------------------------|--------------------------------------------------------------------------------------------------------------|
  | topo_report*.txt            | probe_topology.sh, run in-allocation by the sbatch wrappers (host-suffixed, so a 2-node run doesn't collide) |
  | *result*.txt                | xrail_worker.sh (see the reply on line 158)                                                                  |
  | *node*.png, *crossrail*.png | new make_diagrams.py — emits topology_<host>_node.{dot,png} and topology_<A>__<B>_crossrail.{dot,png}        |
 | topology_*.md               | still authored, and still optional — the report no longer needs it                                        |

.claude/skills/cluster-network-topology/make_report.py
lcskrishna: Both halves of the ask are now done — aligned the format and specified the grammar.

.claude/skills/cluster-network-topology/xrail_matrix.sh
lcskrishna: Fixed.

jhchouuu (approved):
Thanks — all six points look addressed. I ran this on one of our 8× MI355X / 8× ionic node and it works.

LGTM

PR #617, opened 2026-08-28
fix(EPv2): align staging pool capacity with EpMaxRecv to prevent metadata loss

kawhil-amd:
## Summary

  `d335c033` replaced the hardcoded `CUSPLIT_MAX_GPUS` (8) with `MORI_EP_WORLD_SIZE` in the staging pool macro. For a 4-GPU configuration this halved
  `_stgCap` from 65536 to 32768.

  However, `_stgCap` is guarded against `ab` — the **absolute** recv-buffer index (cumulative across all peers) — which can reach `EpMaxRecv` (= `worldSize
  × maxTokPerRank` = 65536). When `ab > _stgCap`, two guards silently skip metadata operations:

  - **Write side** (lines 424/446): `if (myDestTokId < _stgCap)` / `if (dt >= _stgCap) continue`
  - **Read side** (lines 520/600): `if (ab + cc > _stgCapM) continue`

  Token payloads are copied unconditionally in Phase 1, so they land correctly — but **indices, weights, scales, and srcmap are silently dropped** for any
  token whose absolute position exceeds `_stgCap`.

  ## Root Cause

  `CUSPLIT_POOL_SLOTS = MORI_EP_WORLD_SIZE * 32768` gives `_stgCap = 32768` for EP4, but `EpMaxRecv = 65536`. The staging pool is undersized relative to the
  recv-buffer capacity it must cover.

  ## Fix

  - Emit `MORI_EP_MAX_RECV` (= `EpMaxRecv(cfg)`) from `RenderEpSource` alongside the existing `MORI_EP_WORLD_SIZE`.
  - Redefine `CUSPLIT_POOL_SLOTS` as `MORI_EP_WORLD_SIZE × MORI_EP_MAX_RECV`.

  This makes `_stgCap = EpMaxRecv`, matching `recvCapM`, so the guards never fire incorrectly.

### DIFF
--- a/src/ops/dispatch_combine_v2/ep_intranode_1250x.hpp
+++ b/src/ops/dispatch_combine_v2/ep_intranode_1250x.hpp
@@ -63,10 +63,14 @@ using index_t = int32_t;
 #define MORI_COMB_BARSLEEP 15
 #define MORI_COMB_BARSPREAD 16
 
-// MORI_EP_WORLD_SIZE is emitted by RenderEpSource before #include-ing this
-// header, so the global arrays below are sized to the exact config.
+// MORI_EP_WORLD_SIZE and MORI_EP_MAX_RECV are emitted by RenderEpSource before
+// #include-ing this header, so the global arrays below are sized to the exact
+// config. The fallbacks are for bare C++ callers only.
 #ifndef MORI_EP_WORLD_SIZE
 #define MORI_EP_WORLD_SIZE 8
+#endif
+#ifndef MORI_EP_MAX_RECV
+#define MORI_EP_MAX_RECV (MORI_EP_WORLD_SIZE * 32768)
 #endif
 
 template <typename T>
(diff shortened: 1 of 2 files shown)


jhchouuu:
LGTM — this matches what we hit on gfx1250: at EP4 / maxTok 16384, `_stgCap` (32768) fell below `EpMaxRecv` (65536), so the `ab + cc > _stgCapM` guards silently dropped metadata for the upper half of the recv buffer. Rolling mori back to `4e801d4a` (where `8 * 32768 / 4` happened to equal `EpMaxRecv`) made the symptom go away, which lines up exactly with your root cause.

One thing worth adding while you're here: the pool is now `worldSize^2 * maxTokPerRank`, and there is no upper bound on it — EP8/16384 is 132 MB per compiled variant, EP16 ~554 MB. Since #606 just relaxed the check to `worldSize <= EpBlockThreads`, a wide config can quietly walk into a multi-GB static allocation. The scale pool right below has a guard for exactly this; mirroring it is cheap:

```cpp
static_assert((size_t)CUSPLIT_POOL_SLOTS * CUSPLIT_MAX_TOPK * 8 <= (size_t)1 << 30,
              "EP staging pool exceeds 1 GiB per compiled variant -- it grows as "
              "world_size^2 * maxTokPerRank; re-index it block-locally before going wider");
```

Not a blocker — happy to see this go in as is.

jhchouuu (approved):

PR #623, opened 2026-08-31
SDMA HIP device management fixes

pemeliya:
## Summary

Key intra-node SDMA queues on the **host-global KFD topology node id** (== `HSA_AGENT_INFO_NODE`) instead of a HIP device ordinal / within-node index. HIP ordinals are only a per-process slice of `HIP_VISIBLE_DEVICES`, and the previous within-node index (`SameHostPeersBefore`, `hipDev`/`lsa` used directly as anvil's device-pair key) only lined up with physical GPUs in the full-8-GPU-visibility case. Under sliced `HIP_VISIBLE_DEVICES` (the JAX-style multi-process launch pattern) those indices diverge from topology, so SDMA queues were wired to the wrong peer. KFD node ids are resolved without HIP/HSA calls (pure sysfs) and exchanged once per rank, so they stay correct regardless of what each process can see.

## What changed

- **`Context`**: `CollectHostNames()` now also allgathers each rank's local KFD node id (resolved via `anvil::kfdNodeIdForHipDevice()`, sysfs-only, no HSA init) alongside the existing pid/rail-flag/nodeId record, stored per-peer as `PeerInfo::kfdNodeId`. New accessors `KfdNodeId(rank)` / `LocalKfdNode()`. Removed `SameHostPeersBefore()` and the `rankInNode` member — `rankInNode` is now read once from `peerInfos[LocalRank()].rankInNode` (already computed during the allgather) instead of being recomputed by a separate loop in `InitializeTopologyAndTransports()`.
- **`EnsureSdmaTransport()`**: connects peers via `anvil::anvil.connect(localNode, peerNode, ...)` keyed on KFD node ids (aborting loudly if either side's node id failed to resolve), and drops the `EnablePeerAccess`/`hipDeviceEnablePeerAccess` call entirely — the SDMA copy engine operates on already-mapped fabric/IPC VAs, so HIP peer access was never required and was the only remaining hard dependency on HIP-visibility-derived device ids.
- **`anvil`** (`AnvilLib`): `sdma_channels_` is now keyed on `(srcNode, dstNode)` node-id pairs rather than HIP device ids. `connect(srcNode, dstNode, numChannels)` and `getSdmaQueue(srcNode, dstNode, channelIdx)` both take node ids now (the old device-id-taking `connect()` overload is gone — there's a single, consistently node-id-keyed API).
  - The engine-selection helpers (`getSdmaEngineId`, `getRecommendedEngineMask`, `getHostLinkEngineMask`, `isGfx1250`, `getOamId`) all now take node ids instead of device ids.
  - `SdmaQueue`'s constructor drops the unused `remoteDeviceId_` member and the HSA-agent-lookup-by-device-id it did internally; it now takes the resolved node id directly.
  - New `nodeForHipDevice()` (via HSA, for single-process callers that only have HIP device ids) and `kfdNodeIdForHipDevice()` (sysfs-only, used by `Context`) statics.
  - `getBusId()` now returns the packed BDF integer directly (previously a hex string that callers re-parsed); `getOamId()` and `kfdNodeIdForHipDevice()` share a new `locIdAndDomainForNode()` sysfs-parsing helper instead of duplicating the `location_id`/`domain` file-scan logic.
  - Fixed an infinite loop in `kfdNodeIdForHipDevice()`: the node-id scan didn't stop when it ran past the last real KFD node (sysfs read failure was silently treated as "keep scanning" rather than "no more nodes"); now breaks out once `locIdAndDomainForNode()` reports an unavailable node.
- **`cco_init.cpp`** (`ccoSdmaSetupCommQueues`) and **`symmetric_memory.cpp`** (`RegisterSymmMemObj`): both switch their `getSdmaQueue()` calls from a within-node index (`comm->hipDev` / `lsa`, or a manually-recounted "SDMA peers before me" index) to `context->LocalKfdNode()` / `context->KfdNodeId(pe)`. The kernel-facing device-handle arrays are unchanged — still indexed by logical rank (`lsaRank`, or global `pe`), since only the *anvil lookup key* was wrong, not the on-device layout.
- **`examples/sdma/{sdma_bw,sdma_rate,sdma_latency,sdma_bw_allgather}.cpp`**: these single-process (and, for `sdma_bw_allgather`, full-visibility MPI) examples now resolve each HIP device id to its KFD node id once via `anvil::anvil.nodeForHipDevice()`, and use the resulting node id consistently for both `connect()` and every `getSdmaQueue()` lookup (previously they mixed the two, which happened to work only because HIP ordinal == KFD node id on this particular box).
- **`tools/run_cco_tests.sh`**: skips `test_sdma_hip_dev_assign` in the default sweep (alongside the other SDMA tests already skipped there), since it takes its own `--n_processes`/`--n_local_devices`/`--visible_devices` flags instead of a bare positional rank count and needs `MORI_ENABLE_SDMA=1`.

## New test coverage

Added `tests/cpp/cco/test_sdma_hip_dev_assign.cpp` — a self-contained CCO SDMA all-gather test (auto-discovered by the existing `test_sdma_*` CMake glob, built only under `BUILD_CCO_SDMA`, self-SKIPs when `MORI_ENABLE_SDMA` is unset) that exercises both real-world multi-process GPU-visibility patterns the node-id keying is meant to unify:

- **jax-style** (default): each of `--n_processes` forked processes slices `HIP_VISIBLE_DEVICES` into a disjoint `--n_local_devices`-sized chunk *before* its first HIP call, then spawns one thread per local device (SPMT), each binding local ordinal `0..n_local_devices-1`. This is the sliced-visibility case where a peer's HIP ordinal doesn't correspond to the same physical GPU across processes — exactly what broke with device-id-based keying.
- **torch-style** (`--torch_style`): every process shares the same (unsliced) visible-device set and binds device ordinal == its own rank via `hipSetDevice(i)`, mirroring a real PyTorch multi-process launch where all GPUs are visible to every process.
- `--visible_devices D0,D1,...` lets either mode run over a custom/non-contiguous physical device order (jax-style: sliced per process, exact length required; torch-style: full list shared unsliced, only needs `len >= n_processes`).

## Testing

- Built `test_sdma_hip_dev_assign` and smoke-tested on an 8-GPU ROCm host (`pemeliya-rocm7.14` container):
  - jax-style defaults, and `--n_processes`/`--n_local_devices`/`--visible_devices` combinations — all ranks PASSED.
  - `--torch_style` alone (auto-detected GPU count), with `--n_processes`, and with `--visible_devices` — all ranks PASSED.
  - `--torch_style --n_processes 99` (exceeds available devices) fails loudly with a clear error instead of hanging or silently clamping.
- Not yet re-validated on multi-node / non-uniform topologies beyond this box.

## Risk / follow-ups

- Touches core SDMA queue wiring (`anvil`, `Context`, `cco_init`, `symmetric_memory`); worth a careful look at the sysfs-based node-id resolution (`kfdNodeIdForHipDevice`/`locIdAndDomainForNode`) on hosts with unusual KFD topology numbering or multi-segment PCI domains.
- `EnablePeerAccess`/`hipDeviceEnablePeerAccess` removal in `EnsureSdmaTransport` is intentional (SDMA doesn't need HIP peer access) but changes a previously "always-on" call, so worth flagging explicitly in review.
- `EnsureSdmaTransport` now `std::abort()`s if a local or peer KFD node id fails to resolve, rather than silently falling back — this is a stricter failure mode than before and should be validated doesn't fire spuriously on other topologies.

### DIFF
--- a/src/application/context/context.cpp
+++ b/src/application/context/context.cpp
@@ -29,7 +29,10 @@
 #include <unistd.h>
 
 #include <algorithm>
+#include <cstdint>
+#include <cstdio>
 #include <cstdlib>
+#include <fstream>
 #include <iostream>
 #include <map>
 #include <string>
@@ -145,6 +148,8 @@ void Context::BuildInitialEndpoints() {
 
 Context::~Context() {}
 
+namespace {
+
 std::string GetLocalIP() {
   struct ifaddrs *ifaddr, *ifa;
   char host[NI_MAXHOST];
@@ -178,6 +183,8 @@ std::string GetLocalIP() {
   return localIP;
 }
 
+}  // namespace
+
 bool Context::CanUseP2P(int destRank) const {
   if (destRank == LocalRank()) {
     return false;  // Cannot use P2P with self
@@ -197,51 +204,50 @@ void Context::CollectHostNames() {
   gethostname(hostname, HOST_NAME_MAX);
   myHostname = std::string(hostname);
 
+  int hipDev = 0;
+  HIP_RUNTIME_CHECK(hipGetDevice(&hipDev));
   // Key co-location on node id, not hostname: identical hostnames would mark
   // cross-node ranks as co-located, over-counting rankInNode (trips assert below).
   std::string nodeId = ResolveNodeId(myHostname);
+  // Allgather a fixed-layout {pid, kfdNodeId, railFlag, nodeId} record.
+  struct Pack {
+    pid_t pid;
+    int32_t kfdNodeId;
+    bool railFlag;
+    char nodeId[256];
+  } my = {
+      .pid = getpid(),
+      // Local GPU's KFD node id (host-global, HIP_VISIBLE_DEVICES-independent). Used
+      // as the stable key for wiring SDMA queues to same-host peers.
+      .kfdNodeId = anvil::anvil.kfdNodeIdForHipDevice(hipDev),
+      .railFlag =
+          env::IsEnvVarEnabled("MORI_ENABLE_RAIL_ONLY") || env::IsEnvVarEnabled("MORI_ENABLE_RAIL"),
+  };
+  snprintf(my.nodeId, sizeof(my.nodeId), "%s", nodeId.c_str());
 
-  // Allgather a fixed-layout {pid, railFlag, nodeId} record.
-  constexpr int kPidSize = sizeof(pid_t);
-  constexpr int kFlagSize = sizeof(uint8_t);
-  constexpr int kStrMax = 256;
-  constexpr int kRecordSize = kPidSize + kFlagSize + kStrMax;
+  std::vector<Pack> global(WorldSize());
+  bootNet.Allgather(&my, global.data(), sizeof(Pack));
 
-  pid_t myPid = getpid();
-  uint8_t myRailFlag =
-      env::IsEnvVarEnabled("MORI_ENABLE_RAIL_ONLY") || env::IsEnvVarEnabled("MORI_ENABLE_RAIL");
-  char localBuffer[kRecordSize] = {};
-  memcpy(localBuffer, &myPid, kPidSize);
-  memcpy(localBuffer + kPidSize, &myRailFlag, kFlagSize);
-  snprintf(localBuffer + kPidSize + kFlagSize, kStrMax, "%s", nodeId.c_str());
-
-  std::vector<char> global(kRecordSize * WorldSize());
-  bootNet.Allgather(localBuffer, global.data(), kRecordSize);
-
-  std::string myNodeId(localBuffer + kPidSize + kFlagSize);
   peerInfos.resize(WorldSize());
   std::map<std::string, int> seenPerNode;
   bool anyRailMismatch = false;
   for (int i = 0; i < WorldSize(); i++) {
-    const char* rec = global.data() + i * kRecordSize;
-    pid_t peerPid;
-    memcpy(&peerPid, rec, kPidSize);
-    uint8_t peerRailFlag;
-    memcpy(&peerRailFlag, rec + kPidSize, kFlagSize);
-    std::string peerNodeId(rec + kPidSize + kFlagSize);
-    peerInfos[i].sameHost = (peerNodeId == myNodeId);
-    peerInfos[i].sameProcess = peerInfos[i].sameHost && (peerPid == myPid);
+    const Pack& other = global[i];
+    std::string peerNodeId(other.nodeId);
+    peerInfos[i].sameHost = (peerNodeId == my.nodeId);
+    peerInfos[i].sameProcess = peerInfos[i].sameHost && (other.pid == my.pid);
     peerInfos[i].rankInNode = seenPerNode[peerNodeId]++;
-    if (peerRailFlag != myRailFlag) {
+    peerInfos[i].kfdNodeId = other.kfdNodeId;
+    if (other.railFlag != my.railFlag) {
       MORI_APP_ERROR("MORI_ENABLE_RAIL_ONLY mismatch: rank {} has {}={}, this rank ({}) has {}", i,
-                     peerRailFlag ? "on" : "off", peerRailFlag, LocalRank(),
-                     myRailFlag ? "on" : "off");
+                     other.railFlag ? "on" : "off", other.railFlag, LocalRank(),
+                     my.railFlag ? "on" : "off");
       anyRailMismatch = true;
     }
     if (LocalRank() == 0) {
-      MORI_APP_TRACE("rank {} nodeId={} pid={} rankInNode={} sameHost={} sameProcess={}", i,
-                     peerNodeId, peerPid, peerInfos[i].rankInNode, peerInfos[i].sameHost,
-                     peerInfos[i].sameProcess);
+      MORI_APP_TRACE("rank {} nodeId={} pid={} rankInNode={} kfdNode={} sameHost={} sameProcess={}",
+                     i, peerNodeId, other.pid, peerInfos[i].rankInNode, other.kfdNodeId,
+                     peerInfos[i].sameHost, peerInfos[i].sameProcess);
     }
   }
   if (anyRailMismatch) {
@@ -278,18 +284,10 @@ bool Context::RailOnlyEligible() const {
 // / Context::IsP2PDisabled() instead of getenv anywhere outside the
 // constructor.
 
-int Context::SameHostPeersBefore(int rank) const {
-  int n = 0;
-  for (int j = 0; j < rank; j++)
-    if (peerInfos[j].sameHost) n++;
-  return n;
-}
-
 void Context::InitializeTopologyAndTransports() {
-  // Find my rank in node
-  for (int i = 0; i <= LocalRank(); i++) {
-    if (peerInfos[i].sameHost) rankInNode++;
-  }
+  // Local rank within this host. Already derived in CollectHostNames from the
+  // exchanged nodeId ordering (seenPerNode), so reuse it instead of recounting.
+  int rankInNode = peerInfos[LocalRank()].rankInNode;
   assert(rankInNode < 8);
 
   // Init rdma context — proxy uses vendor-agnostic IBVerbs, IBGDA uses DirectVerbs
@@ -321,8 +319,7 @@ void Context::InitializeTopologyAndTransports() {
     std::cout << "MORI Topology detection is disabled, use static matching" << std::endl;
     if (!activeDevicePortList.empty()) {
       devicePortId = (rankInNode % activeDevicePortList.size());
-      device = activeDevicePortList[devicePortId].first;
-      portId = activeDevicePortList[devicePortId].second;
+      std::tie(device, portId) = activeDevicePortList[devicePortId];
       rdmaDeviceContext.reset(device->CreateRdmaDeviceContext());
     }
   } else {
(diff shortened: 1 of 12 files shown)


src/application/context/context.cpp
```diff
@@ -197,51 +204,50 @@ void Context::CollectHostNames() {
   gethostname(hostname, HOST_NAME_MAX);
   myHostname = std::string(hostname);
 
+  int hipDev = 0;
+  HIP_RUNTIME_CHECK(hipGetDevice(&hipDev));
   // Key co-location on node id, not hostname: identical hostnames would mark
   // cross-node ranks as co-located, over-counting rankInNode (trips assert below).
   std::string nodeId = ResolveNodeId(myHostname);
+  // Allgather a fixed-layout {pid, kfdNodeId, railFlag, nodeId} record.
+  struct Pack {
```
jhchouuu, line 213: This goes on the wire as raw bytes, and aggregate initialization only value-initializes *members* — the trailing padding (265 bytes of members, 268 with alignment) stays indeterminate. The old `char localBuffer[kRecordSize] = {}` zeroed everything. Nothing reads those bytes, so it is only an MSAN/valgrind nuisance, but `struct Pack { ... } my = {};` before the designated initializers fixes it.

Minor: `pid` could be `int32_t` to match `kfdNodeId`, since the rest of the struct is deliberately fixed-width.

src/application/context/context.cpp
```diff
@@ -197,51 +204,50 @@ void Context::CollectHostNames() {
...
+    pid_t pid;
+    int32_t kfdNodeId;
+    bool railFlag;
+    char nodeId[256];
+  } my = {
+      .pid = getpid(),
+      // Local GPU's KFD node id (host-global, HIP_VISIBLE_DEVICES-independent). Used
+      // as the stable key for wiring SDMA queues to same-host peers.
+      .kfdNodeId = anvil::anvil.kfdNodeIdForHipDevice(hipDev),
```
jhchouuu, line 222: This runs on every `Context` construction, regardless of `sdmaEnabled`. Two small consequences:

- `hipGetDevice()` above pulls HIP init into `CollectHostNames()`, which previously made no HIP call at all.
- On failure `kfdNodeIdForHipDevice()` logs at ERROR level, so a process that never touches SDMA (or a container without `/sys/class/kfd`) gets an error line for something that did not break.

The cost itself is nothing — I measured 7 extra `topology/nodes/*/properties` opens vs baseline with SDMA off — so this is about the init ordering and the log level, not performance. It also cuts against the comment on `EnsureSdmaTransport` ("do it lazily here ... so processes that never touch SDMA don't pay for it"). Resolving lazily, or demoting the error when SDMA is off, would keep that intent.

Also: the description says node ids are resolved "without HIP/HSA calls (pure sysfs)". "No HSA init" is right, but `kfdNodeIdForHipDevice()` -> `getBusId()` -> `hipDeviceGetPCIBusId()` is a HIP call.

src/application/transport/sdma/anvil.cpp
```diff
@@ -37,18 +37,19 @@
...
-    const char* hsa_err_msg;
-    hsa_status_string(s, &hsa_err_msg);
-    throw(std::runtime_error{std::string("HSA error at ") + file + std::string(":") +
-                             std::to_string(line) + std::string(" - ") + hsa_err_msg});
-  }
-};
+namespace {
 
-#define CHECK_HSA_ERROR(cmd) checkHsaError((cmd), #cmd, __FILE__, __LINE__)
+#define CHECK_HSA_ERROR(cmd)                                                               \
```
jhchouuu, line 46: Unbraced `if` in a macro: `if (c) CHECK_HSA_ERROR(f()); else g();` binds the `else` to the macro's `if`. Only one call site today (L283, not inside an `if`), so nothing is broken — but `CHECK_HSAKMT_SUCCESS` right below already uses `do { ... } while (0)`, so the two are inconsistent in the same file.

src/application/transport/sdma/anvil.cpp
```diff
@@ -307,65 +277,78 @@ void AnvilLib::init() {
...
+  return nodeId;
+}
+
+// Resolve the KFD topology node id of the given HIP device WITHOUT initializing HSA.
+/*static*/ int AnvilLib::kfdNodeIdForHipDevice(int hipDev) {
+  uint32_t wantDomain = 0, wantLocId = getBusId(hipDev, &wantDomain);
+  // KFD node ids are contiguous from 0; stop at the first gap.
+  for (int node = 0;; node++) {
+    auto [locId, domain] = locIdAndDomainForNode(node);
+    if (locId == ~0u && domain == ~0u) break;
```
jhchouuu, line 293: `{~0u, ~0u}` means both "no such node" and "node exists but has neither key", so a node missing `location_id`/`domain` would silently truncate the scan rather than be skipped.

Not reachable on current hardware — I checked an 8-GPU gfx950 box and the CPU nodes (0 and 1) do carry `location_id=0 domain=0` — but checking for directory existence would separate the two cases.

The "contiguous from 0" assumption is also load-bearing; a line on why that holds (partitioning, hotplug) would help.

src/application/memory/symmetric_memory.cpp
```diff
@@ -288,11 +283,10 @@ SymmMemObjPtr SymmMemManager::RegisterSymmMemObj(void* localPtr, size_t size, bo
...
                   numDevices * numOfQueuesPerDevice * sizeof(anvil::SdmaQueueDeviceHandle*)));
 
-    for (auto& peer : sdmaPeers) {
-      int dstPe = peer.first;         // global pe -> array index (kernel-facing)
-      int dstDeviceId = peer.second;  // within-node id -> anvil queue key
+    for (int dstPe : sdmaPeers) {
+      int dstNode = context.KfdNodeId(dstPe);  // KFD node id -> anvil queue key
       for (size_t q = 0; q < numOfQueuesPerDevice; q++) {
-        auto* anvilHandle = anvil::anvil.getSdmaQueue(srcDeviceId, dstDeviceId, q)->deviceHandle();
+        auto* anvilHandle = anvil::anvil.getSdmaQueue(srcNode, dstNode, q)->deviceHandle();
```
jhchouuu, line 289: `getSdmaQueue()` returns `nullptr` for an unconnected pair or an out-of-range channel, and this (plus the same line in `cco_init.cpp`) dereferences it unchecked.

Pre-existing, not introduced here, and the new `abort()` in `EnsureSdmaTransport` keeps the two peer sets consistent — but since the key now comes from data exchanged across ranks, a null check with a clear message would beat a segfault. Fine as a follow-up.

src/application/context/context.cpp
```diff
@@ -482,15 +474,24 @@ void Context::EnsureSdmaTransport(int requestedChannels) {
...
-  // (not globalRank % 8, which faults under sliced HIP_VISIBLE_DEVICES).
-  int localDevId = SameHostPeersBefore(LocalRank());
+  // Key SDMA queues on the KFD topology node id (host-global, exchanged in
+  // CollectHostNames), NOT a HIP device ordinal. This is correct even under
+  // sliced HIP_VISIBLE_DEVICES, where a peer GPU is not in this process's HIP
+  // device list.
+  int localNode = LocalKfdNode();
+  if (localNode < 0) {
+    MORI_APP_ERROR("EnsureSdmaTransport: local KFD node id unresolved for rank {}", LocalRank());
+    std::abort();
```
jhchouuu, line 484: Worth a second thought: `abort()` takes down the user's process from library code. Aborting rather than silently wiring the wrong peer is defensible, but the likely trigger is environmental (`/sys/class/kfd` not mounted in a container), where disabling SDMA and falling back would be friendlier. Your call — not blocking.

jhchouuu (approved):
LGTM. Thanks for tracking this one down — the three-way confusion between HIP ordinal, rank-derived within-node index and KFD node id was not obvious, and keying on the node id is the right fix.

Tested on gfx950 (MI355X, HIP 7.14) and gfx1250, baseline at merge-base `93d71b88`, same flags for both trees.

`test_sdma_hip_dev_assign` does not use any API added by this PR, so I backported it onto the baseline to check that it actually fails there:

| config | gfx950 base | gfx950 PR | gfx1250 base | gfx1250 PR |
|---|---|---|---|---|
| jax-style, 2 proc x 2 dev | FAIL | PASS | FAIL | PASS |
| jax-style, 4 proc x 1 dev | FAIL | PASS | FAIL | PASS |
| torch-style, 2 proc | PASS | PASS | PASS | PASS |

Identical failure signature on both arches:

```
[rank 2/4] pid=1850934 hipDev=0 visibleDevices=2
HIP error at anvil.hpp:150 - hipDeviceCanAccessPeer(&canAccess, deviceId, peerDeviceId)
  Code: 101 (invalid device ordinal)
rank 0 failed (status=139)
```

That is `EnablePeerAccess` — the call this PR removes — so the diagnosis in the description is confirmed independently.

No regressions: the seven existing SDMA tests pass on both trees, 8/8 on gfx950 and 4/4 on gfx1250. I did not measure perf (keying change, no data path touched).

One thing worth a comment somewhere: `cap.canSDMA = cap.sameHost` is now load-bearing for correctness. Node ids are only host-unique, so two GPUs on different hosts can share one — it is only safe as an anvil key because SDMA is same-host only. Worth guarding if the planned probe-based `canSDMA` ever lands.

Comments below are all minor; none block.

pemeliya:
Thanks for taking a look! I am going to travel next week, so let us put this PR on hold until the next week.

Issue #557, opened 2026-08-13
[RFC] torch SymmetricMemory allocator over CCO: staged design (pointer array -> LSA window -> scale-out/SDMA)


carlushuang:
## Goal

A host-side, torch-native way to allocate buffers that mori CCO can consume. torch owns the lifecycle (device, stream, caching, `rendezvous`, tensor); CCO owns the window semantics (layout, peer addressing, MR, signals). The allocator is the seam, not a third memory subsystem.

This is the design #544 is being refactored around. Packaging of the resulting module is #549; this issue is only about the model.

### Stage 1 — pointer-array model (what torch symm-mem actually is)

`SymmetricMemory` exposes peers as `buffer_ptrs` / `buffer_ptrs_dev`: an array of per-peer base addresses, one entry per rank. Every shipped backend does this, including the ones whose heap is not scattered at all:

| backend | peer VA derivation | flat reservation underneath? |
|---|---|---|
| CUDA / IPC | `buffer_ptrs[r]` from imported handles | no |
| NVSHMEM | `nvshmem_ptr(p)` → `pe_bases[pe] + off` | yes — `cuMemAddressReserve(npes * heap_size)`, peers at `globalBase + slot*heapSize`, but slots are rotated from `mype+1`, so `pe` is not the slot index and the array is unavoidable |
| NCCL/RCCL | `buffer_ptrs` is empty today; pytorch#192524 is adding LSA peer pointers | yes — `ncclGetLsaPointer(w, off, peer) = add4G(lsaFlatBase, peer*stride4G) + off` |

So stage 1 is the plain torch contract and nothing else: `alloc` / `rendezvous` / `buffer_ptrs(+_dev)` / `rank` / `world_size`, no flat view in the public API, signal pad compiled out while its ops are unimplemented. That makes mori a drop-in backend for any existing torch symm-mem kernel or op.

This costs nothing measurable. A/B of the two addressing modes in the all2all example, same kernel, same window:

| | flat (`base + rank*stride`) | pointer array |
|---|---|---|
| gfx950, 8 ranks, 4 MiB | 1912.0 GB/s | 1909.4 GB/s |
| gfx1250, 4 ranks, 4 MiB | 1714.7 GB/s | 1702.9 GB/s |

### Stage 2 — LSA window model (`base + rank·stride`), reusing CCO's scheme

Rather than invent a third layout, adopt CCO's window verbatim — `include/mori/cco/cco.hpp`, `ccoWindowDevice`:

```
[...]

PR #628, opened 2026-09-01
feat(allocator): back symm_mem.rendezvous() with a CCO window

jhchouuu:
Stage 2 of #557: `symm_mem.rendezvous()` becomes a CCO window registration, so the backend and CCO stop building two peer mappings for the same memory.

Modelled on torch's own NCCL backend, which draws the same lines — `NCCLSymmetricMemory` registers the buffer window at rendezvous, gives the signal pad a separate allocation and window, exposes `get_window()` outside torch's interface, and hands out `ncclDevComm`s lazily from a two-level `group -> key` map.

### What changes

| | before | after |
|---|---|---|
| peer exchange | `FdChannel` over SCM_RIGHTS + a hand-rolled flat span | `ccoWindowRegister`; `buffer_ptrs` are `ccoGetPeerPtr` results |
| signal pad | 9216 B appended to the buffer, **compiled out by default** | its own CCO allocation and window, sized from torch |
| `MORI_SYMM_SIGNAL_PAD` | build flag | gone |
| `barrier()` | raises | `ccoBarrierAll` |
| group resolution | torch's deprecated `GroupInfo` registry | `resolve_process_group()`, as CUDA/NCCL do |
| null peer in `buffer_ptrs` | rejected | allowed; `world_within_direct_access()` answers from the array |
| window / devcomm | not reachable | `window_handle(tensor)`, `dev_comm(group, key=...)` |

`torch.ops.symm_mem.one_shot_all_reduce` now works without a rebuild, and callers no longer have to make the deprecated `enable_symm_mem_for_group()` call. Net **−70 lines** in the backend, and the ROCm 7.14 fd-lifetime hazard leaves the file entirely — CCO already closes its exported fds after `hipMemRelease`.

The DevComm is deliberately **not** created at rendezvous. `ccoDevCommRequirements` is signal counts, QP counts and connection type — properties of the algorithm about to run, not of the buffer — so it is created on first use and cached per `(group, key)`, as NCCL does.

### Ordering hazards found while building this

Each costs a full allocation per cycle and is silent:

- `ccoCommCreate` leaves a tolerated probe failure latched in HIP's sticky error slot, and torch checks that slot around every launch, so the caller's next kernel gets blamed. Consumed with `hipGetLastError()` at our own call site.
- `free()` must drop our handle reference **before** the window: CCO closes the fd it exported in `ccoMemFree`, and on ROCm 7.14 that fd has to outlive every release of the allocation it names. `test_symm_release` catches it.
- The pad is `ccoMemAlloc` memory, so it must not be `hipMemset` — CCO hit this first (`CcoZeroWindowMem`, `cco_init.cpp:406`) and routes its own resource window around it. This would have thrown "out of memory" for a 9216 B buffer on a fabric box, *after* the buffer's collective register, hanging the peers.

### Review fixes

- **Group cache keyed to the PG incarnation.** torch recycles group names — `destroy_process_group()` resets `_world.group_count` — so a name-keyed cache could serve a communicator for a group that no longer exists. Each incarnation gets a fresh Store; a marker in it tells them apart, and a hit also checks rank/world_size.
- **`barrier()` takes the group mutex** like every other call into the communicator.
- **Pad sized from `get_signal_pad_size()`**, not a constant, so `set_signal_pad_size()` cannot hand out a tensor larger than the allocation behind it.
- **One `ReleaseBlock()`** for `free()` and `Shutdown()`, which had drifted into opposite teardown orders.

### Verified

8x gfx950 / ROCm 7.14 / torch 2.12, at 2, 4 and 8 ranks. CI green including `CCO unit test`, where the allocator tests run.

- `tests/python/allocator` — covers `one_shot_all_reduce`, `barrier(0)`, `window_handle` (including on a view), and the DevComm caching rules.
- `all2all_triton.py`, which only walks `buffer_ptrs_dev` — ~1.9 TB/s, unchanged.
- `all2all_lsa.py` and cco example 09 — both take the window and DevComm from the backend now, and lost ~30 lines of bootstrap each.

### Known issues

- **The gfx1250 / fabric path has no coverage.** The `hipMemset` bug above was on that path, so it is "was broken, believed fixed, still untested". Handle-type agreement between `alloc()` and CCO's re-export is also load-bearing and only exercised on the POSIX-fd side.
- **POSIX-fd free order.** CCO names its rendezvous socket after the local slot offset (`cco_init.cpp:1261`), and freeing returns that slot to a per-rank allocator, so ranks freeing in different orders can make a later `rendezvous()` fail with an fd-exchange timeout. torch places no ordering requirement on frees, so this is a real constraint; it belongs in CCO and is documented in the module docstring for now.
- `put_signal`/`wait_signal` raise. They need CCO signals, which now have a DevComm to hang off.
- `barrier()` drains the current stream because `ccoBarrierAll` is a host collective — correct and stronger than torch asks for, but heavier, and unmeasured.
- The locking gap this exposed in CCO is #632. This PR carries its own per-communicator mutex, so it does not depend on it.

### DIFF
--- a/examples/cco/python/09_torch_symm_lsa_all2all/main.py
+++ b/examples/cco/python/09_torch_symm_lsa_all2all/main.py
@@ -31,7 +31,9 @@ whole exchange through CCO, so the two halves of mori meet:
   * ``mori.allocator`` owns the allocation and its lifetime -- it is a torch
     tensor, freed when Python drops it;
   * CCO owns the peer addressing -- ``rendezvous()`` registers the CCO window,
-    and ``window_handle()`` hands it back for the kernel to address through.
+    and ``window_handle()`` hands it back for the kernel to address through;
+    ``dev_comm()`` hands over the matching device communicator, which is what
+    tells the kernel which LSA rank it is.
 
 No ``Communicator`` is built here: the backend keeps one per process group and
 creates it on the first rendezvous. The kernel is Triton (``mori.ir.triton.cco``)
@@ -73,16 +75,21 @@ def _free_port() -> int:
         return int(sk.getsockname()[1])
 
 
-@triton.jit(do_not_specialize=["chunk_elems", "rank_id"])
-def a2a_lsa_kernel(window, send_ptr, chunk_elems, rank_id, BLOCK: tl.constexpr):
+@triton.jit(do_not_specialize=["chunk_elems"])
+def a2a_lsa_kernel(window, dev_comm, send_ptr, chunk_elems, BLOCK: tl.constexpr):
     """One program per (peer, block): push my chunk into that peer's slot."""
     peer = tl.program_id(0)
     blk = tl.program_id(1)
 
+    # lsa_ptr indexes by LSA rank, which is what the DevComm answers. On one node it
+    # equals the world rank; asking the DevComm is what keeps that an implementation
+    # detail rather than an assumption baked into the kernel.
+    my_lsa = cco.DevComm.lsa_rank(dev_comm)
+
     chunk = chunk_elems.to(tl.int64)
     # The peer's window base plus the slot it reserves for me. lsa_ptr is
     # arithmetic on the flat VA, so no pointer table is dereferenced.
-    dst_addr = cco.Window.lsa_ptr(window, peer, rank_id.to(tl.int64) * chunk * 4)
+    dst_addr = cco.Window.lsa_ptr(window, peer, my_lsa.to(tl.int64) * chunk * 4)
     dst = dst_addr.to(tl.pointer_type(tl.int32), bitcast=True)
     src = (
         send_ptr.to(tl.pointer_type(tl.int32), bitcast=True) + peer.to(tl.int64) * chunk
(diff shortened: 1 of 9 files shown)


kawhil-amd (approved):

jhchouuu:
### The fabric path is no longer uncovered

Ran this branch on **j07-7** — gfx1250 x4, ROCm 10.1, torch 2.12.0 — which is the first machine that takes the fabric branch. `ProbeHandleType` picks it there, so everything below exercised code that had never run:

```
HANDLE TYPE: fabric
tests/python/allocator ......................... 2 passed
all2all_triton.py  np=2  handle=fabric   293.7 GB/s  SUCCESS
all2all_triton.py  np=4  handle=fabric   996.1 GB/s  SUCCESS
all2all_lsa.py     np=2                  291.3 GB/s  SUCCESS
all2all_lsa.py     np=4                  983.0 GB/s  SUCCESS
```

Same-box A/B, which is the only comparison that means anything here: pointer array 996.1 vs LSA window 983.0 GB/s at 4 ranks / 4 MiB — no difference, matching what #557 measured on gfx950. Do not compare these against the 1705 GB/s in the example README: that was a different box and a different ROCm.

### One correction to the `hipMemset` finding

The review found that the signal pad is `ccoMemAlloc` memory and must not be `hipMemset`, citing CCO's own `CcoZeroWindowMem` comment: *"On UALink fabric-exportable pools hipMemset returns hipErrorOutOfMemory (observed on ROCm 7.15)"*. I fixed it by zeroing with a host copy.

On this box, **`hipMemset` on a fabric-exportable pool succeeds** — ROCm 10.1, not 7.15. So the fix is still right (a host copy is correct everywhere, and synchronous, which also orders the zeroing against a peer's first write) but I cannot claim to have reproduced the failure it prevents. Someone on a ROCm 7.15 gfx1250 box could.

### What is still not covered

Everything else in the Known issues section stands. In particular the POSIX-fd free-order constraint is untested in either direction — this box is on the fabric path, where it does not apply.

PR #642, opened 2026-09-07
perf(EPv2): hide the dispatch drain read behind the grid barrier on gfx1250

kawhil-amd:
## Summary

- Hide the dispatch drain read (uncached peer-memory round trip to confirm the previous recv-count was consumed) behind the grid barrier spin on gfx1250, so the fabric latency overlaps the wait instead of sitting on the critical path.
- Replace the grid barrier spin with an arrival-ticket scheme: every block draws one ticket via `atomicAdd`, the first arriver draws a second after its drain read completes, and whoever holds the highest ticket (`gridDim.x`) knows both preconditions for the signal store — all blocks arrived and drain observed — without polling.
- Move the inbound wait to the first arriver so it is already parked when peers' signals land, breaking the deadlock cycle (this rank waits on a peer that waits on this rank) by running inbound and outbound on separate tickets/blocks.
- Move `destPeTokenCounter` clearing from the inbound loop to the outbound loop to avoid a race now that inbound runs on a different block.

## Performance

EP4, hidden 7168, topk 6, bf16, 5 runs each:

| tokens | mode  | base      | branch    | delta  |
|--------|-------|-----------|-----------|--------|
| 512    | graph | 44.0 us   | 39.5 us   | -10.2% |
| 512    | eager | 42.2 us   | 37.7 us   | -10.7% |
| 4096   | graph | 138.1 us  | 134.9 us  | -2.3%  |
| 4096   | eager | 136.6 us  | 133.5 us  | -2.3%  |

### DIFF
--- a/src/ops/dispatch_combine_v2/ep_intranode_1250x.hpp
+++ b/src/ops/dispatch_combine_v2/ep_intranode_1250x.hpp
@@ -664,22 +664,35 @@ __device__ void EpDispatch1250xBody(EpArgs args) {
   }
   __syncthreads();
 
-  if (thdId == 0) atomicAdd(args.gridBarrier, 1u);
+  // Tickets on the arrival counter instead of a spin on it: every block draws
+  // one, the first arriver draws a second once its drain read is done, and the
+  // highest draw implies both preconditions of the signal store -- all blocks in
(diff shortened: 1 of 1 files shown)


jhchouuu (approved):

zhangfei829 (approved):
nice work

PR #648, opened 2026-09-08
test(EPv2): emit a JSON result table and per-point clock/power telemetry

jhchouuu:
Two additions to the EP v2 microbenchmark, both aligned with the cross-framework ubench standardisation (AITER / FlyDSL / perftest / mori reporting comparable numbers). They cover requirement 7 (JSON output) and the clock/power half of requirement 1 (frequency & power reporting).

### JSON result table

`_report.print_json_table` writes one single-line `{"name": ..., "rows": [...]}` object at the end of a run, one row per `(backend, mode, m)`, carrying the case arguments alongside the results:

```json
{"name": "mori ep dispatch_combine_v2 summary", "rows": [{"backend": "hip", "mode": "graph", "ep": 4, "m": 512, "hidden": 7168, "topk": 6, "experts_per_rank": 96, "dispatch_dtype": "bf16", "combine_dtype": "bf16", "combine_in": "inplace", "data_init": "norm", "seed": 1234, "iters": 50, "recv_tokens": 1674, "dispatch_us": 53.9, "combine_us": 32.76, "pair_us": 86.67, "dispatch_gbps": 444.8, "combine_gbps": 719.7, "verified": true, "gfx_clk_mhz": 2333.0, "gfx_clk_mhz_min": 2322, "power_w": 1333.0, "smi_status": "ok"}]}
```

This is aiter's `print_json_table` format (`aiter/test_common.py`), record-oriented and on one line so a parent driver can validate and forward it without parsing a human table. It is a copy rather than an import for the same reason `_data.py` is, and it is pure Python rather than pandas -- a benchmark should not need a dataframe to print its own results. A point that fails its correctness check still gets a row, with `err_msg`, so a gap in the sweep cannot be mistaken for a tier nobody ran. `JSON=0` drops the line; the human table is unchanged either way.

### Clock and power telemetry

A latency number means little without the clock it was measured at: a power-limited part throttles and then reports a different time for the same kernel. With `AITER_SMI_MONITOR=1`, each point is replayed under `amdsmi` and the run reports gfx/soc clock, socket power, hotspot temperature, activity and VRAM.

Three things about that window:

* **It is a replay, after warmup and graph capture -- not the timed loop.** `ITERS` pairs are ~10 ms against a 50 ms sampling tick, too short to sample even once. This is also what keeps input generation, compilation and warmup out of the reported clocks, which is what the requirement asks for.
* **The replay count is computed on rank 0 and broadcast.** These kernels barrier across devices, so a per-rank duration loop would leave one rank waiting on a peer that had already finished.
* **Every rank monitors its own GPU; summaries are gathered to rank 0.** The row keeps the min across ranks as well as the mean, because one throttled GPU sets the pair time for all of them and a mean hides it. HIP ordinals are mapped to amdsmi handles by PCIe BDF, not by index -- amdsmi enumerates every GPU in the box while HIP enumerates what `HIP_VISIBLE_DEVICES` left it.

Telemetry never fails the benchmark. If `amdsmi` is missing, or any rank cannot start its monitor, all ranks agree to skip the window and the sweep continues. ROCm ships the binding without a `setup.py`, so the import falls back to `/opt/rocm/share/amd_smi` without leaving that path on `sys.path`.

Env vars, the emitted `AITER_SMI_RESULT` record and `AITER_SMI_OUTPUT_PATH` are aiter's, so `bench_gfx1250_combo.py` can consume mori's rows with the collector it already has.

### Testing

4x MI4xx, EP4, DSv4 shape (hidden 7168, topk 6, 96 experts/rank):

* JSON table verified against the printed human table for every point.
* SMI: 19-20 samples per 1 s window on **every** rank, `sample_status: ok`, no contention between the four concurrent samplers. Clocks agreed across ranks to within 1% (2322-2345 MHz) and power tracked the token count (1.33 kW at M=512, 1.49 kW at M=4096) -- the window is measuring the replay, not idle.
* Both backends (`flydsl`, `hip`), both modes (`eager`, `graph`), all three dispatch wires (`bf16`, `fp8`, `fp4`), with and without telemetry. All points passed their correctness gate.
* `black==25.1.0` and `ruff==0.12.0` at the repo's pinned config, clean.

### DIFF
--- a/tests/python/ops/dispatch_combine_v2/bench_ep.py
+++ b/tests/python/ops/dispatch_combine_v2/bench_ep.py
@@ -46,14 +46,14 @@ scales and recv-cap, across both backends.
 Two machine-readable outputs sit alongside the human table. A final JSON line in
 aiter's print_json_table shape carries every point, one row per (backend, mode,
 m), so a parent driver reads results instead of parsing columns. And with
-AITER_SMI_MONITOR=1 each point also replays under amdsmi: the clocks a number was
+MORI_SMI_MONITOR=1 each point also replays under amdsmi: the clocks a number was
 measured at belong with the number, since a throttled part reports a different
 time for the same kernel. That replay is a window of its own, AFTER warmup and
 graph capture, because ITERS pairs are ~10 ms against a 50 ms sampling tick.
 
     torchrun --standalone --nproc_per_node=8 bench_ep.py
     BACKENDS=flydsl,hip SWEEP=512,4096 ITERS=200 torchrun ... bench_ep.py
-    AITER_SMI_MONITOR=1 AITER_SMI_DURATION=1.0 torchrun ... bench_ep.py
+    MORI_SMI_MONITOR=1 MORI_SMI_DURATION=1.0 torchrun ... bench_ep.py
 """
 
 import math
--- a/tests/python/ops/dispatch_combine_v2/bench_ep.py
+++ b/tests/python/ops/dispatch_combine_v2/bench_ep.py
@@ -43,12 +43,23 @@ that computes garbage never gets a bandwidth number. That check is deliberately
 one invariant, not a matrix -- test_op.py owns dtypes, quant, scatter, StdMoE,
 scales and recv-cap, across both backends.
 
+Two machine-readable outputs sit alongside the human table. A final JSON line in
+aiter's print_json_table shape carries every point, one row per (backend, mode,
+m), so a parent driver reads results instead of parsing columns. And with
+AITER_SMI_MONITOR=1 each point also replays under amdsmi: the clocks a number was
+measured at belong with the number, since a throttled part reports a different
+time for the same kernel. That replay is a window of its own, AFTER warmup and
+graph capture, because ITERS pairs are ~10 ms against a 50 ms sampling tick.
+
     torchrun --standalone --nproc_per_node=8 bench_ep.py
     BACKENDS=flydsl,hip SWEEP=512,4096 ITERS=200 torchrun ... bench_ep.py
+    AITER_SMI_MONITOR=1 AITER_SMI_DURATION=1.0 torchrun ... bench_ep.py
 """
 
+import math
 import os
 import sys
+import time
 
 import torch
 import torch.distributed as dist
@@ -57,6 +68,8 @@ import mori.cco as cco
 from mori.ops.dispatch_combine_v2 import EpDispatchCombineConfig, EpDispatchCombineOp
 
 import _data
+import _report
+import _smi
 
 HIDDEN = int(os.environ.get("HIDDEN", 7168))
 TOPK = int(os.environ.get("TOPK", 8))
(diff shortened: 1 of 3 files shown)


tests/python/ops/dispatch_combine_v2/bench_ep.py
```diff
@@ -290,8 +309,128 @@ def time_pairs(mode, one_pair, capture):
...
+            if "power_w" in r["metrics"]
+        ]
+        out = {}
+        if med:
+            out["gfx_clk_mhz"] = round(sum(med) / len(med), 1)
+            out["gfx_clk_mhz_min"] = round(min(med), 1)
+        if pwr:
+            out["power_w"] = round(sum(pwr) / len(pwr), 1)
+        if records:
+            out["smi_status"] = (
```
yujiaoliang: `sample_status` currently counts timestamp-only samples as successful. `_collect_sample()` suppresses every metric-query failure and still appends a sample, so a rank can report `ok` with no clock or power data. `clocks()` then silently excludes that rank when computing `gfx_clk_mhz_min`, which can hide exactly the throttled or unreadable GPU this telemetry is meant to detect. Could we require a sufficient number of valid samples for the required metric(s), and propagate a missing-metrics status if any rank has none?

tests/python/ops/dispatch_combine_v2/bench_ep.py
```diff
@@ -350,6 +495,28 @@ def eager_legs():
...
+                    rows.append(
+                        dict(
+                            case(ct, name, mode),
+                            recv_tokens=round(float(got[2]) / n),
+                            dispatch_us=round(float(got[0]) / n, 2),
+                            combine_us=round(float(got[1]) / n, 2),
+                            pair_us=round(float(got[0] + got[1]) / n, 2),
+                            dispatch_gbps=round(d_bw, 1),
+                            combine_gbps=round(c_bw, 1),
+                            verified=bool(was_checked),
```
yujiaoliang: For FP4, and for initializers where the combine check is vacuous, `checked` is false but `prime()` returns `was_checked=True` as long as `CHECK=1`. The JSON row therefore marks `verified: true` even though the reported `combine_us` path was never correctness-checked. The human summary explains “dispatch bytes only”, but that information is lost in the machine-readable output. Please encode the verification scope explicitly, or keep `verified` false unless the complete reported pair was checked.

tests/python/ops/dispatch_combine_v2/bench_ep.py
```diff
@@ -350,6 +495,28 @@ def eager_legs():
...
+                lockstep()
+                if rank == 0:
+                    rows.append(
+                        dict(
+                            case(ct, name, mode),
+                            recv_tokens=round(float(got[2]) / n),
+                            dispatch_us=round(float(got[0]) / n, 2),
+                            combine_us=round(float(got[1]) / n, 2),
+                            pair_us=round(float(got[0] + got[1]) / n, 2),
+                            dispatch_gbps=round(d_bw, 1),
```
yujiaoliang, line 555: `dispatch_gbps` and `combine_gbps` are computed from this rank’s local `total` and event times before `got` is reduced, while the JSON row stores cross-rank averages for `recv_tokens` and latency. On rank 0 this can produce a row whose bandwidth cannot be derived from its own reported inputs, especially because receive counts vary with routing. Could we compute the bandwidth from the reduced values, or reduce the per-rank bandwidth explicitly and name the aggregation?

jhchouuu:
All three are right, and all three are cases where the row read healthier than the run behind it. Fixed in 7034f21f.

**Bandwidth from mixed aggregation.** Correct, and this one was a real defect I introduced: `d_bw` used this rank's `total` and event times, then went into a row whose every other column was a cross-rank mean. As you say, receive counts vary with routing, so the row's `dispatch_gbps` did not follow from its own `recv_tokens` and `dispatch_us`. Both bandwidths now come from the reduced values, which the human table prints too, so the two outputs agree. Checked by hand: 1674 x 7168 x 2 B / 53.95 us = 445 GB/s, which is what the row reports.

**`verified` scope.** Also correct — `prime()` returns `was_checked=bool(CHECK)` on the not-checked path, so fp4 and an all-zero payload claimed `verified: true` next to a `combine_us` nothing had compared. Rather than force it false (which would lose the fact that the dispatch bytes *were* checked byte for byte, the only check fp4 can pass at all), it is now a scope token:

    none | dispatch_bytes | dispatch_bytes+combine

Measured: `dispatch_bytes` for `DISP=fp4`, `dispatch_bytes+combine` for bf16, `none` for `CHECK=0`.

**`sample_status` counting timestamp-only samples.** Correct, and the consequence you point at is the worse half — `clocks()` dropped an unreadable rank from `gfx_clk_mhz_min` silently, narrowing the min to the healthy GPUs, which is exactly the rank this telemetry exists to surface. Changed:

* status is driven by samples that actually carry `gfx_clk_mhz`, not by samples that exist;
* a rank with none reports `no_metrics` (new state, distinct from `insufficient`);
* the **worst** rank's status reaches the row, not "all ok or else insufficient";
* the row gains `smi_ranks`, e.g. `3/4`, so a narrowed min is visible in the row itself;
* the per-rank record gains `clock_samples` next to `samples`.

`_collect_sample` still swallows the failed query — one bad read must not kill the polling thread — so the fix is in how the samples are counted, not in letting it raise.

Verified by fault injection rather than by reading the code: a permanent metric-read failure on rank 2, on 4x MI4xx EP4. That rank now reports `clock_samples: 0` / `no_metrics` and the row carries `smi_status: no_metrics`, `smi_ranks: "3/4"`. Before the change the same run reported `sample_status: ok` on all four ranks.

PR #652, opened 2026-09-09
feat(ccl): support HIP graph capture in AllGather SDMA

alanshwang:
Add a `capturing` flag to `AllgatherSdma::finish_sync` and `finish_async_wait` (threaded through the Python wrapper and pybind bindings) that skips the host-side stream/device syncs. Those syncs only make the returned duration meaningful and are illegal during graph capture; the SDMA transfer, the wait kernel and the copy-out are already stream-ordered, so a consumer enqueued on the same stream still sees complete data, and the launch thread stays free to queue work into the window where the copy engines are busy but the CUs are idle.

With the host sync gone, the cross-GPU visibility it incidentally provided must be re-established inside the wait kernel: the flags are written by a peer GPU's SDMA engine, so a device-scope load does not order the payload behind them. Switch the flag spin to SYSTEM-scope relaxed loads, add a seq-cst SYSTEM acquire load after the spin, and emit a tail `__threadfence_system()` before retire so the consumer kernel reads the peer's payload rather than a stale cache copy.

Default `capturing=false` preserves the existing synchronous contract.

## Submission Checklist

### DIFF
--- a/src/collective/core/oneshot_allgather_sdma_class.cpp
+++ b/src/collective/core/oneshot_allgather_sdma_class.cpp
@@ -430,28 +430,23 @@ int64_t AllgatherSdma<T>::prepare_sync_param_contiguous(T* input, T* output, siz
 }
 
 template <typename T>
-double AllgatherSdma<T>::finish_sync(T* output, size_t total_count, hipStream_t stream) {
-  if (stream != nullptr) {
-    hipError_t err = hipStreamSynchronize(stream);
+double AllgatherSdma<T>::finish_sync(T* output, size_t total_count, hipStream_t stream,
+                                     bool capturing) {
+  // See the header: the fused kernel already blocks on every peer's flag
+  // before it retires, so these syncs are not what makes the result visible.
+  if (!capturing) {
+    hipError_t err = stream ? hipStreamSynchronize(stream) : hipDeviceSynchronize();
     if (err != hipSuccess) {
-      fprintf(stderr, "PE %d: Stream synchronization failed: %s\n", myPe_, hipGetErrorString(err));
-      throw std::runtime_error("Stream synchronization failed");
-    }
-  } else {
-    hipError_t err = hipDeviceSynchronize();
-    if (err != hipSuccess) {
-      fprintf(stderr, "PE %d: Device synchronization failed: %s\n", myPe_, hipGetErrorString(err));
-      throw std::runtime_error("Device synchronization failed");
+      fprintf(stderr, "PE %d: Synchronization failed: %s\n", myPe_, hipGetErrorString(err));
+      throw std::runtime_error("Synchronization failed");
     }
   }
 
   bool direct = find_registered(output).first.IsValid();
   if (!direct && copy_output_to_user_) {
     copy_output_to_user(output, total_count, stream);
-    if (stream != nullptr) {
-      (void)hipStreamSynchronize(stream);
-    } else {
-      (void)hipDeviceSynchronize();
+    if (!capturing) {
+      (void)(stream ? hipStreamSynchronize(stream) : hipDeviceSynchronize());
     }
   }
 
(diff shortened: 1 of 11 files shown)


src/collective/core/oneshot_allgather_sdma_class.cpp
```diff
@@ -430,28 +430,23 @@ int64_t AllgatherSdma<T>::prepare_sync_param_contiguous(T* input, T* output, siz
...
-    hipError_t err = hipStreamSynchronize(stream);
-    if (err != hipSuccess) {
-      fprintf(stderr, "PE %d: Stream synchronization failed: %s\n", myPe_, hipGetErrorString(err));
-      throw std::runtime_error("Stream synchronization failed");
-    }
-  } else {
-    hipError_t err = hipDeviceSynchronize();
+double AllgatherSdma<T>::finish_sync(T* output, size_t total_count, hipStream_t stream,
+                                     bool capturing) {
+  // See the header: the fused kernel already blocks on every peer's flag
```
wuyl1, line 449: Thanks for taking this on — making AllGather SDMA capturable is worth doing, and
the kernel-side memory-ordering change looks right.

I think this premise doesn't hold under replay, though. The kernel does block on
every peer's flag, but which value it waits for is decided on the host — at
line 386 in this file:

```cpp
uint64_t flag_token = call_seq_.fetch_add(1, std::memory_order_relaxed) + 1;
jit_args_.flagVal = flag_token;
```

`flagVal` is passed by value, so capture freezes it into the graph node. The
flags are monotonic AMO_SET and never reset, so from the second replay onward
`flags[peer] < flagVal` is already satisfied on entry and the wait returns
immediately — the consumer reads a buffer whose SDMA is still in flight. Silent
data corruption, not a crash.

The same reasoning applies to the `finish_async_wait` comment below.

`AllreduceSdma` derives the generation on the device instead
(`twoshot_sdma_kernel.hpp:204`, `s_next = barrier->flag + 1ULL`). Could AllGather
do the same?

wuyl1 (changes requested):

src/collective/core/oneshot_allgather_sdma_class.cpp
alanshwang: You're right, and thanks for catching it. Confirmed before fixing — reverting to
this branch's commit and running a capture + 5-replay AllGather with injected rank
skew reproduces exactly what you described:

```
  replay 1: ok
PE 0: replay 2 MISMATCH
    replay 2 chunk from PE 1: want 17838, got [9919]   <-- PREVIOUS replay's value
    replay 2 chunk from PE 2: want 18838, got [10919]  <-- PREVIOUS replay's value
    replay 2 chunk from PE 3: want 19838, got [11919]  <-- PREVIOUS replay's value
    ... PE 4 through PE 7 identical
```

First replay fine, everything from the second onward reads the previous round's
payload, no crash. Every peer is stale, not just the skewed ones: with the token
frozen the comparison is already satisfied regardless of arrival order.

**Fix — device-derived generation, as you suggested.** `CclAllgatherArgs::flagVal`
becomes `uint64_t* genCounter` (same 8 bytes, so the struct layout is unchanged).
The kernel derives `*genCounter + 1` itself and stores it back once every peer's
flag is in. `call_seq_` and `async_flag_token_` are gone, so what capture freezes
into the node is the pointer and every replay derives a fresh value. This is what
`AllreduceSdma` already does (`twoshot_sdma_kernel.hpp:204`). The counter gets its
own allocation to keep it off the cacheline the peers `AMO_SET` into, and
`resetFlags()` now clears both — a zeroed counter beside surviving flag values
would reintroduce the same skip.

Applied to all three kernels that derive the generation: the fused sync gather, the
param-contiguous gather, and the async wait. The param-contiguous variants reach
the same `finish_sync`/`wait_async` and had the identical problem; I also added the
`capturing` parameter that was missing from `enqueue_param_contiguous`, which was
the last AllGather entry point still forcing a host-side stream sync.

The subgroup and broadcast classes still mint a host-side token, but they have no
`capturing` flag on their APIs so no capture path reaches them today — happy to
convert them in a follow-up.

**Regression test** — new file `tests/python/ccl/test_allgather_graph_capture.py`.
It captures one AllGather into a HIP graph and replays it, writing a fresh input
value before each replay, so a stale read surfaces as the previous replay's value
rather than as a crash. Three modes cover the three kernels above.

Verified on 8× gfx950, `MORI_ENABLE_SDMA=1` — same test file against both builds:

| mode | before the fix | after |
|---|---|---|
| `sync` | fails from replay 2, all 7 peers stale | 8/8 pass |
| `async` | fails from replay 2, all 7 peers stale | 8/8 pass |
| `param_contiguous` | n/a, `capturing` is new on this entry point | 8/8 pass |

Also passing after the change: 20 replays at 64 MB/PE in async mode;
`test_allgather` (375 GB/s, unchanged); `test_allgather_overlap`;
`test_allgather_param_contiguous`.

One note for anyone running the test: the injected rank skew is deliberate. A
capture/replay test without it can pass against a knowingly broken build, because
all ranks finish their puts within microseconds of each other and a wait that
returns immediately still finds the data there.

wuyl1 (approved):
LGTM

PR #654, opened 2026-09-09
Add DMA buffer support to mlx5

avbokovoy:
## Summary
- Add dmabuf-based registration for mlx5 GPU control-ring memory (CQ, WQ, DBR umems and atomic ibuf MR), removing the hard dependency on peer-mem kernel modules — enables mlx5 on KVM guests and hosts without peer-mem
- Load `mlx5dv_devx_umem_reg_ex` optionally at runtime via dlopen, falling back silently on older rdma-core
- Expose `TryExportDmabufFd` from rdma.cpp so the mlx5 provider can share the HSA-based dmabuf export logic
- Controlled by `MORI_MLX5_CONTROL_DMABUF` env var: `auto` (default, try dmabuf then peermem), `force` (dmabuf or abort), `off` (peermem only)

## Test plan
- [x] Existing mlx5 internode CI bench (`v1`/`v1_ll` kernels, 64–4096 tokens)
- [x] Verify on a host without peer-mem that QP creation succeeds via dmabuf path
- [x] Verify `MORI_MLX5_CONTROL_DMABUF=off` forces the peermem-only path
- [x] Verify `MORI_MLX5_CONTROL_DMABUF=force` aborts cleanly when dmabuf is unavailable


QizhouZhang97:
Thanks for tackling this — dropping the hard peer-mem dependency for the mlx5 control rings is the right direction, and making `mlx5dv_devx_umem_reg_ex` an optional `dlsym` with a silent miss is the right call for older rdma-core. A few things I think need fixing before this lands.

**1. `force` mode is bypassed when `mlx5dv_devx_umem_reg_ex` is missing.** In `Mlx5RegisterControlUmem` the abort lives *inside* the guard that requires the symbol:

```cpp
if (mode != kOff && Mlx5DvApi::Instance().devx_umem_reg_ex != nullptr) {
    ...
    if (mode == kForce) { std::abort(); }   // only reachable when reg_ex exists
}
devx_umem_reg(...);                         // unconditionally reached when reg_ex == nullptr
```

`devx_umem_reg_ex` is resolved with `warn_on_miss=false` and is (correctly) not part of the `&&` chain in `Mlx5DvApi::Load()`, so `nullptr` is a normal runtime state on older libmlx5. There, `MORI_MLX5_CONTROL_DMABUF=force` silently falls back to peermem instead of aborting — and "rdma-core too old to have `umem_reg_ex`" is exactly the case where the user wants the loud failure. Please hoist the `force` check out of the guard and log why dmabuf was unavailable (missing symbol vs. failed export vs. failed registration).

**2. `struct mlx5dv_devx_umem_in` is only defined in the stub branch.** `mlx5_dv.h` defines it under the `#else` of `__has_include(<infiniband/mlx5dv.h>)`, so a host with an *older* `mlx5dv.h` (present, but without `mlx5dv_devx_umem_in` / `MLX5DV_UMEM_MASK_DMABUF`) won't compile, since `mlx5.cpp` assigns into the complete type. The forward declaration in `dv_loader.hpp` doesn't help — you can't assign through a pointer-only forward decl. Suggest guarding the struct and enum with `#ifndef MLX5DV_UMEM_MASK_DMABUF` and placing them where both `__has_include` branches see them.

**3. `pgsz_bitmap` is a log2 bitmap, not a page size.** `in.pgsz_bitmap = sysconf(_SC_PAGESIZE)` asks for "4K only" on a 4K-page host — it works, but only because the value happens to equal `1 << 12`. That is strictly narrower than the peermem path it replaces: rdma-core's `mlx5dv_devx_umem_reg` passes `UINT64_MAX & ~(MLX5_ADAPTER_PAGE_SIZE - 1)` internally (`providers/mlx5/verbs.c`), i.e. all page sizes >= 4K. On a 64K-page arm64 kernel it becomes "64K only", where an 8-byte DBR umem almost certainly can't be registered, so `auto` falls back to peermem — which on a KVM guest doesn't exist, so QP creation fails outright. That's the very scenario this PR enables. Note also that `attr_optional(pgsz_bitmap)` in `__mlx5dv_devx_umem_reg_ex` is applied only on the `legacy` path, so via `_ex` the value is mandatory and kernel-enforced. `MLX5_ADAPTER_PAGE_SHIFT` is already available via `mlx5_prm.hpp`:

```cpp
// log2(page-size) bitmap, not a page size. Mirror what devx_umem_reg passes
// internally so the dmabuf path accepts the same layouts as the peermem path.
in.pgsz_bitmap = ~((1ULL << MLX5_ADAPTER_PAGE_SHIFT) - 1);
```

I'd avoid plain `UINT64_MAX` — it also sets bits 0-11 (sub-4K), which rdma-core deliberately masks off.

**4. `errno` is clobbered before it is (not) logged.** The failure path logs no `errno` at all, and just adding it wouldn't work either since `close(dmabufFd)` runs between the call and the log. Please capture it right after the call (`int err = errno;`) and include it. This compounds with #3: a rejected `pgsz_bitmap` returns `EINVAL`, but all you'd see is an unexplained "failed" followed by a silent downgrade to peermem.

**5. The peermem fallback logs success without checking for `NULL`.** `devx_umem_reg` returns `NULL` on a host without peer-mem, and the trace still says "registered via peermem". The caller's `assert` catches it today only because `CMakeLists.txt` overrides `CMAKE_CXX_FLAGS_RELEASE` to a bare `-O3` and thus drops CMake's default `-DNDEBUG`; any build that does define `NDEBUG` turns this into a `cqUmem->umem_id` null dereference. A null check with an `errno`-carrying `MORI_APP_ERROR` ("both dmabuf and peermem failed; is a peer-mem module loaded?") would be far more actionable. Same for the `ibv_reg_mr` fallback in `Mlx5RegisterMrDmabuf`.

Two smaller notes: the default `auto` mode WARNs five times per GPU endpoint (cq, cq_dbr, wq, wq_dbr, atomic_ibuf) on hosts without dmabuf export, which reads like a failure for what is an expected fallback — `TRACE` or a once-per-process flag would be quieter. And `MORI_MLX5_CONTROL_DMABUF` reads as control-ring-only but also gates the `atomic_ibuf` payload MR including its `force` abort, while `MORI_ENABLE_DMABUF_REG` already exists for the payload side; worth renaming or splitting, and documenting how the two interact.

avbokovoy:
> Thanks for tackling this — dropping the hard peer-mem dependency for the mlx5 control rings is the right direction, and making `mlx5dv_devx_umem_reg_ex` an optional `dlsym` with a silent miss is the right call for older rdma-core. A few things I think need fixing before this lands.
> 
> **1. `force` mode is bypassed when `mlx5dv_devx_umem_reg_ex` is missing.** In `Mlx5RegisterControlUmem` the abort lives _inside_ the guard that requires the symbol:
> 
> ```c++
> if (mode != kOff && Mlx5DvApi::Instance().devx_umem_reg_ex != nullptr) {
>     ...
>     if (mode == kForce) { std::abort(); }   // only reachable when reg_ex exists
> }
> devx_umem_reg(...);                         // unconditionally reached when reg_ex == nullptr
> ```
> 
> `devx_umem_reg_ex` is resolved with `warn_on_miss=false` and is (correctly) not part of the `&&` chain in `Mlx5DvApi::Load()`, so `nullptr` is a normal runtime state on older libmlx5. There, `MORI_MLX5_CONTROL_DMABUF=force` silently falls back to peermem instead of aborting — and "rdma-core too old to have `umem_reg_ex`" is exactly the case where the user wants the loud failure. Please hoist the `force` check out of the guard and log why dmabuf was unavailable (missing symbol vs. failed export vs. failed registration).
> 
> **2. `struct mlx5dv_devx_umem_in` is only defined in the stub branch.** `mlx5_dv.h` defines it under the `#else` of `__has_include(<infiniband/mlx5dv.h>)`, so a host with an _older_ `mlx5dv.h` (present, but without `mlx5dv_devx_umem_in` / `MLX5DV_UMEM_MASK_DMABUF`) won't compile, since `mlx5.cpp` assigns into the complete type. The forward declaration in `dv_loader.hpp` doesn't help — you can't assign through a pointer-only forward decl. Suggest guarding the struct and enum with `#ifndef MLX5DV_UMEM_MASK_DMABUF` and placing them where both `__has_include` branches see them.
> 
> **3. `pgsz_bitmap` is a log2 bitmap, not a page size.** `in.pgsz_bitmap = sysconf(_SC_PAGESIZE)` asks for "4K only" on a 4K-page host — it works, but only because the value happens to equal `1 << 12`. That is strictly narrower than the peermem path it replaces: rdma-core's `mlx5dv_devx_umem_reg` passes `UINT64_MAX & ~(MLX5_ADAPTER_PAGE_SIZE - 1)` internally (`providers/mlx5/verbs.c`), i.e. all page sizes >= 4K. On a 64K-page arm64 kernel it becomes "64K only", where an 8-byte DBR umem almost certainly can't be registered, so `auto` falls back to peermem — which on a KVM guest doesn't exist, so QP creation fails outright. That's the very scenario this PR enables. Note also that `attr_optional(pgsz_bitmap)` in `__mlx5dv_devx_umem_reg_ex` is applied only on the `legacy` path, so via `_ex` the value is mandatory and kernel-enforced. `MLX5_ADAPTER_PAGE_SHIFT` is already available via `mlx5_prm.hpp`:
> 
> ```c++
> // log2(page-size) bitmap, not a page size. Mirror what devx_umem_reg passes
> // internally so the dmabuf path accepts the same layouts as the peermem path.
> in.pgsz_bitmap = ~((1ULL << MLX5_ADAPTER_PAGE_SHIFT) - 1);
> ```
> 
> I'd avoid plain `UINT64_MAX` — it also sets bits 0-11 (sub-4K), which rdma-core deliberately masks off.
> 
> **4. `errno` is clobbered before it is (not) logged.** The failure path logs no `errno` at all, and just adding it wouldn't work either since `close(dmabufFd)` runs between the call and the log. Please capture it right after the call (`int err = errno;`) and include it. This compounds with #3: a rejected `pgsz_bitmap` returns `EINVAL`, but all you'd see is an unexplained "failed" followed by a silent downgrade to peermem.
> 
> **5. The peermem fallback logs success without checking for `NULL`.** `devx_umem_reg` returns `NULL` on a host without peer-mem, and the trace still says "registered via peermem". The caller's `assert` catches it today only because `CMakeLists.txt` overrides `CMAKE_CXX_FLAGS_RELEASE` to a bare `-O3` and thus drops CMake's default `-DNDEBUG`; any build that does define `NDEBUG` turns this into a `cqUmem->umem_id` null dereference. A null check with an `errno`-carrying `MORI_APP_ERROR` ("both dmabuf and peermem failed; is a peer-mem module loaded?") would be far more actionable. Same for the `ibv_reg_mr` fallback in `Mlx5RegisterMrDmabuf`.
> 
> Two smaller notes: the default `auto` mode WARNs five times per GPU endpoint (cq, cq_dbr, wq, wq_dbr, atomic_ibuf) on hosts without dmabuf export, which reads like a failure for what is an expected fallback — `TRACE` or a once-per-process flag would be quieter. And `MORI_MLX5_CONTROL_DMABUF` reads as control-ring-only but also gates the `atomic_ibuf` payload MR including its `force` abort, while `MORI_ENABLE_DMABUF_REG` already exists for the payload side; worth renaming or splitting, and documenting how the two interact.

@QizhouZhang97 Thanks for your feedback. Addressed your comments:
1. Fixed in https://github.com/ROCm/mori/pull/654/commits/c53516d22ee9ea2a6c9e6531ad3e6f93386984b1
2. For that case, I just put everything in standalone structure outside of the guards in https://github.com/ROCm/mori/pull/654/commits/037dad7d4aa51292af1ed3bf7b2a8d5d2b9a8516
3. Fixed in https://github.com/ROCm/mori/pull/654/commits/4320f26ca26f0d187f78c0d4ba4938394c924283
4. Fixed in https://github.com/ROCm/mori/pull/654/commits/70bcfae81770fd13c3333889863570e82427f413
5. Fixed in https://github.com/ROCm/mori/pull/654/commits/a8a5dc35b22a884dbc82c57e8c00dfe0a433f6f3

Regarding your notes, the WARN-TRACE is addressed in https://github.com/ROCm/mori/pull/654/commits/47b1a7e911420181bad6856f2ccad62778ebc7ed, env variable rename - in https://github.com/ROCm/mori/pull/654/commits/75979ad164ce8bb0cc9378e3eb204a0601b97c71. However, I'm not sure that the last one reflects your initial intention, so may forth having a look at that once again.

Let me know if there is anything else to work on

QizhouZhang97 (approved):

PR #658, opened 2026-09-10
Fix/env check mlnx qos no sudo

amirakb89:
# fix(env_check): make the mlx5 QoS and DCQCN checks work without root

  `mori check` steps 2 and 3 both failed on any host where `sudo` needs a password — a
  Slurm step, CI, an unprivileged container. Two independent causes, one commit each.

  ## Step 2 — `check_mlx5_qos`

  Called `sudo mlnx_qos`, but the DCB netlink getters it reads are unprivileged; the
  `sudo` was never needed. When it failed, `trust` parsed as `'?'`, every port took the
  `log_fail` path, and the loop `continue`d before deriving `MORI_RDMA_SL`/`TC` — so the
  step reported "could not derive SL/TC from QoS info on any RoCE port" and returned 1.

  Now queried without `sudo`, with a `sudo -n` fallback for hosts that do gate the getters
  (`-n` so a missing password fails fast instead of blocking on a TTY prompt).

  ## Step 3 — `check_mlx5_dcqcn`

  Only knew how to ask `mlxconfig`, which refuses unless it is uid 0. This is not a file
  permission problem: `strace` shows it call `geteuid()`, print `Failed to open the
  device`, and exit 3 **without opening anything** — so no `chmod` or alternate device
  argument helps.

  The mlx5 driver exposes the same state world-readable under `/sys/class/net/<eth>/ecn/`,
  so read that first and keep `mlxconfig` as the fallback. sysfs is also the better source:
  it reports what the driver is running with, while `ROCE_CC_PRIO_MASK_P1` is only what the
  firmware will boot with next time — a distinction the `cluster-network-topology` skill
  also calls out ("a DCQCN fix written to NV config needs a firmware reset or reboot before
  it takes effect").

  A priority counts as enabled only when both legs are on — `roce_np` (emit a CNP on seeing
  an ECN mark) and `roce_rp` (cut the send rate on receiving one) — since one leg alone
  cannot close the control loop.

  ## Measured

  Two ConnectX-7 / ConnectX-6 Dx hosts, no passwordless sudo:

  | | before | after |
  |---|---|---|
  | step 2 | 10 ports `[FAIL]` trust `'?'`, no SL/TC derived | 8 rails `trust=dscp, PFC prio 0 5`, `SL=5 TC=104` |
  | step 3 | 10× `[WARN] cannot query ROCE_CC_PRIO_MASK_P1` | 10× `[OK] DCQCN enabled (live prio mask=255, CNP_DSCP=46)` |

  The two remaining FAILs are the ConnectX-6 Dx management ports, which genuinely have PFC
  disabled.

  Step 3 also gains a check that was previously dead code: `cnp_dscps` is now populated on
  unprivileged hosts, so the CNP-vs-data priority comparison runs and confirms
  `CNP DSCP=46 → prio 6` sits apart from `data DSCP=26 → prio 5`.

  ## Behaviour change worth flagging

  Step 3's source of truth moves from firmware NV config to live driver state, so verdicts
  differ on hosts where the two disagree. That is the intent, but it is a semantic change —
  and a host that previously only WARNed for want of root can now FAIL.

  ## Not exercised

  The `sudo -n` fallback (step 2) and the `mlxconfig` fallback (step 3). On both test hosts
  the primary path answers for every port. A reviewer with an older driver or a sudo-gated
  host would shake those out.

### DIFF
--- a/tools/env_check.sh
+++ b/tools/env_check.sh
@@ -2049,7 +2049,17 @@ check_mlx5_qos() {
         local qos_output trust pfc_line data_dscp data_prio dl p pp
         local -a enabled_arr=() nd_prios=()
         local -A prio_dscp=()
-        qos_output=$(sudo mlnx_qos -i "$eth" 2>&1)
+        # The DCB netlink getters mlnx_qos reads are unprivileged, so query
+        # without sudo first. Where sudo needs a password -- a Slurm step, CI,
+        # an unprivileged container -- the sudo call fails outright, every port
+        # parses trust as '?', and the step fails for want of a password rather
+        # than for a real fault. Keep a fallback for hosts that do gate the
+        # netlink getters, with -n so a missing password never blocks on a
+        # prompt (sudo would otherwise try to read a TTY that is not there).
+        qos_output=$(mlnx_qos -i "$eth" 2>&1)
+        if ! grep -q "Priority trust state" <<< "$qos_output"; then
+            qos_output=$(sudo -n mlnx_qos -i "$eth" 2>&1)
+        fi
 
         trust=$(echo "$qos_output" | grep "Priority trust state" | head -1 | awk '{print $NF}')
         if [[ "$trust" != "dscp" ]]; then
--- a/tools/env_check.sh
+++ b/tools/env_check.sh
@@ -2124,9 +2124,40 @@ check_mlx5_qos() {
     [[ $fail -eq 0 ]]
 }
 
-# Check DCQCN (ECN-based congestion control) for mlx5 RoCE ports via
-# mlxconfig (firmware NV config: ROCE_CC_PRIO_MASK_P1 / CNP_DSCP_P1).
-# Requires NVIDIA MFT (mst/mlxconfig); skips gracefully if unavailable.
+# DCQCN state from the mlx5 driver's ECN sysfs (/sys/class/net/<eth>/ecn/),
+# the unprivileged counterpart to the mlxconfig query below. Prefer it: it is
+# world-readable and instant, where mlxconfig refuses outright unless it is
+# uid 0 -- it checks geteuid() and exits with "Failed to open the device"
+# without opening anything, so no file permission can be relaxed to allow it.
+# Inside a Slurm step or an unprivileged container the firmware query is simply
+# unavailable and the check degrades to one WARN per port. sysfs is also the
+# better answer: it reports the setting the driver is running with, while
+# ROCE_CC_PRIO_MASK_P1 is only what the firmware will boot with next time.
+#
+# Prints "<mask> <cnp_dscp>" with mask in ROCE_CC_PRIO_MASK_P1 form: bit p set
+# when priority p has DCQCN on both legs -- roce_np (emit a CNP on seeing an ECN
+# mark) and roce_rp (cut the send rate on receiving one). One leg alone cannot
+# close the control loop, so it does not count as enabled.
+# Returns 1 if the tree is absent (pre-ECN-sysfs driver, or a non-mlx5 netdev).
+mlx5_dcqcn_sysfs() {
+    local eth="$1"
+    local base="/sys/class/net/$eth/ecn"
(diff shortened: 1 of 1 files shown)


tools/env_check.sh
```diff
@@ -2114,42 +2124,96 @@ check_mlx5_qos() {
...
+# unavailable and the check degrades to one WARN per port. sysfs is also the
+# better answer: it reports the setting the driver is running with, while
+# ROCE_CC_PRIO_MASK_P1 is only what the firmware will boot with next time.
+#
+# Prints "<mask> <cnp_dscp>" with mask in ROCE_CC_PRIO_MASK_P1 form: bit p set
+# when priority p has DCQCN on both legs -- roce_np (emit a CNP on seeing an ECN
+# mark) and roce_rp (cut the send rate on receiving one). One leg alone cannot
+# close the control loop, so it does not count as enabled.
+# Returns 1 if the tree is absent (pre-ECN-sysfs driver, or a non-mlx5 netdev).
+mlx5_dcqcn_sysfs() {
```
QizhouZhang97, line 2142: `mlx5_dcqcn_sysfs` requires `roce_np/enable/$p` **and** `roce_rp/enable/$p` for priority `p`
to count. The "one leg alone can't close the control loop" reasoning is right, but only RP is
per-priority — NP is a port-wide switch that happens to be exposed as eight files. From
NVIDIA's [DCQCN guide](https://enterprise-support.nvidia.com/s/article/howto-configure-dcqcn--roce-cc--values-for-connectx-4--linux-x):

> Notification Point: Sending CNP packets is handled globally per port, **any priority
> enabled here will set it on**. Reaction Point: Handling CNP is per priority configured.

So a host with NP on prio 3 and RP on prio 5 has working DCQCN on prio 5, but this returns
mask 0 and we report `DCQCN disabled` → `fail=1` → `env_check.sh` exits 1. False red in CI,
which is the thing this PR is trying to fix.

Second, smaller one in the same function: `[[ "$np" == "1" ]]` treats anything that isn't a
literal `1` as off, so an `enable` file that reads back empty or `N/A` also lands on FAIL with
no hint we just failed to parse it. Since this branch FAILs rather than WARNs, "off" and
"couldn't read" should diverge — the latter should fall back to mlxconfig.

```bash
    # roce_np is port-wide ("any priority enabled here will set it on"); only
    # roce_rp is per-priority. A value that is not a literal 0/1 means we failed
    # to read the state, not that it is off -- fall back rather than report a
    # FAIL we cannot stand behind.
    local np_any=0 mask=0 p v
    for p in 0 1 2 3 4 5 6 7; do
        v=$(cat "$base/roce_np/enable/$p" 2>/dev/null) || return 1
        [[ "$v" == "0" || "$v" == "1" ]] || return 1
        [[ "$v" == "1" ]] && np_any=1
    done
    for p in 0 1 2 3 4 5 6 7; do
        v=$(cat "$base/roce_rp/enable/$p" 2>/dev/null) || return 1
        [[ "$v" == "0" || "$v" == "1" ]] || return 1
        (( np_any )) && [[ "$v" == "1" ]] && mask=$(( mask | (1 << p) ))
    done
```

Tested against fake ECN trees, this moves exactly one case and leaves the rest alone —
`np=all rp=all` still `255`, and both genuinely-broken configs (`rp=none`, `np=none`) still
correctly report `0`. Only `np=p3 rp=p5` flips from `0` (FAIL) to `32` (OK).

amirakb89: Thanks @QizhouZhang97.  Both addressed in `c5ebb666`. NP is now read as a port-wide flag (OR across all roce_np/enable/*) and applied to each per-priority roce_rp, so the NP-on-3 / RP-on-5 case no longer false-fails. Unreadable or non-0/1 enable values now return 1 so the caller falls back to mlxconfig instead of reporting DCQCN off. Verified on a 10-port CX-7/CX-6 Dx host: step 3 reports mask=255, CNP_DSCP=46 on all ports. Note the mlxconfig fallback itself remains unexercised there, since sysfs answers for every port.

QizhouZhang97 (approved):

PR #662, opened 2026-09-11
feat(cco): Fuse DSV4-Pro wo_b's GEMM with its all-reduce over cco SDMA

yangyuhuiling:
Fuses DSV4-Pro's `wo_b` GEMM with its all-reduce over cco's SDMA copy engines,
shipped as `mori.ops.gemm_ar`. The fused kernel writes each destination's row
band straight into a symmetric window and hands it to the copy engines as soon
as its tiles are done, so the transfer overlaps the GEMM.

* **-19%** on the layer at the model shape (`[16384, 7168]` K=2048, 8x MI355X,
  blockscale): 1419.5 us as the model runs it today (GEMM + NCCL) against
  1148.8 fused. An optional fp8 all-gather leg takes it to 979.3 us, **-31%**.
* **-2.3%** GPU busy end to end in SGLang behind a flag, **-5.0%** with the fp8
  gather. Smaller than the layer win because `wo_b` is ~12% of the profile.
* Public API is `GemmAllReduceOp`, mirroring `EpDispatchCombineOp`: it owns the
  symmetric window, the per-M kernel cache and the row padding. `supports()`
  answers whether a shape is expressible.
* **Needs `BUILD_CCO_SDMA=ON`.** With it off every put silently does nothing,
  the model still answers fluently, and the fused path measures *faster* than it
  is because it moves no data. `GemmAllReduceOp.self_test()` now catches this at
  construction; the CI image and install script set the flag.

Full measurements, how to reproduce each one, the model-level evaluation of the
fp8 option, and the negative results kept deliberately are in
[`docs/MORI-GEMM-AR-BENCHMARK.md`](https://github.com/ROCm/mori/blob/xiangch/flydsl-cco-ar/docs/MORI-GEMM-AR-BENCHMARK.md).


python/mori/ops/gemm_ar/op.py
```diff
@@ -0,0 +1,308 @@
...
+                f"M={m} must be a multiple of world_size * block_m = "
+                f"{self.world_size * self.block_m}; use padded_m()/pad_rows()"
+            )
+        gemm, parts, c, out = self._compiled(m)
+        stream = fx.Stream(torch.cuda.current_stream())
+        gemm(
+            a_fp8.contiguous().view(torch.int8).view(-1),
+            b_preshuffled.contiguous().view(torch.int8).view(-1),
+            c.view(-1),
+            a_scale.reshape(-1),
```
jhchouuu: **[P1] Preserve the documented column-major A-scale order.**

For a logical `[M, K/128]` column-major scale tensor, `reshape(-1)` creates a copy in logical row order. The kernel instead indexes scales as `kb * M + row`, so this changes which scale belongs to each K block. The benchmark and single-GPU test avoid this by passing `sa.t().reshape(-1).contiguous()`.

Reproduced through the installed public `GemmAllReduceOp` on 2 MI355X ranks with `M=512, N=1024, K=512`, using `make_operands(..., quant="blockscale")` from this PR's benchmark. `sa.shape == (512, 4)` and `sa.stride() == (1, 512)`. Passing `sa` gives **relL2=0.26081455** on both ranks; passing `sa.t().reshape(-1)` to the same op with the same operands gives **0.00244654**.

Please preserve the physical order for the documented 2-D layout and explicitly distinguish it from an already flattened physical-order input. This needs a public-op test, since the existing kernel-level tests bypass this conversion.

python/mori/ops/gemm_ar/op.py
```diff
@@ -0,0 +1,308 @@
...
+        """``m`` rounded up to what :meth:`__call__` accepts."""
+        return padded_m(m, self.world_size, self.block_m)
+
+    def _make_cfg(self, m: int) -> ArConfig:
+        cfg = ArConfig(
+            world_size=self.world_size,
+            m=m,
+            n=self.n,
+            recv_slots=self.world_size,
+            counter_chunks=counter_chunks(m, self.world_size, self.block_m),
```
jhchouuu: **[P1] Keep control-region offsets stable when reusing the op for another M.**

`counter_chunks` changes the counter-region size, which moves `lock_off` and `input_off`, but this instance reuses one window that was zeroed only during construction. On 8 ranks:

| M | chunks | lock_off | input_off |
|---|---:|---:|---:|
| 4096 | 4 | 5632 | 5760 |
| 8192 | 8 | 5760 | 5888 |

Thus the second configuration interprets the previous GEMM payload as its submit locks. With one `GemmAllReduceOp(comm, n=7168, k=512, m_max=8192)`, the sequence `M=4096 -> M=8192` produced a correct first result (**relL2=0.00235044**) and then hung until a 90-second timeout (exit 124). A separate inspection read nonzero payload words such as `964770028` and `958839021` from the new lock region before launching the second call. Scales were flattened into the correct physical order to isolate this from the scale issue.

As a diagnostic control, synchronizing ranks and zeroing the window before each call made both calls complete with relL2 approximately **0.00235**. Please allocate a fixed control layout for the capacity and define how counters are initialized/reused across shapes; the completion modulus also depends on the shape. The solution should remain valid when graphs for different M values are replayed alternately, rather than relying only on a Python-side last-M check.

python/mori/ops/gemm_ar/op.py
```diff
@@ -0,0 +1,308 @@
...
+    """Whether this shape is *expressible*, which is not whether it is faster.
+
+    Profitability depends on how much GEMM there is to hide the transfer behind
+    and is the caller's call -- see the measured curve in ``README.md``.
+    """
+    return (
+        2 <= world_size <= 8
+        and n % block_n == 0
+        and n % SCALE_BLOCK_K == 0
+        and k % SCALE_BLOCK_K == 0
```
jhchouuu: **[P2] Reject K=128 unless a single-K-block pipeline is implemented.**

The divisibility check admits K=128, but the GEMM unconditionally prefetches block 1 and executes both `K_ITERS - 2` and `K_ITERS - 1` tail steps. Reproduced with `supports(512, 1024, 128, 2) == True`: construction and execution succeed, yet both ranks return finite output with **relL2=1.41160023**. The otherwise equivalent K=256 control gives **0.00244162**.

Please require `K >= 256` in both the support predicate and execution validation, or implement the one-block case. There is a related mismatch for world sizes 3/5/6/7: `supports()` returns True, while `ArConfig.validate()` rejects them because 512 threads is not divisible by that world size. These checks should share the actual execution constraints.

python/mori/ops/gemm_ar/layout.py
```diff
@@ -0,0 +1,577 @@
...
+
+    @property
+    def remote_bytes_per_rank(self) -> int:
+        """Bytes crossing xGMI per rank: (P-1)/P*2 for 2-stage, (P-1) for 1-stage."""
+        p = self.world_size
+        if self.stage == 1:
+            return (p - 1) * self.nbytes
+        return 2 * (p - 1) * self.nbytes // p
+
+    def validate(self) -> None:
```
jhchouuu, line 645: **[P2] Validate mode strings and configuration constraints before allocating resources.**

`validate()` does not check the dtype/transport strings: `gather_dtype="FP8"` is accepted but selects BF16, and `gather_dtype="fp8", gather_transport="sla"` is accepted but selects SDMA. A configuration typo silently changes the precision or performance path.

Other constraints are checked only after construction. A host-side probe with a recording FakeComm confirmed that `sdma_queues=0`, `block_m=64`, and `n=512, gather_dtype="fp8"` all reach allocation/window/dev-comm creation before `_compiled()` rejects them; `block_m=64` yields an empty `AssertionError`. This probe tests Python validation order, not whether the real CCO API accepts invalid queues.

Please centralize legal mode values, positive sizes/queue counts, tile restrictions, and the FP8 N-alignment constraint before acquiring resources. In particular, the public op always enables permlane, which requires `block_n=256`. `Literal` annotations would help callers discover values, with runtime `ValueError`s reporting the supplied and allowed values.

python/mori/ops/gemm_ar/op.py
```diff
@@ -0,0 +1,308 @@
...
+            )
+        if m % (self.world_size * self.block_m):
+            raise ValueError(
+                f"M={m} must be a multiple of world_size * block_m = "
+                f"{self.world_size * self.block_m}; use padded_m()/pad_rows()"
+            )
+        gemm, parts, c, out = self._compiled(m)
+        stream = fx.Stream(torch.cuda.current_stream())
+        gemm(
+            a_fp8.contiguous().view(torch.int8).view(-1),
```
jhchouuu, line 573: **API usability: validate tensor metadata before byte reinterpretation.**

The public call currently checks M only. BF16 A/B tensors and a BF16 A-scale tensor can reach this launch boundary: A/B are reinterpreted as int8 bytes and the scale dtype is forwarded unchanged. I confirmed this with a kernel stub, without launching a GPU kernel on invalid inputs.

Please validate the cheap metadata here: operand dimensions and N/K, the supported FP8 encoding (or an explicitly documented raw-byte representation), FP32 scale dtype/layout, and the expected GPU/device. Also make the contiguous-input policy explicit. This would turn common integration mistakes into actionable Python errors before JIT/launch, rather than letting the kernel interpret incompatible storage. Type annotations on the public boundary can complement these checks.

python/mori/ops/gemm_ar/op.py
```diff
@@ -0,0 +1,308 @@
...
+        reqs.gda_signal_count = 0
+        reqs.gda_counter_count = 0
+        # One queue per (source, destination) pair is all the pipeline uses -- a
+        # destination is one xGMI link, and splitting a transfer across queues
+        # only costs bandwidth. Asking for world_size would create world_size
+        # queues *per peer* (56 on an 8-rank node, to use 7); inside a process
+        # that already holds SDMA engines that overruns the per-engine queue
+        # slots and hsaKmtCreateQueueExt fails.
+        reqs.sdma_queue_count = sdma_queues
+        self.dev_comm = comm.create_dev_comm(reqs)
```
jhchouuu: **[P2] Provide an explicit op lifecycle for long-lived communicators.**

The op owns three handles but has no `close()` or context-manager support. `Communicator._resources` strongly retains them, so deleting the op does not release its allocation/window/SDMA queues. In a real two-rank probe, resource count went `0 -> 3`; after `del op; gc.collect()`, the op's weakref was empty but the communicator still held all three resources. They were released only when the communicator was destroyed.

This matters when replacing an op or reloading a model while keeping the communicator alive; the model-shape window alone is about 700 MiB per rank. Please follow the existing `dispatch_combine_v2.EpDispatchCombineOp.close()` pattern: an idempotent close/context manager, dependency-ordered cleanup of this op's dev-comm/window/memory and cached views, and rollback of handles acquired before a constructor failure. Document the synchronization requirement before releasing asynchronous resources; ownership of the supplied communicator should remain with the caller.

python/mori/ops/gemm_ar/op.py
```diff
@@ -0,0 +1,308 @@
...
+    def pad_rows(self, x: torch.Tensor, m_pad: int) -> torch.Tensor:
+        """Zero-extend ``x`` to ``m_pad`` rows, in a buffer reused across calls.
+
+        A GEMM row and a reduce-scatter row both depend only on the same input
+        row, so the added rows produce zeros that the caller slices off. Pad
+        *before* quantising: that is what keeps the A scale's column-major
+        ``[K/128, M]`` layout intact without touching it.
+        """
+        m, k = x.shape
+        if self._pad_in is None or self._pad_in.shape[0] < m_pad:
```
jhchouuu: **[P2] Include dtype/device/K in the padding-cache compatibility check.**

Only row capacity is checked. After one BF16 `pad_rows` call, a same-size FP32 call reuses the BF16 buffer and silently rounds its input. A CPU-tensor probe returned BF16 for the FP32 input and changed **1.0010000467 to 1.0**. The method promises zero-extension, not an additional conversion.

Please reallocate when `(device, dtype, K)` changes, or explicitly restrict these properties and reject a mismatch. Validate `m_pad >= m` as well. As with the output of `__call__`, document that the returned padding buffer aliases reusable storage and can be overwritten by later calls.

benchmark/cco/flydsl/gemm_ar/bench_gemm_ar.py
```diff
@@ -0,0 +1,701 @@
...
+        validated = bool(ok.item())
+
+        if rank == 0:
+            result = {
+                "mode": args.mode,
+                "world_size": world_size,
+                "m": args.m,
+                "n": args.n,
+                "k": args.k,
+                "quant": args.quant,
```
jhchouuu, line 481: **[P2] Record the effective gather/transport configuration in RESULT_JSON.**

Neither `RESULT_JSON` nor `--json-out` records `gather_dtype`, `gather_transport`, or `fuse_quantize`. My FP8/LSA result (**979.8 us**) and FP8/SDMA result (**1109.0 us**) both say only `mode="fused-sdma"`. Once separated from the command line or a manually chosen filename, the records cannot identify which path was measured.

Please include the effective dtype/transport, queue count, atomic ordering, quantize-fusion flag, and whether validation ran, plus the resolved tile order rather than only `"auto"`. Keeping the effective configuration in one small record would also avoid forgetting a field whenever a new benchmark option is added.

python/mori/ops/gemm_ar/README.md
```diff
@@ -0,0 +1,542 @@
...
+op = GemmAllReduceOp(comm, n=7168, k=2048, m_max=16384)
+b_shuffled = preshuffle_b(b_fp8)              # [N, K] -> MFMA preshuffled
+out = op(a_fp8, b_shuffled, a_scale, b_scale)  # [M, N] bf16, already reduced
+```
+
+`a_scale` is the A 1x128 block scale, physically `[K/128, M]` fp32 (i.e. `[M,
+K/128]` column-major -- what `aiter_per1x128_quant(transpose_scale=True)`
+emits). `b_scale` is `[N/128, K/128]` fp32 row-major.
+
+`M` must be a multiple of `world_size * block_m`; `op.padded_m(m, world_size)`
```
jhchouuu: **Fix the padding example and provide one runnable end-to-end API example.**

`op.padded_m(m, world_size)` raises `TypeError`: the instance method takes only `m`; the module-level helper takes `world_size`. I verified the documented call fails.

The four-line quickstart also leaves the hard integration steps implicit. Please include a minimal example covering rank/device selection, communicator/VMM sizing, padding **before quantization**, the required scale layout, one-time B preshuffling, warmup for each M before graph capture, invocation, slicing off padding, and cleanup. Include the SDMA build/runtime prerequisites.

The API contract should state the collective ordering/configuration requirements, use of the current stream and asynchronous completion, and whether an instance can be used concurrently given its shared window/cache/padding buffer. The existing output-alias warning is useful and should appear in the quickstart too. A complete example may be sufficient without adding a new warmup API.

python/mori/ops/gemm_ar/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+import importlib
+
+from .layout import ArConfig, MAX_WORLD, ar_config, select_stage
+from ._shuffle import preshuffle_b
+
+_LAZY = {
+    "GemmAllReduceOp": "op",
+    "counter_chunks": "op",
+    "padded_m": "op",
+    "supports": "op",
```
jhchouuu, line 39: **Non-blocking suggestion: keep the pure shape helpers independent of FlyDSL.**

`padded_m`, `counter_chunks`, and `supports` perform integer arithmetic but resolve through `op.py`, importing the kernel stack. With FlyDSL hidden through an import hook, package import and `ArConfig` work, while accessing all three helpers raises `ModuleNotFoundError`. Callers cannot even calculate padding or check a shape without the compiler dependency.

Moving these functions and their constants into the existing `layout.py` would preserve lazy loading for actual kernels while keeping CPU-side configuration and tests lightweight.

python/mori/ops/gemm_ar/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+    "build_lsa_ar": "kernels_lsa",
+}
+
+__all__ = [
+    "ArConfig",
+    "ar_config",
+    "GemmAllReduceOp",
+    "MAX_WORLD",
+    "build_lsa_ar",
+    "build_sdma_ar",
```
jhchouuu, line 56: **Non-blocking suggestion: define a smaller stable API and align its defaults.**

These package-level entry points have different defaults: `GemmAllReduceOp` requests one SDMA queue, while `compile_fused_gemm_scatter`/`build_sdma_phases`/`build_sdma_ar` default to eight; FP8 gather defaults to LSA in the op but to SDMA in `ArConfig`. Moving between entry points therefore changes resource requirements or transport.

Please distinguish the stable caller-facing surface from low-level experimental builders, and either share default constants or make intentional differences explicit. `ArConfig` versus `ar_config` is another avoidable ambiguity: comments mention `ArConfig(comm_dtype=...)`, but only `ar_config` accepts that argument. This is a maintainability/API-scope suggestion, not a request for another backend abstraction.

python/mori/ops/gemm_ar/op.py
```diff
@@ -0,0 +1,308 @@
...
+        # One queue per (source, destination) pair is all the pipeline uses -- a
+        # destination is one xGMI link, and splitting a transfer across queues
+        # only costs bandwidth. Asking for world_size would create world_size
+        # queues *per peer* (56 on an 8-rank node, to use 7); inside a process
+        # that already holds SDMA engines that overruns the per-engine queue
+        # slots and hsaKmtCreateQueueExt fails.
+        reqs.sdma_queue_count = sdma_queues
+        self.dev_comm = comm.create_dev_comm(reqs)
+
+        self._cache: dict[int, tuple] = {}
```
jhchouuu: **Non-blocking suggestion: give the cached plan named fields and cache the ordered tail.**

`dict[int, tuple]` hides the meaning and types of `(gemm, parts, c, out)`. `parts` in turn mixes launchers with tuples of phase names, so the hot path needs to understand two conventions and perform `parts[phase]` lookups after reading `parts["fused_order"]`.

A small NamedTuple/dataclass with `gemm`, `tail`, `input`, and `output` fields would make the contract clearer. Resolve `tail = tuple(parts[name] for name in parts["fused_order"])` when building the entry and iterate the callables directly. The benchmark can keep named phases for per-phase timing. This can stay a small local change rather than becoming a general execution-graph framework.

python/mori/ops/gemm_ar/kernels_fused.py
```diff
@@ -0,0 +1,1603 @@
...
+    if transport not in ("sdma", "lsa"):
+        raise ValueError(f"transport must be sdma or lsa, got {transport!r}")
+    # The C-store variant is picked by a chain of const_expr branches, so an
+    # unimplemented combination does not fail -- it quietly falls through to a
+    # weaker one. That already cost a wrong conclusion once: --lane-transpose had
+    # no direct-LSA branch, so `--mode fused-lsa --lane-transpose` silently ran
+    # stages 1+2 and was measured as "no benefit from stage 3".
+    if permlane and not swap_ab:
+        raise ValueError(
+            "--permlane requires --swap-ab. Both are on by default in "
```
jhchouuu, line 857: **Non-blocking suggestion: make library errors describe Python arguments.**

A Python caller currently gets instructions to pass `--no-permlane`, `--no-swap-ab`, or change `bench_gemm_ar.py`, even when calling the library from a server with no such CLI. Please report the actual constraint, e.g. `permlane=True requires swap_ab=True`, with the supplied values. CLI-specific guidance can live in the benchmark parser.

This also makes it easier to share validation between the public op and low-level builders without embedding benchmark defaults in library error messages.

benchmark/cco/flydsl/gemm_ar/bench_gemm_ar.py
```diff
@@ -0,0 +1,701 @@
...
+            "leader",
+            "none",
+            "writethrough",
+            "wt-agent",
+            "raw-wt",
+            "raw-wt-agent",
+            "raw-wt-leader",
+        ),
+        default="none",
+        help="release before the epilogue push. Correct: 'agent' (default, "
```
jhchouuu: **Correct the help text to describe the current synchronization defaults.**

The parser defaults `--fence` to `"none"`, but this help says `"agent"` is the default and lists `"none"` as incorrect. That tells users the default benchmark is invalid, despite the current default passing the reproduced numerical tests.

Please rewrite this around the current barrier/atomic protocol and explicitly identify which experimental combinations remain unsupported or racy. The history of earlier fence experiments belongs in the benchmark document rather than contradictory current CLI guidance.

python/mori/ops/gemm_ar/kernels_fused.py
```diff
@@ -0,0 +1,1603 @@
...
+                "rocdl.waves_per_eu": waves_per_eu,
+                "rocdl.flat_work_group_size": "512,512",
+            },
+        ).launch(grid=(grid_x, 1, 1), block=(512, 1, 1), stream=stream)
+
+    return launch
+
+
+LSA_NOTE = """\
+The LSA transport is not implemented, and the reason is a measurement rather than
```
jhchouuu: **Non-blocking cleanup: remove stale exported notes and consolidate the experiment narrative.**

`LSA_NOTE` still says the LSA transport is not implemented, although this file implements direct LSA and the benchmark exposes `fused-lsa`. It is also exported in `__all__`.

There are related stale descriptions elsewhere: `kernels_sdma.py` says fused quantize is the default while the signature uses `False`; the README switches between K=2048 and K=1024 and retains “Nothing here is a production kernel”; comments still describe a second blockscale accumulator although `_BlockScaleK` now rescales one running accumulator.

Please remove the obsolete exported string, keep current invariants and their reasons beside the code, and maintain measurements/history in one document. This should make the synchronization decisions easier to review without losing the useful negative results.

python/mori/ops/gemm_ar/kernels_fused.py
```diff
@@ -0,0 +1,1603 @@
...
+    b_preshuffled: bool = True,
+    waves_per_eu: int = 2,
+    xcd_swizzle: int = 0,
+    fuse: bool = True,
+    rotated: bool | None = None,
+    n_stripe: int | None = None,
+    quant: str = "ptpc",
+    sdma_queues: int = 8,
+    swap_ab: bool = False,
+    store_probe: bool = False,
```
jhchouuu, line 790: **Non-blocking simplification: move deliberately incorrect probes out of the shipped kernel path.**

This entry point has 22 keyword-only arguments and 11 fence modes, including `_AdjacentLaneProbeC`/`_WideStoreProbeC` paths whose output is wrong by construction and `raw-wt*` experiments documented as racy. The public op uses a much smaller, fixed subset of these choices.

Please consider moving the incorrect probes and historical tuning variants under benchmark/experimental, leaving the supported kernel path with a small set of validated choices. Negative results and their reproduction commits are useful to retain in documentation. The completion counters, submit locking, and barrier invariants themselves have a clear purpose and should not be removed merely to reduce line count.

python/mori/ops/gemm_ar/_gemm_a8w8_8wave.py
```diff
@@ -0,0 +1,625 @@
...
+    # Group-of-WGM row swizzle for L2 reuse within one XCD.
+    num_wgid_in_group = wgm * num_pid_n
+    group_id, intra_group = split_row_major_2d(wgid_remap, num_wgid_in_group)
+    first_pid_m = group_id * wgm
+    group_size_m = _min(num_pid_m - first_pid_m, wgm)
+    pid_n, intra_group_m = split_row_major_2d(intra_group, group_size_m)
+    return first_pid_m + intra_group_m, pid_n
+
+
+def compile_fp8_gemm_8w(
```
jhchouuu, line 371: **Non-blocking simplification: vendor the components actually used, or make this full pipeline an explicit benchmark reference.**

The module docstring says `kernels_fused.py` reuses the loaders/MFMA/store helpers but does not call this pipeline. Repository search found no caller of `compile_fp8_gemm_8w` in this PR, yet the remainder of this file carries another full GEMM compile/launch implementation and documents that its upstream under-wait remains unfixed.

Keeping roughly 250 unused lines of a second pipeline in the installed package adds another implementation for maintainers to reason about. Please retain the required components with upstream revision/license attribution; if the original full pipeline is needed as a reference, give it an explicit benchmark location and caller.

python/mori/ops/gemm_ar/layout.py
```diff
@@ -0,0 +1,577 @@
...
+    #: ``"fp8"`` quantises to e4m3 with fp32 scales, halving the bytes. The
+    #: reduce always accumulates in fp32 and ``output`` is always bf16 -- only
+    #: the wire changes.
+    #:
+    #: The legs are separate because they are not equivalent. ``gather`` carries
+    #: *finished* values, so fp8 is one rounding; ``scatter`` carries partial
+    #: sums that are then added across every rank, so its error compounds. They
+    #: also quantise in different places, hence the two scale granularities
+    #: below. ``ArConfig(comm_dtype=...)`` sets both at once.
+    scatter_dtype: str = "bf16"
```
jhchouuu, line 187: **Non-blocking simplification: defer the unimplemented FP8-scatter layout surface.**

The scatter dtype adds payload/scale/recv layout branches and dedicated tests, but `build_sdma_phases` unconditionally raises `NotImplementedError` for FP8 scatter. It also means the public `comm_dtype="fp8"` shorthand selects a configuration the SDMA implementation cannot execute.

The working FP8-gather feature does not require this unimplemented scatter surface. Consider landing the scatter layout/options together with its implementation and numerical tests, keeping this PR focused on BF16 scatter plus the supported gather choices.

tests/python/cco/test_gemm_ar.py
```diff
@@ -0,0 +1,553 @@
...
+        seen.append(record["rel_l2"])
+    assert all(v == pytest.approx(seen[0], rel=1e-6) for v in seen), (
+        f"fused all-reduce is not deterministic across runs: relL2 = {seen}. "
+        "That is an ordering bug in the epilogue, not numerical noise."
+    )
+    assert all(v < 5e-3 for v in seen), seen
+
+
+@pytest.mark.parametrize("mode", ["split-sdma", "fused-sdma", "split-lsa"])
+def test_gemm_ar_modes_agree(mode):
```
jhchouuu, line 541: **Add public-op and changing-input coverage alongside the kernel benchmarks.**

The current tests construct kernels/phases directly; the CI check for `GemmAllReduceOp` only imports the class. I reproduced **48 passed, 1 skipped** across the three PR test files while the public op still had the column-major scale conversion error and the `4096 -> 8192` window-reuse hang described in the inline comments.

Please exercise `GemmAllReduceOp` itself with the documented scale layout, a same-instance `4096 -> 8192 -> 4096` sequence, repeated changing inputs, and graph replay after warmup. Explicitly test both FP8 gather transports: the existing standalone FP8 test uses SDMA, whereas the public op defaults to LSA. This would cover caller-visible behavior and state transitions that additional layout-getter assertions or constant-input benchmark repetitions cannot check.

jhchouuu (changes requested):
Reviewed commit `693fd5451280bc0a8664b67847fb981284e07839`. Requesting changes for the reproduced public-API correctness issues: column-major A-scale handling, same-instance dynamic-M window reuse, and the unsupported K=128 pipeline being accepted. Additional inline comments cover API validation, lifecycle, usability, and clearly marked non-blocking maintainability suggestions.

Validation used an independent Podman container on 8x MI355X, ROCm 7.14.60850, PyTorch 2.12.0+rocm7.14.0, FlyDSL 0.2.4, and a wheel built with `BUILD_CCO_SDMA=ON`. Imports were checked outside the source checkout. Kernel P2P/VMM prerequisites were enabled. The existing test files passed: **48 passed, 1 skipped** (the optional aiter comparison).

The main performance claim reproduced at M=16384/N=7168/K=2048, blockscale, median of 11 graph replays after 3 warmups, maximum over ranks:

| Mode | Time | relL2 |
|---|---:|---:|
| split-sdma | 1471.4 us | 2.35e-3 |
| fused-sdma | 1137.8 us | 2.35e-3 |
| fused-sdma + FP8/LSA | 979.8 us | 2.49e-2 |

That is a 22.7% latency reduction for BF16 fused versus split. The performance result does not establish the correctness of the public wrapper, since the benchmark calls the lower-level builders. Full SGLang/model-level evaluation and the NCCL baseline were not rerun. Host-only/stub probes are identified as such in the comments, separately from real GPU reproductions.

kawhil-amd:
I think there remain two things to try, first is to tune the sdma queue number to max the xgmi bw (like the multi-qp mode in ibgda); second is the  reduce kernel impl way, investigate the one-shot all-reduce or two-shot all-reduce which will be better. Anyway, that is a nice work! @yangyuhuiling

kawhil-amd:
And for local reduce + gather path, is possible to combine them to one kernel, use direct vector store rather than using sdma?

python/mori/ops/gemm_ar/op.py
```diff
@@ -0,0 +1,721 @@
...
+        )
+    except ValueError:
+        return False
+    return True
+
+
+#: fp8 encodings the kernel's byte reinterpretation is valid for. Both are 1
+#: byte and the GEMM only ever sees the bytes, but forwarding a bf16 tensor here
+#: reinterprets two rows as one and produces finite nonsense.
+FP8_DTYPES = (torch.float8_e4m3fnuz, torch.float8_e4m3fn)
```
jhchouuu: **[P1] Reject FNUZ inputs until the kernel supports their encoding.**

These two dtypes are both one byte, but their encodings are not interchangeable. The kernel's MMA atom is fixed to `fx.Float8E4M3FN` (`_gemm_a8w8_8wave.py:286`), while `__call__` only reinterprets the input as int8 bytes. Admitting `float8_e4m3fnuz` here therefore produces incorrect results for an input the API explicitly accepts.

Reproduced on both MI355X machines with an initialized 2-rank communicator, `M=512, N=1024, K=512`, A/B filled with numeric ones, FP32 A/B scales filled with ones, and B passed through `preshuffle_b`:

| Input dtype | Byte encoding of 1 | Expected output | Actual output on both ranks |
|---|---:|---:|---:|
| `float8_e4m3fn` | `0x38` | 1024 | 1024 |
| `float8_e4m3fnuz` | `0x40` | 1024 | 4096 |

Every output element is wrong by the same factor of four; relL2 is approximately 3. `op.self_test()` passed before this comparison, so the collective was moving data correctly.

Please restrict this gfx950 path to the encoding it actually implements, or add an explicit FNUZ conversion/kernel path. Add a numerical test for each accepted dtype; checking element size is insufficient.

python/mori/ops/gemm_ar/op.py
```diff
@@ -0,0 +1,721 @@
...
+            # inside a process that already holds SDMA engines that overruns the
+            # per-engine queue slots and hsaKmtCreateQueueExt fails.
+            reqs.sdma_queue_count = sdma_queues
+            self.dev_comm = comm.create_dev_comm(reqs)
+        except Exception:
+            # Roll back whatever was acquired: the communicator holds a strong
+            # reference to each handle, so a half-built op leaks the window for
+            # as long as the communicator lives.
+            self._closed = False
+            self.close()
```
jhchouuu, line 435: **[P2] Initialize cleanup state before entering the resource-acquisition try block.**

This rollback calls `close()`, which immediately executes `self._cache.clear()`, but `_cache` is not initialized until line 381, after the entire try block succeeds. Any allocation/window/dev-comm failure therefore reaches cleanup with `_cache` absent, masking the original exception and preventing the handle-release loop from running.

A host fault-injection probe let `alloc_mem` and `register_window` succeed and made `create_dev_comm` raise `RuntimeError`. The observed result was `AttributeError: 'GemmAllReduceOp' object has no attribute '_cache'`, with **2 resources acquired and 0 closed**. The injected error survived only in `__context__`. These were recording fake resources, not a deliberately induced hardware OOM.

Please initialize `_cache`, `_pad_in`, and all other cleanup state before resource acquisition, and test failures at each acquisition step. Rollback should release the handles acquired by this construction and propagate the original error. This does not require destroying communicator-scoped shared SDMA queues.

python/mori/ops/gemm_ar/op.py
```diff
@@ -0,0 +1,721 @@
...
+        if block_n != DEFAULT_BLOCK_N:
+            raise ValueError(
+                f"block_n must be {DEFAULT_BLOCK_N}: the op always compiles with "
+                f"permlane, whose lane transpose is written for that width"
+            )
+        if max_shapes is None:
+            max_shapes = default_max_shapes(
+                padded_m(m_max, comm.nranks, block_m), comm.nranks, block_m
+            )
+        if block_m < 1 or m_max < 1 or sdma_queues < 1 or max_shapes < 1:
```
jhchouuu: **[P2] Share the actual tile constraints between supports(), construction, and execution.**

The K=128/world-size cases are fixed, but `block_m` is only checked for positivity here, whereas the kernel requires `BLOCK_M >= 128` and `BLOCK_M % 128 == 0`.

Reproduced at `M=512, N=1024, K=512, world_size=2`:

- `supports(..., block_m=64)` returns `True`.
- Construction with `block_m=64` reaches all three resource acquisitions; `_compiled(512)` then raises an empty `AssertionError`.
- `supports(..., block_m=0)` raises `ZeroDivisionError`; construction also calculates padding/default_max_shapes before reaching this positive-value check.

The constructor/preparation ordering was checked with recording fake resources, without launching an invalid tile on a GPU. Please validate tile parameters before any padding arithmetic or allocation and use the same execution constraints in `supports()`. Unsupported shapes should be rejected at that boundary instead of failing during the first real call.

python/mori/ops/gemm_ar/README.md
```diff
@@ -0,0 +1,591 @@
...
+
+# The window comes out of the communicator's VMM reservation, so size that
+# first -- the op cannot grow it later.
+N, K, M_MAX = 7168, 2048, 16384
+vmm = 2 * GemmAllReduceOp.window_bytes_for(world, m_max=M_MAX, n=N) + (64 << 20)
+
+uid = [bytes(Communicator.get_unique_id()) if rank == 0 else None]
+dist.broadcast_object_list(uid, src=0)
+
+with Communicator.init(world, rank, uid[0], per_rank_vmm=vmm) as comm:
```
jhchouuu: **[P2] Reconstruct UniqueId before passing the broadcast bytes to Communicator.init.**

`uid[0]` is a `bytes` object, but `Communicator.init` accepts the Python `UniqueId` wrapper or its underlying Cython object and does not convert bytes automatically. Executing this call raises:

```text
TypeError: Argument 'uid' has incorrect type (expected mori.cco.cco.UniqueId, got bytes)
```

Please import `UniqueId` and call `Communicator.init(world, rank, UniqueId.from_bytes(uid[0]), per_rank_vmm=vmm)`, as the benchmark and new worker already do.

Please also bring the adjacent contract in line with the final code: `_flatten_a_scale` now rejects logical row-major `[M,K/128]` rather than accepting either memory order; the default shape capacity is no longer 8; and `close()` deliberately retains the dev-comm until communicator teardown. The example's returned view also needs to be consumed or cloned, with the necessary synchronization, before the op context frees its backing window.

jhchouuu (changes requested):
Re-reviewed `1531cfe8ddd23abed2897182b7cc73ccf73b11c3`. **The four inline issues below are the remaining merge blockers from this review.** The earlier style, vendoring-attribution, and module-organization suggestions are non-blocking.

The original column-major scale, dynamic-M window, and K=128 findings have targeted fixes, and those regression cases now pass. Validation on 8x MI355X with ROCm 7.14.60850, PyTorch 2.12.0+rocm7.14.0, FlyDSL 0.2.4, and `BUILD_CCO_SDMA=ON`:

- Existing benchmark + numerical/public-op/ABI tests: **88 passed, 1 skipped** (optional aiter comparison).
- Additional 8-rank changing-input tests alternating captured graphs for M=4096/8192/16384 passed for BF16, FP8/LSA, and FP8/SDMA.
- After waiting for another CI workload to exit, the isolated model-shape performance test passed: **1130.0 us fused vs 1466.9 us split**, approximately 23% lower latency. No external GPU processes were observed in that measurement interval.
- The dependency audit found no aiter runtime dependency or external prebuilt-GEMM import in this path; a GEMM numerical test also passed with those imports blocked.

Performance and dependency structure are therefore not the reasons for this request-changes review. The accepted FNUZ input produces incorrect values, and the remaining exception/validation/quickstart paths need the concrete fixes described inline. Host fault-injection results are identified separately from GPU numerical reproductions.

yangyuhuiling:
@kawhil-amd Thanks — both of these have measurements behind them already, so let me put the data here rather than just an opinion.

### 1. Tuning the SDMA queue count

The multi-QP analogy does not carry over, because the hardware queues are **already per `(source, destination)` pair**. At `sdma_queues=1` a rank on an 8-rank node still gets 7 concurrent queues, one per peer — one destination is one xGMI link, and cco's concurrency rule is about concurrently-*issuing warps* per queue, of which this pipeline has one. Raising the parameter creates `world_size` queues **per peer** and touches exactly one of them: 56 queues to use 7. Standalone that is only wasteful; inside a server that already holds SDMA engines for its own copies, `hsaKmtCreateQueueExt` starts failing (`anvil.cpp:237`). That is why the op defaults to 1.

The lever that *does* move xGMI bandwidth here is **chunk size**, and it is tuned. Per-kernel at the model shape (M=16384, N=7168, K=2048, 8 ranks):

| chunks | GEMM | drain | reduce | gather | total |
|---:|---:|---:|---:|---:|---:|
| 1 | 257.5 | 493.8 | 44.0 | 496.4 | 1291.7 |
| 2 | 266.5 | 379.4 | 44.2 | 496.4 | 1186.6 |
| **8** | 265.0 | **303.7** | 43.8 | 496.6 | **1109.1** |

The drain falls 38% while the GEMM grows 7.5 µs for the extra counter atomics and the lock. Going finer loses: at chunks=8 each PUT is 3.5 MiB; 16 halves it to 1.75 MiB, under the knee in the SDMA bandwidth curve, for 1130.8 µs. Bandwidth on this path is a function of bytes per op and essentially nothing else, so the tuning is "do not split the transfer more than you need for overlap" rather than "add queues".

### 2. Fusing local reduce + gather into one kernel

Two related things have been measured, and your suggestion is a third that neither quite covers. Being precise about which is which:

**(a) Fusing them with SDMA puts fired from the reduce** — commit `2673ea8` — loses at every band count against the 1148.0 µs unfused baseline:

```
bands        1       4       8      16      32
us      1227.3  1380.1  1630.3  2137.2  3026.7
```

The cause is not the traversal or the puts. Handing a range to a copy engine needs a system-scope release from *every block that wrote it*: 256 blocks × bands fences at ~0.24 µs. The whole reduce is only 42 µs, so the overlap cannot repay it. Isolated at 8 bands: fence without the vmcnt drain 1649.8 µs, drain without the fence 1164.5 µs — the drain is free, the fence is the entire cost. Dropping the fence is not a fix: 1157.9 µs at 4 bands (still behind) and racy, relL2 1.8e-2 at 32 bands against a 2.35e-3 floor, differing per rank.

**(b) Direct vector store instead of SDMA, on the scatter leg** — that is `fused-lsa`, 1571.4 µs against fused-sdma's 1114.1. It spends **730 µs of its GEMM pushing C over xGMI where the copy engines move the same bytes in 499**, and ATT shows 99% of that store time is stall — so coalescing does not help it (the three C-store stages moved it 954.5 → 952.1 µs).

**(c) Your exact suggestion** — one kernel that reduces and then vector-stores the result to peers — is *not* the same as either, and I want to be clear that I have not run it. It sidesteps (a)'s objection entirely: with no external engine there is no range to publish, so the release fence disappears. What it inherits is (b)'s ratio, 730/499 ≈ 1.46× for CU stores against copy engines on the push direction. Applied to the 496.6 µs gather that is ~725 µs, against 540 µs for reduce+gather today — so the expectation is that it loses by roughly that margin, but that is an extrapolation across two different legs, not a measurement.

Worth noting the pull direction already does what you are describing and **wins**: `gather_transport="lsa"` has each rank pull its peers' slices over xGMI with vector loads and widen them on the way in — one kernel, no SDMA, no read-back of landed fp8 from HBM. 957 µs against SDMA's 1019 on the fused layer. So "direct vector rather than SDMA" is right for the gather; it is the *push* direction that loses to the copy engines.

If you think (c) is worth settling rather than extrapolating, I am happy to build it — the reduce already has the values in registers at the right moment, so it is a contained change. Say the word and I will post numbers.

The transferable result from all of this, for whatever the next fusion candidate is: hiding a transfer behind compute pays only when what is hidden is much larger than the cost of publishing it. That held for the GEMM's fused scatter (437 µs of transfer behind a compute-bound GEMM) and failed for the reduce (42 µs, bandwidth-saturated).

yangyuhuiling:
@jhchouuu All four are fixed in `f4caced`. Every one reproduced as described; thank you for the fault-injection detail, it made three of them a matter of reading rather than guessing.

**[P1] FNUZ.** Rejected — `FP8_DTYPES` is now `(torch.float8_e4m3fn,)`. Your 4× is exactly the bias difference: `0x40` is 1.0 under FNUZ's bias of 8 and 2.0 under OCP's 7, on both operands.

The comment justifying the old tuple was the real defect. It argued from element size, which is evidence of nothing, and I have replaced it with the encoding argument plus a note that widening this needs a kernel path and a numerical test per dtype. Worth recording that this was reachable from a real caller, not just from the API surface: sglang runs `normalize_e4m3fn_to_e4m3fnuz` on gfx942, so a weight arriving here as FNUZ would have been silently 4× wrong.

**[P2] Cleanup state.** `_closed`, `mem`/`win`/`dev_comm`, `_cache` and `_pad_in` are all initialised before the first statement that can raise. Tested by injecting a failure at each of the three acquisition steps and asserting both halves of what you asked for: the injected error propagates (not an `AttributeError` with the real one demoted to `__context__`), and everything acquired before the failure was released. The fakes stub `from_gpu_ptr` — their handles carry a null pointer, and without the stub the window-zeroing write faults the context and the fault then surfaces in whatever test runs next.

**[P2] Tile constraints.** New `_tile_constraints(block_m, block_n)` states the granules once and runs first in both `supports()` and `__init__`, before `padded_m` or `default_max_shapes` can divide by them. `supports(..., block_m=64)` is now `False`; so are 1, 192, 0 and -128, and `block_n` 128/384/0, while 256×512 is still accepted. `world_size < 1` is rejected there too, for the same divide-by-zero reason you found with `block_m=0`.

**[P2] README.** `UniqueId.from_bytes(uid[0])` in the quickstart, and the example now synchronises and clones before the context frees the window behind the returned view. The three drifted contracts are corrected: the `a_scale` rule (row-major `[M, K/128]` is rejected, and the paragraph now says *why* — the two 2-D spellings are indistinguishable by shape but need opposite treatment, so guessing gets one of them silently wrong), the `max_shapes` default (every legal M under `m_max`, not 8), and `close()` (the dev-comm is deliberately left to the communicator, since its SDMA queues are comm-scoped). I also added a dtype line so the FNUZ rejection is documented rather than only enforced.

**Verification.** 44 pass in `test_gemm_ar_op.py`, including the 2-rank GPU cases. One note on my own run for completeness: `test_flydsl_ar.py::test_flydsl_ar_matches_aiter_bitwise` fails in this environment, but it fails identically on the unmodified tree, so it is not from this change — it passed for you, so I assume it is local. Two `test_flydsl_ar_gpu_smoke[*-sdma]` cases also failed when I ran the whole `tests/python/cco/` directory in one go and pass in isolation; that is the back-to-back SDMA queue teardown race (`anvil.cpp:237`) that `test_gemm_ar_op.py` already retries around and `test_flydsl_ar.py` does not.

python/mori/ops/gemm_ar/op.py
```diff
@@ -0,0 +1,776 @@
...
+            gather_transport=gather_transport,
+            counter_capacity=MAX_CHUNKS,
+            shape_slots=default_max_shapes(
+                padded_m(m, world_size, block_m), world_size, block_m
+            ),
+            shape_index=0,
+        )
+    except ValueError:
+        return False
+    return True
```
jhchouuu, line 230: **Follow-up 1 — non-blocking, but please prioritize checking the support contract.**

Could you recheck the accepted 256x512 case? At `M=4096, N=7168, K=2048, world_size=8`, `supports(..., block_m=256, block_n=512)` returns `True`, while constructing `GemmAllReduceOp` with the same configuration still raises `ValueError: block_n must be 256` because the op always enables permlane. I reproduced this through the constructor; it rejects before allocating resources.

The new `test_supports_accepts_the_legal_tile_multiples` checks only the predicate, so it currently enshrines that disagreement. If `supports()` describes the public op, it should apply the same restriction; if it intentionally describes a broader kernel family, that scope should be explicit, since its current docstring promises agreement with construction. Please consider validating both sides in the test.

python/mori/ops/gemm_ar/README.md
```diff
@@ -0,0 +1,614 @@
...
+each already *is* that order: flat 1-D, `[K/128, M]`, and `[M, K/128]`
+**column-major** (stride `(1, M)`), which is transposed for free.
+
+`[M, K/128]` **row-major is rejected** with a `ValueError`. It is not that the
+op cannot transpose it -- it is that the two `[M, K/128]` tensors are
+indistinguishable by shape, and they need opposite treatment: the column-major
+one must be read flat, the row-major one must be transposed. Guessing gets one
+of them silently wrong (relL2 0.26 either way). Pass `.reshape(-1)` or `.t()`
+yourself to say which you have. `aiter_per1x128_quant(transpose_scale=True)`
+returns the column-major one, so `.reshape(-1)` is right for it.
```
jhchouuu, line 90: **Follow-up 2 — non-blocking, but please prioritize rechecking this scale guidance.**

For a real PyTorch column-major `[M, K/128]` tensor with stride `(1, M)`, `reshape(-1)` walks logical order and can reorder the physical scale buffer. In a real 2-rank run at `M=512, N=1024, K=512`, the same operands gave:

| A scale passed to the op | relL2 |
|---|---:|
| column-major `sa` directly | 0.00244654 |
| `sa.reshape(-1)` as suggested here | 0.26081455 |
| `sa.t().reshape(-1)` | 0.00244654 |

The wrapper handles the direct column-major case correctly now. Please consider changing this sentence so readers do not recreate the original ordering error before calling it. If a particular quantizer writes kb-major data into a tensor whose metadata remains row-major, describe that specific contract separately from a tensor whose stride is actually `(1, M)`; `.reshape(-1)` has different implications in those two cases. Passing `sa` directly is the simplest documented option for the latter.

tests/python/cco/test_flydsl_ar.py
```diff
@@ -0,0 +1,282 @@
...
+
+def test_flydsl_ar_matches_aiter_bitwise():
+    """Every element is reduced once on its owner rank, so the two agree exactly."""
+    world_size = 2
+    if torch.cuda.device_count() < world_size:
+        pytest.skip(f"requires {world_size} GPUs")
+    pytest.importorskip("aiter", reason="aiter baseline not installed in this venv")
+    record = _run_bench(world_size, "lsa", 256, 1024, {"AR_CHECK_VS_AITER": "1"})
+    assert record["validated"] is True
+    assert record.get("rel_l2_vs_aiter") == 0.0
```
jhchouuu, line 282: **Follow-up 3 — non-blocking: please check whether the optional aiter comparison is fully wired.**

The worker benchmark currently does not read `AR_CHECK_VS_AITER`, call an aiter reference, or emit `rel_l2_vs_aiter`. Once the dependency-presence gate is passed, this therefore fails as `None == 0.0` even when the actual AR result validates.

I reproduced that by supplying an empty module only to bypass `importorskip` and then running the real 2-rank benchmark. No aiter numerical implementation was supplied or executed; this was a diagnostic of the test's missing comparison path. To clarify my earlier validation totals, this test was **SKIPPED**, not passed, in our normal environments because aiter was absent.

This predates the latest two commits, but please consider implementing the optional reference and result field, or adjusting/removing the test until that comparison exists, so an environment with aiter installed does not fail on a missing record field.

jhchouuu (approved):
Approving after re-reviewing `dc8ba7fbb6dd4f8d2dea523de76237c49a51e533` and rerunning the experiments. The FNUZ rejection and acquisition-failure rollback fixes are verified, and the original UID conversion issue is corrected.

The three inline points are **non-blocking follow-ups for your assessment**, with **1 (supports vs construction) and 2 (README scale flattening) especially worth another look**. Their concrete observations are still relevant; this approval should not be read as confirmation that those details are already resolved.

Validation on MI355X with ROCm 7.14.60850, PyTorch 2.12.0+rocm7.14.0, FlyDSL 0.2.4, and `BUILD_CCO_SDMA=ON`:

- Existing public-op, numerical, ABI, and benchmark tests: **101 passed, 1 skipped**.
- FNUZ was rejected independently for A and B; the supported FN path computed the expected result.
- Failure injection at allocation, window registration, and dev-comm creation propagated the original errors and released the resources already acquired.
- At M=16384/N=7168/K=2048, blockscale: **1130.5 us fused vs 1462.7 us split**, approximately **22.7% lower latency**, with both numerical checks passing.

The separate presence-gate probe for follow-up 3 intentionally exposed an assertion failure and is not included in the passing-test count. Thanks for addressing the earlier correctness issues.

Issue #655, opened 2026-09-09
MoRI-IO: a single lost KV-write ACK permanently wedges PD-disagg decode (no QP-error recovery, no consumer-side reclaim)


MohitAMD:
## Summary
In vLLM PD-disaggregated serving (WideEP 1P1D over RoCEv2), prefill sends the KV cache to decode via **one-sided RDMA WRITE**. If **one write's acknowledgment (ACK) is lost on the return path**, the sender's queue-pair (QP) exhausts its hardcoded retry budget (~0.5 s) and goes to **ERROR**. Today that turns into a **permanent per-rank decode wedge**: the prefill write worker gives up/blocks, the decode side is never told, and it holds the request's KV blocks forever. Recovery needs a manual restart.

## What happens
- Prefill = sender, decode = receiver. KV moves by one-sided RDMA WRITE; the sender waits for the receiver NIC's ACK.
- The **data arrives fine**, but the **ACK (the "got it" receipt) is dropped on the way back**.
- The sender waits, retries the fixed number of times, never hears back, and tears its connection down to ERROR; its in-flight writes are flushed.
- The receiver got the data and sent its receipt, so **from its side the connection looks perfectly healthy** — it just waits forever for a transfer that will never resume, holding the KV blocks and the request.
- Net: **one lost receipt -> permanent hang.**

## Why this is a MoRI-IO issue (not just fabric)
The *trigger* (a lost ACK) can happen occasionally on any fabric. The *permanence* is the bug: (a) `BatchWrite` doesn't surface the QP-error, (b) there is **no consumer-side reclaim/timeout**, and (c) the dead QP is not rebuilt/retried.

## Proof (from a production capture, redacted)
**Live QP state from the wedged engine:**
```
prefill  lqpn 125913  state ERR   sq-psn 5498844   (sender)
decode   lqpn 11529   state RTS   rq-psn 5498844   (receiver, still healthy hours later)
```
Sender `sq-psn` == receiver `rq-psn` -> **all data was delivered, in order**; the failure is on the **ACK path only**. Sender QP = ERROR, receiver QP = healthy RTS.

**NIC counter delta, stage-start -> wedge (only these moved; all else 0):**
```
                                  prefill  decode
local_ack_timeout_err                6        0
req_transport_retries_exceeded       1        0
packet_seq_err / out_of_sequence     0        0
np_cnp_sent / ECN-marked             0        0
rnr_nak_retry_err                    0        0
symbol_error / link_downed           0        0
```
[...]

PR #670, opened 2026-09-14
Feat/io rdma retry tunables

amirakb89:
# io/rdma: surface QP-error on fatal async events + expose the RC retry budget

## Summary

Two related hardening changes for the MoRI-IO RDMA producer path, motivated by
ROCm/mori#655 (a single lost KV-write ACK permanently wedging PD-disagg decode).

1. **Make the QP retry budget tunable** via `MORI_IO_QP_TIMEOUT`,
   `MORI_IO_QP_RETRY_CNT`, and `MORI_IO_QP_RNR_RETRY`. Previously the RC QP was
   built with a hardcoded `timeout=14` / `retry_cnt=7` (~0.5 s before teardown),
   which operators could not widen for GPU-loaded receivers that are alive but
   slow to ACK.

2. **Fail in-flight transfers on fatal verbs async events.** QP-error CQEs
   already drive `TransferStatus` to failed, but a dead CQ or device produces no
   CQE at all — leaving every in-flight transfer `IN_PROGRESS` forever and
   wedging the producer. This forwards the async events that make outstanding
   work uncompletable (`QP_FATAL` / `QP_REQ_ERR` / `QP_ACCESS_ERR`, `CQ_ERR`,
   `DEVICE_FATAL`) to a handler that fails the affected ledgers and marks the
   endpoints degraded so later submissions fail fast instead of queueing onto a
   dead QP.

## Relationship to #655

Issue #655 requests five fixes. **This PR addresses #1 (producer surfaces the
QP-error instead of polling forever) and #4 (expose the retry budget).**
Net effect of this PR: the producer degrades a transient QP failure to a failed
transfer (freeing the sender) and lets operators widen the retry window to reduce
trigger frequency — instead of the producer hanging silently on a dead QP.

### Retry-budget env vars

| Env var | Default | Range | Meaning |
|---|---|---|---|
| `MORI_IO_QP_TIMEOUT` | 14 (~537 ms) | 0–31 | Local ACK timeout exponent. `0` disables it (an unresponsive peer will stall). |
| `MORI_IO_QP_RETRY_CNT` | 7 | 0–7 | Transport retries before `RETRY_EXC_ERR`. |
| `MORI_IO_QP_RNR_RETRY` | 7 | 0–7 | RNR-NAK retries (`7` = infinite, the ibverbs default this path uses). |

Defaults are unchanged, so behavior is identical unless an operator opts in.
Out-of-range values fall back to the built-in default. The applied values are
logged at INFO on the application module
(`ibverbs attr.timeout:X attr.retry_cnt:Y attr.rnr_retry:Z`).

### Async fatal-event handling

- `RdmaManager::FailInFlightTransfers` matches affected endpoints by qpn, CQ
  handle, or `ibv_context` (per event scope), fails their ledgers via the new
  `SubmissionLedger::FailAll`, and marks them degraded.
- `PORT_ERR` is deliberately excluded: the link usually recovers and the QP
  survives, so failing transfers on it would be wrong.
- `CqCallbackMeta::status` becomes atomic and is claimed with
  `exchange(nullptr)`, since the CQ poller and the monitor thread can now reach
  the same meta concurrently. Claiming on terminal success also fixes a latent
  bug where a late flush CQE could flip an already-successful transfer to failed.
- Gated by `MORI_IO_FAIL_ON_ASYNC_EVENT` (default on) under the existing
  `MORI_IO_ENABLE_ASYNC_EVENTS`. The monitor is reset first in `~RdmaManager` so
  its handler cannot run against torn-down endpoints.

## Tests

- New unit test `submission_ledger_fail_all` (`tests/cpp/io/test_engine.cpp`)
  covers `SubmissionLedger::FailAll`: each distinct transfer failed exactly once
  (shared-meta batches counted once), orphaned records drained, SQ depth
  released for every posted WR, failure message propagated, and the
  post-`FailAll` CQE race (a late CQE must not resurrect a failed transfer).

## Performance

No regression on the happy path. The only change on the hot completion path is
`CqCallbackMeta::status` becoming `std::atomic` (one relaxed load + one
`exchange` per completion); everything else runs only on the error/teardown path.

A/B measured with `bench_engine`, this branch vs. its base (`f9697e16`), on a
two-node ionic/RoCEv2 setup (MI300-class, `ionic_0`, single NIC pair). Write,
batch 1, GPU memory, 4 QP/transfer, 500 iters, 50 warmup:

| Msg size | Base BW (GB/s) | Branch BW (GB/s) | Δ |
|---|---|---|---|
| 1 MiB  | 27.37 | 27.40 | +0.1% |
| 4 MiB  | 37.03 | 37.23 | +0.5% |
| 8 MiB  | 39.50 | 39.57 | +0.2% |
| 16 MiB | 40.49 | 40.56 | +0.2% |
| 24 MiB | 41.00 | 41.00 |  0.0% |
| 32 MiB | 41.03 | 41.03 |  0.0% |

Every point is within ±0.5% (run-to-run RDMA jitter); peak (~41 GB/s), the
latency curve, and the overall shape are identical, and both validate
byte-identical.

> **Read path not measured on this fabric.** RDMA READ fails at the hardware
> level on this ionic pair — even raw `ib_read_bw` returns `status 12 / syndrome
> 0xb` (the same responder-resource limitation that blocks RDMA atomics here).
> This is environmental, not introduced by this PR: the base commit fails
> identically. The atomic `status` change is exercised by the write completion
> path all the same, so write is a sufficient A/B signal.

## Verification

Exercised on the same two-node ionic/RoCEv2 setup:

- **Retry tunables take effect end-to-end** — with
  `MORI_IO_QP_TIMEOUT=20 MORI_IO_QP_RETRY_CNT=5 MORI_IO_QP_RNR_RETRY=4`, the
  applied-attr log changed 1:1 from the default `14/7/7`; transfers validated
  byte-identical.
- **Async fatal-event path** — a synthetic `DEVICE_FATAL` injected mid-transfer
  drove `FailInFlightTransfers`, marked all endpoints on the context degraded,
  and the next submission was rejected immediately
  (`EP is degraded, rejecting new submissions`) rather than hanging.

Refs ROCm/mori#655

### DIFF
--- a/include/mori/application/transport/rdma/rdma.hpp
+++ b/include/mori/application/transport/rdma/rdma.hpp
@@ -198,6 +198,13 @@ std::optional<uint8_t> ReadIoServiceLevelEnv();
 std::optional<uint8_t> ReadIoTrafficClassEnv();
 bool ReadIoTrafficClassDisableEnv();
 
+// RC retry attributes applied at RTS. Each is bounded to the width the QP
+// context allows, so an out-of-range value is ignored rather than failing
+// ibv_modify_qp and taking down connection setup.
+std::optional<uint8_t> ReadIoQpTimeoutEnv();
+std::optional<uint8_t> ReadIoQpRetryCntEnv();
+std::optional<uint8_t> ReadIoQpRnrRetryEnv();
+
 bool ReadIbEnableRelaxedOrderingEnv();
 int MaybeAddRelaxedOrderingFlag(int accessFlag);
 
--- a/src/application/transport/rdma/providers/ibverbs/ibverbs.cpp
+++ b/src/application/transport/rdma/providers/ibverbs/ibverbs.cpp
@@ -393,10 +393,12 @@ void IBVerbsDeviceContext::ConnectEndpoint(const RdmaEndpointHandle& local,
   attr = {};
   attr.qp_state = IBV_QPS_RTS;
   attr.sq_psn = 0;
-  attr.timeout = 14;
-  attr.retry_cnt = 7;
-  attr.rnr_retry = 7;
+  attr.timeout = ReadIoQpTimeoutEnv().value_or(14);
+  attr.retry_cnt = ReadIoQpRetryCntEnv().value_or(7);
+  attr.rnr_retry = ReadIoQpRnrRetryEnv().value_or(7);
   attr.max_rd_atomic = devAttr->orig_attr.max_qp_init_rd_atom;
+  MORI_APP_INFO("ibverbs attr.timeout:{} attr.retry_cnt:{} attr.rnr_retry:{}", attr.timeout,
+                attr.retry_cnt, attr.rnr_retry);
   flags = IBV_QP_STATE | IBV_QP_SQ_PSN | IBV_QP_TIMEOUT | IBV_QP_RETRY_CNT | IBV_QP_RNR_RETRY |
           IBV_QP_MAX_QP_RD_ATOMIC;
   ModifyOrThrow("RTS", attr, flags);
--- a/src/application/transport/rdma/rdma.cpp
+++ b/src/application/transport/rdma/rdma.cpp
@@ -333,6 +333,45 @@ bool ReadIoTrafficClassDisableEnv() {
   return disable.has_value() && disable.value() == 1;
 }
 
+// The RC retry attributes occupy narrow fields in the QP context, so a value
+// that parses as a uint8_t can still be rejected by ibv_modify_qp. Bound each
+// one here and keep the built-in default when the override is out of range.
+static std::optional<uint8_t> ReadQpRetryAttrEnv(const char* name, uint8_t maxValue) {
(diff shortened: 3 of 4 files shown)


maning00:
Thanks for the work on this. I think two issues need to be fixed before merging:

1. `FailAll()` can run after `Insert()` but before `ibv_post_send()`. The ledger then retires and accounts for the WR, while the submitter can still post it or release its SQ reservation again if posting fails. This can leave RDMA work untracked or corrupt `sqDepth`. The closed-ledger `InsertOrphaned()` cases appear to be part of the same accounting race. Please make the post/retirement transition consistent and test this interleaving.
2. With notifications disabled, `notifPerQp == 0` is valid. The first `Insert()` then returns `0`, which is also `kInvalidRecordId`, so the first submission on that ledger is incorrectly rejected. Please use a sentinel that cannot be a valid record ID and add coverage for this configuration.

PR #676, opened 2026-09-16
tune(ep/v2): add hidden_dim=7168 (DSV4) internode launch geometry

yoqiu-amd:
## Summary
  Adds a per-token-count tuned V2/CCO launch geometry (`_MI308X_EP16_H7168`) for hidden_dim=7168 (DeepSeek-V4-Pro shape), EP16, on MI308X. Before this, hidden_dim=7168 had no table entry and fell through to
  the generic untuned default `(96, 64, 8)` for both dispatch and combine at every token count -- `block_num=96` exceeds MI308X's 80 CU count, so every launch ran a second, short wave behind the first.

  ## Methodology
  - Measured with `--cmd bench` (fp8 dispatch / bf16 combine, EP16, num_qp=1, `MORI_RDMA_TC=160`, `MORI_RDMA_SL=5`) on a 2-node MI308X rig.
  - Per-token-count, per-phase grid sweep (dispatch swept first, then combine with dispatch held fixed at its winner, respecting the dispatch/combine coupling this kernel has).
  - All measurements screened for a known rig-level artifact (AMD PTL feature can cause an abrupt, persistent mid-run latency step in ~25-40% of runs); any run showing this was discarded and re-run.
  - tok=16/32 rows were independently re-verified after an earlier sweep pass was found to be contaminated by an unrelated co-resident job saturating the host's GPUs; the clean re-verification confirmed these
  two rows match `_MI308X_EP16_H6144`'s geometry (measured independently, not copied).

  ## Results (dispatch / combine latency, before -> after, us)
  | tokens | dispatch before (96,64,8) | dispatch after | combine before (96,64,8) | combine after |
  |---|---|---|---|---|
  | 4  | 46.5 | 43.5 (~6%)  | 56.3  | 48.8 (~13%) |
  | 8  | 47.8 | 47.3 (~1%, within noise) | 57.7  | 47.3 (~18%) |
  | 16 | 50.3 | 45.2 (~10%) | 62.3  | 53.0 (~15%) |
  | 32 | 60.4 | 53.9 (~11%) | 75.8  | 68.4 (~10%) |

  ## Files changed
  - `python/mori/ops/dispatch_combine_v2/internode_tuning_configs.py`: new `_MI308X_EP16_H7168` entry + `_TABLE` registration.


jhchouuu (approved):
LGTM, thanks！

PR #677, opened 2026-09-16
feat(EPv2): default to HIP kernel backend on gfx125x

kawhil-amd:
## Summary

- EPv2 kernel backend now defaults to **HIP** (JIT C++) on gfx125x, aligning with the comm path which already defaults to `cco` on that arch.
- Other architectures (gfx942, gfx950) keep `flydsl` as the default — no behavior change for them.
- `MORI_V2_KERNEL_BACKEND` env var and `cfg.kernel_backend` still override the default on all archs.

## Motivation

gfx125x already routes symmetric buffers through the CCO/HIP-VMM path (`MORI_EP_COMM=cco`). The kernel compilation backend was the remaining piece still on FlyDSL by default. With the JIT link fix landed in #fix/allocator-jit-link-c10-hip (`c10_hip` now linked in the JIT path), the HIP backend is ready to be the default on this arch.

## Change

Replaces the static `DEFAULT_BACKEND = "flydsl"` with a `_default_backend()` function that reuses the existing `_is_gfx125x()` arch detection from `dispatch_combine.py`. The function is called lazily at backend-resolution time (config tuning lookup + `__new__` dispatch), not at import time.

## Test plan

- [ ] gfx1250: verify EPv2 dispatch+combine picks the HIP backend without setting any env var
- [ ] gfx942/gfx950: verify EPv2 still picks flydsl by default
- [ ] `MORI_V2_KERNEL_BACKEND=flydsl` on gfx1250: verify override still works
- [ ] `MORI_V2_KERNEL_BACKEND=hip` on gfx942: verify override still works


jhchouuu (approved):
LGTM

PR #680, opened 2026-09-16
docs: refresh installation, EP backend guidance and site validation

sunway513:
The website's installation page omits the published packages, several guide links resolve to missing `.md` pages, and EPv2/UMBP features are difficult to discover. This refresh shares installation instructions with the README, adds a source-backed EP backend selection guide, and exposes the existing UMBP and developer guides.

- Document stable/nightly/source installation together, optional FlyDSL integration, and the HIP internode preview's limitations.
- Label incomplete quickstart snippets as API sketches and point to maintained correctness checks, including the EPv2 pytest wrapper that fails on printed numerical failures.
- Repair local links and anchors, show the source commit or release tag instead of version 0.1.0, and gate publication on a strict CPU-only Sphinx build plus an offline rendered-link check.

Validation: a clean `sphinx -E -n -W --keep-going -b html` build passes with zero warnings. The new checker passes 8,277 local links across 24 HTML pages; positive/negative fixtures cover missing targets and anchors. Black 25.1.0, Ruff 0.12.0 with the repository configuration, and `git diff --check` pass. GPU and fabric examples were checked against current source but were not executed.


jhchouuu (approved):
Verified locally before approving. On this branch `sphinx -E -n -W --keep-going -b html` builds clean (rc=0), while the same command on `main` fails with 143 warnings — orphaned toctree entries, the `myst.xref_missing` anchors in MORI-CCO-GUIDE, and the unsupported `display_version` theme option. The new checker reproduces your numbers exactly (8,277 local links across 24 HTML pages, 0 errors), and I confirmed it actually fails on a broken target and a missing anchor rather than passing vacuously.

Dropping the MORI/torch install is safe: there are no `automodule`/`autoclass`/`autofunction` directives anywhere under `docs/`, so autodoc has nothing to fail on. The referenced test paths, guide links, and the `[bf16-gather]` pytest node id all resolve.

LGTM — merging.

jhchouuu (approved):

PR #691, opened 2026-09-20
gemm_ar: drop the fused epilogue's `fence` knob, which nothing selects

yangyuhuiling:
`compile_fused_gemm_scatter` took a `fence` argument with eleven values. Every caller in the tree — `op.py`, the tests, and `bench_gemm_ar.py`'s own default — left it at `none`, so the two `const_expr` chains it drove never emitted anything and three `StoreC` subclasses were unreachable.

### Why they are unreachable

**The counter is already the release.** The `acq_rel` `atomic_add` emits its own `buffer_wbl2`/`buffer_inv` pair, on thread 0 only, so an added `__threadfence_system` redoes that work per lane. gcnasm's `opus_chunk_sdma_submit` has the same shape and no fence, which is what pointed at it.

Measured at m=16384 blockscale on 8 ranks, twice each — the data #662 already had, in `--help`:

| `--fence` | time | relL2 |
|---|---:|---|
| `none` | 1143.5 / 1144.8 us | 2.350e-3 |
| `agent` | 1797.0 / 1798.5 | 2.350e-3 |
| `all` | 1936.1 / 1937.5 | 2.350e-3 |

The fences cost 650–790 us and bought no accuracy. That paragraph now sits in the epilogue, next to the code it explains, instead of in the help text of an option nobody picks.

### What was removed

The parameter, its eleven-value validation list, the `{fence[0]}` kernel-name tag, both dispatch chains, `_RawWriteThroughStoreC` / `_WriteThroughStoreC` / `_NonTemporalStoreC`, the now-dead `release_fence` import, and the benchmark's `--fence` with its result field.

`direct_fence` is **untouched** — it is Direct LSA's, it is live, and it defaults to `leader`. `kernels_sdma`'s `publish` is a different knob and is also still in use; only its two comments pointing at the deleted `_WriteThroughStoreC` were trimmed.

### Two stale pieces of text went with it

Both contradicted the `none` default they shipped alongside in #662: the module docstring's protocol sketch still listed `__threadfence_system()` as a step, and the epilogue comment still read *"This fence is not optional"*. Read literally, either would send someone looking for a release that is not there.

### Verification

Not assumed — 3 runs a side at m=16384 blockscale on 8 ranks:

```
before (main)  1145.8, 1147.2, 1145.4   median 1145.8
after          1146.7, 1143.8, 1146.2   median 1146.2   +0.03%
```

+0.03% against a 1.9–2.8 us spread within each side, and relL2 bit-identical across all six runs (`2.350369e-03`).

### Note for reviewers

`tests/python/cco/test_gemm_ar.py::test_fused_is_stable_across_repeats` fails on this branch — but it **also fails on a clean `origin/main` (e7e97479)**, with `ERROR ... src/application/transport/sdma/anvil.cpp:237`. It is a pre-existing failure in the SDMA transport, not a regression from this change, and it is not addressed here.


jhchouuu (approved):
LGTM.

